PR #82, opened 2018-08-16
Fix broken links in README

frankh:
The links were going to "https://github.com/CanonicalLtd/dqlite/blob/master/CanonicalLtd/go-dqlite"


freeekanayaka:
Thanks!

Please note that this code base is still under heavy development (although already in used and released inside LXD). In particular I'd like to eventually port the raft code from go-dqlite to dqlite itself, making go-dqlite unnecessary.

PR #98, opened 2019-02-27
Fixed compile issues on macos

nanjj:


freeekanayaka:
Note that dqlite is still a work in progress, and the API will likely change as we approach 1.0. The main change will be that we will not require Go anymore for raft replication, since everything will be done completely in C. You might want to wait for the 1.0 release candidate before trying to do anything with dqlite.

freeekanayaka:
Also, please note that the raft C implementation that dqlite is going to depend on will be Linux-specific, and won't compile on macos. Of course patches will be welcome to add support for macos to the raft C implementation.

PR #155, opened 2019-08-30
Remove the docs and propose them to the website

anthonydillon:
I would like to propose that if discourse is due to be the home of the documentation that we house them in the website for now. I have moved the current docs to [the website project](https://github.com/canonical-web-and-design/dqlite.io/tree/master/docs). These are hosted at dqlite.io/docs.


freeekanayaka:
Sure. Before I actually delete the files, can I get commit rights to the new repository? Also, after a commit is made, is it automatically deployed? Basically especially in this early phase I'll need to iterate a lot on the docs. I'll be on vacation for the next 2 weeks, so I'm not in a hurry.

anthonydillon:
@freeekanayaka Sure of course I'll add you. We like to have a review per PR if that is ok as it could introduce something that breaks the site but we will be on it quickly. Docs will be updated when the site is released which we will kick off when merging the PRs. I just dont want the docs here to go out of date with the docs on the site. 

Now I think about it, docs are now linked to from sites like hackernews so you may want to replace the content of the docs with something like "These docs have been moved to the website project: [link to the docs in GH or live site]".

anthonydillon:
@freeekanayaka ping, has anything changed here? I have updated the diff in the FAQ page on the site.

freeekanayaka:
Sorry of the long delay. I've created another branch in #197 to replace this PR, since there are now conflicts, and as you suggested we need to replace the current docs with pointers to the website project, instead of removing them.

Do you know if the RT about setting up discourse has been making progress?

anthonydillon:
Thank you, makes sense. They are close to having the final discourse instances upgraded which I hope will unblock this discourse from being deployed. I can try to bump the priority of the RT.

PR #157, opened 2019-08-30
Change reference to OS X to macOS

fitojb:
The latest rename of this operating system happened in 2016.

This also fixes a missing word immediately below.


freeekanayaka:
@fitojb thanks!

PR #184, opened 2019-12-03
Case-insensitive column type lookups, DATE support

paulstuart:
The column types are return as defined by the user, so if it was done lowercase it won't match -- this change fixes that.

Also added a DQLITE_ISODATE type to mark the column as being a date only (not a timestamp).

### DIFF
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -5,7 +5,6 @@
 #define DQLITE_UNIXTIME 9
 #define DQLITE_ISO8601 10
 #define DQLITE_BOOLEAN 11
-#define DQLITE_ISODATE 12
 
 #define DQLITE_PROTO 1001 /* Protocol error */
 
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -5,6 +5,7 @@
 #define DQLITE_UNIXTIME 9
 #define DQLITE_ISO8601 10
 #define DQLITE_BOOLEAN 11
+#define DQLITE_ISODATE 12
 
 #define DQLITE_PROTO 1001 /* Protocol error */
 
--- a/src/query.c
+++ b/src/query.c
@@ -7,143 +7,143 @@
  * TODO: find a better way to handle time types. */
 static int value_type(sqlite3_stmt *stmt, int i)
 {
-        int type = sqlite3_column_type(stmt, i);
-        const char *column_type_name = sqlite3_column_decltype(stmt, i);
-        if (column_type_name != NULL) {
-                if ((strcasecmp(column_type_name, "DATETIME") == 0)  ||
-                    (strcasecmp(column_type_name, "DATE") == 0)      ||
-                    (strcasecmp(column_type_name, "TIMESTAMP") == 0)) {
-                        if (type == SQLITE_INTEGER) {
-                                type = DQLITE_UNIXTIME;
-                        } else {
-                                assert(type == SQLITE_TEXT ||
-                                       type == SQLITE_NULL);
-                                type = DQLITE_ISO8601;
-                        }
-                } else if (strcasecmp(column_type_name, "BOOLEAN") == 0) {
-                        assert(type == SQLITE_INTEGER || type == SQLITE_NULL);
-                        type = DQLITE_BOOLEAN;
-                }
-        }
+	int type = sqlite3_column_type(stmt, i);
+	const char *column_type_name = sqlite3_column_decltype(stmt, i);
+	if (column_type_name != NULL) {
+		if ((strcasecmp(column_type_name, "DATETIME") == 0)  ||
+		    (strcasecmp(column_type_name, "DATE") == 0)      ||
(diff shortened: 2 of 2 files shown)


freeekanayaka:
Could you clarify how ```DQLITE_ISODATE``` differs from ```DQLITE_ISO8601```?

According to SQLite's [documentation](https://www.sqlite.org/datatype3.html#date_and_time_datatype), there are only 3 low-level storage "types" for time values: TEXT, REAL and INTEGER. At the moment dqlite supports TEXT (```DQLITE_ISO8601```) and INTEGER (```DQLITE_UNIXTIME```).

paulstuart:
Sorry, I should have commented the code there to that effect. `DQLITE_ISODATE` is intended to indicate the column holds a date value (like in Postgres et al). It is expect that the actual value would be textual (thus the ISO part).

Does this seem reasonable to you? It would also need to be ammended as a type on the Go side (happy to make that change as well).

freeekanayaka:
Right, there's already ```DQLITE_ISO8601``` for a date value stored as text. It's unclear to me what the difference between ```DQLITE_ISO8601``` and ```DQLITE_ISODATE``` would be. Can't we just have:

```C
		if ((strcasecmp(column_type_name, "DATETIME") == 0)  ||
                    (strcasecmp(column_type_name, "DATE") == 0)) {
                    (strcasecmp(column_type_name, "TIMESTAMP") == 0)) {
			if (type == SQLITE_INTEGER) {
			if (type == SQLITE_INTEGER) {
				type = DQLITE_UNIXTIME;
				type = DQLITE_UNIXTIME;
			} else {
			} else {
				assert(type == SQLITE_TEXT ||
				assert(type == SQLITE_TEXT ||
				       type == SQLITE_NULL);
				       type == SQLITE_NULL);
				type = DQLITE_ISO8601;
				type = DQLITE_ISO8601;
			}
```

(i.e. add "DATE" as possible column type declaration string for ```DQLITE_ISO8601```) ?

paulstuart:
The client is only going to see either a UnixTime or or ISO8601 -- both handle a full timestamp. Because we don't get custom type names back there's no way to know if that value is a timestamp or a plain old date.

It's a bit of a nit but I think could be a nice-to-have.

freeekanayaka:
What's the use case for differentiating between "timestamp" and "plain old date"? Aren't they the same thing for a client? As long as they are represented as TEXT.

paulstuart:
I think that sometimes clients want just the date. This is reflected in the fact that PG, MySQL, MSSQL all have date only fields. I encountered this issue in trying to adapt a python ORM to dqlite and this omission broke the test. (I would rather work just in Go but we have a use case to support that and Python).

With Go, this decision can be made at the driver level using the `DatabaseTypeName` call, but that option doesn't work here because there's only the enumerated types available.

While I think it's worth having, I'm not hung up on it and can remove it to get the case-sensitivity fix in. If that path is chose, I only ask that you keep this possibility in mind for the future.

freeekanayaka:
The thing is: unlike PG/MySQL/etc, SQLite does not differentiate between a "timestamp" and a "date", when the representation is TEXT. So it's unclear why the SQLite driver of the Python ORM that you mention would rely on such a difference. It seems a bug of the ORM? Are we talking about SQLAlchemy? Please could you point to the specific test failure that you mention? So I can have more context.

Also, since you mentioned using dqlite with Python, do you have a Python client for dqlite?

freeekanayaka:
Oh I guess it's https://github.com/paulstuart/pydqlite, neat.

paulstuart:
I'm aware of the date/timestamp behavior with SQLite (I'm a longtime user of it) -- this was an attempt to paper over it for a Postres migration.

The python stuff is an adaptation of the rqlite versions (hooray open source!) -- it supports DATE functionality and I while I was making the primary change (case sensitivity) I figured "why not". I'll back it out to keep this PR clean.

freeekanayaka:
Looks good. Would you just use tabs for indentation? Otherwise the diff gets way larger than the actual change.

Tabs are the current default for dqlite, see the ```.clang-format``` file in the top dir:

```
...
UseTab: ForContinuationAndIndentation
...
```

I did that for consistency with Go and Linux kernel C, however I'm planning to switch it back to spaces.

freeekanayaka:
> I'm aware of the date/timestamp behavior with SQLite (I'm a longtime user of it) -- this was an attempt to paper over it for a Postres migration.

As said, it feels something that should be changed in the Python ORM, which I'd expect to transparently deal with the fact that PG/MySQL/etc have "DATE" but SQLite/dqlite do not (like many other feature mismatches), and make consumer code oblivious of that. However, please let me know if this assumption is wrong.

paulstuart:
I'll continue to refine and explore and will let you know. I think this is an awesome project; looking forward to being able to contribute further. Thank you for your patience with me.

PR #201, opened 2019-12-30
fix inverted logic on pid check -- should *NOT* be same pid

paulstuart:
The logic for listening to unix sockets was rejecting requests unless it was coming from the same pid -- it should be the opposite: reject if *same* pid.

With this change applied I can bring up a local cluster on unix sockets with the goal of being to apply an optional TLS proxy for secured inter-server communications.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -415,7 +415,8 @@ static void listenCb(uv_stream_t *listener, int status)
 		if (rv != 0) {
 			goto err;
 		}
-		if (cred.pid != getpid()) {
+		// no need to call ourself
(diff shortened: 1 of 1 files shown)


freeekanayaka:
Hi, the check about the same pid is actually necessary for security reasons, otherwise any other process on the host could connect to the dqlite node.

The recommended way to put a TLS reverse proxy in front of the dqlite node is to run the proxy in the same process of the node. Alternatively you could make the dqlite node use a TCP socket instead of a Unix one.

PR #207, opened 2020-01-31
vfs: vfs__delete: fix double-unlock of &root->mutex

cyphar:
vfs__delete_contents would unlock the passed &root->mutex, but all of
its callers would then also unlock the passed &root->mutex. It turns out
that this works on most architectures without issue, but [apparently][1] on
[some Intel CPUs with TSX enabled][2] this will trigger a general protection
fault.

This was the cause of a [very frustrating bug where LXD would segfault on
start-up][3].

[1]: https://lwn.net/Articles/534758/
[2]: https://software.intel.com/en-us/forums/intel-isa-extensions/topic/675036
[3]: https://bugzilla.opensuse.org/show_bug.cgi?id=1156336

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -619,7 +619,6 @@ static int vfs__delete_content(struct root *root, const char *filename)
 	/* Check that there are no consumers of this file. */
 	if (content->refcount > 0) {
 		root->error = EBUSY;
-		pthread_mutex_unlock(&root->mutex);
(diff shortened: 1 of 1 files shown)


freeekanayaka:
Good catch, thanks. Not sure why the coded ended up that way, perhaps it went unnoticed after a refactoring.

PR #208, opened 2020-02-10
Add timestamped debug logging option

paulstuart:
Having encountered occasional duplicate writes whilst using go-dqlite (https://github.com/canonical/go-dqlite/pull/76) it seemed like it would be beneficial to have an option to track execution state for further debugging.

The emphasis on the debugging is tracking transactions with leadership changes (either by assigning a new leader or killing the existing one).

I expect that there's stuff that might need improvement (or maybe not acceptable at all), but I figured this could be useful.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -33,6 +33,7 @@ void db__close(struct db *db)
 		assert(rc == SQLITE_OK);
 	}
 	if (db->tx != NULL) {
+		DEBUG_TX(db->tx, "freeing open transaction");
 		sqlite3_free(db->tx);
 	}
 	sqlite3_free(db->filename);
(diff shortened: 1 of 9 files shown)


freeekanayaka:
I'm not entirely sure we want to merge this, as it seems ad-hoc for debugging the specific issue of transaction failures due to lost leadership.

I think we'll want to use this code while working on fixing this bug, but then throw it away. At least that's what I've been doing in similar cases (it turns out that debugging messages that you use to debug one issue are very often useless to debug another one, so I tend to write code like this as throw-away).

Let's leave this PR open until the issue is solved tho.

paulstuart:
I have no issue with that -- falling in love with code leads to an unhealthy relationship :-)

freeekanayaka:
Closing this MP for now.

PR #273, opened 2020-11-17
Convert naming of internal identifiers to camelCase

ericcurtin:
A double underscore within an identifier is illegal in C++, for interop.


ericcurtin:
Ran the following:

```
#!/bin/bash
  
for i in $(find . -name "*.c" -o -name "*.h" | xargs grep __ | tr -cs '[:alnum:]_(#' '\n' | grep __ | sort | uniq | cut -d"(" -f1 | sort | uniq | grep -v "^__" | grep -v "##__VA_ARGS__"); do
  echo $i
  lo=$(echo $i | awk -F_ '{print $NF}' | cut -c-1)
  up=$(echo "$lo" | tr '[:lower:]' '[:upper:]')
  up_word=$(echo $i | sed "s/__$lo/$up/g")
  find . -name "*.c" -o -name "*.h" | xargs grep -l "$i" | xargs sed -i "s/$i/$up_word/g"
done

git-clang-format-10 -f
```

Fixes #273

ericcurtin:
I actually think I can fix a few more with another tweak. This isn't quite right:

`lo=$(echo $i | awk -F_ '{print $NF}' | cut -c-1)`

but it fixes a lot of them.

freeekanayaka:
Thanks! I like this change, and it something I wanted to do myself, as it also makes the naming pattern match what we use in libraft.

It would be great if you could convert everything to 100% camelCase, dropping underscores completely, e.g.:

`request_connectDecode` -> `requestConnectDecode`

`tuple_decoderIniT` -> `tupleDecoderInit`

and so on.

ericcurtin:
It can be done, just can be tricky to leave out the function calls, etc. from other libraries. What about defines like:

`SERIALIZE__DEFINE`

what should be the format for those?

freeekanayaka:
> It can be done, just can be tricky to leave out the function calls, etc. from other libraries. What about defines like:
> 
> `SERIALIZE__DEFINE`
> 
> what should be the format for those?

Good question. I'd say `SERIALIZE_DEFINE` is fine.

In general for `#define` if the symbol is declared in a `.h` file (like for `SERIALIZE__DEFINE`) I'd say you can drop the double underscore just fine because those symbols should already be in the form of `<PREFIX>__<VERB>` where `<PREFIX>` is the name of the "module" and it matches the filename of the `.h` file.

On the other hand, if the `#define` is local to a `.c` file probably the `<PREFIX>` part is unnecessary, unless there's a clash with symbols imported by that `.c` file.

ericcurtin:
Ok about to push again, a couple of more thousand lines successfully changed via a fairly hacky script again, not 100% complete. Left anything starting with "dqlite_" alone to be like "raft_

ericcurtin:
Must fix unit tests, I only did a normal "make"

There's also files I should not have touched like the munit* ones, so must revert those changes.

ericcurtin:
I would understand if you wanted this to be two PR's @freeekanayaka One for "s/__/_/g" and another for camelcase. Although there are no functional changes here, in terms of lines of code, there's a huge amount of change.

freeekanayaka:
As you prefer @ericcurtin, I personally don't care if it's one or two PRs. Since it's a mechanical change, size of the diff is not a concern.

Tests in Travis are currently failing for this PR, I didn't look too close at that though.

ericcurtin:
Yeah I suspect I may be text replacing string literals or something like that also. My text replaces are catching comments and other places also, that's useful, until you start breaking stuff I guess.

ericcurtin:
I was changing a string I should not have alright:

` "s/pageSize=/page_size=/g"`

fixed it

ericcurtin:
Good morning Free, hope you had a good weekend! Question, should a token like dqlite_node_set_failure_domain end up as:

1 `dqlite_nodeSetFailureDomain`

or

2 `dqliteNodeSetFailureDomain`

or

3 `dqlite_node_set_failure_domain`

?

raft seems to have any "raft_" identifier as snake_case for the most part. Would this mean we follow number 3 for the "dqlite_" case?

The macro constructs like:

`command_##LOWER##__decode`

and

`#define REQUEST_OPEN(X, ...)`

should these be as camelCase as possible?

freeekanayaka:
> Good morning Free, hope you had a good weekend! Question, should a token like dqlite_node_set_failure_domain end up as:

Hello Eric, I did (lockdown-permitting :), hope you too!

> 1 `dqlite_nodeSetFailureDomain`
> 
> or
> 
> 2 `dqliteNodeSetFailureDomain`
> 
> or
> 
> 3 `dqlite_node_set_failure_domain`

Public interfaces (i.e. anything in the `dqlite.h` header that clients import) should be snake_case, as it's mostly standard in C. The camelCase should be used only for internal non-exported symbols, because it minimizes the risk of conflicts with snake_case symbols imported from other libraries, it makes it clear that those are local symbols and it makes code lines a bit shorter.

> ?
> 
> raft seems to have any "raft_" identifier as snake_case for the most part. Would this mean we follow number 3 for the "dqlite_" case?

Yes, correct, see above.

> 
> The macro constructs like:
> 
> `command_##LOWER##__decode`
> 
> and
> 
> `#define REQUEST_OPEN(X, ...)`
> 
> should these be as camelCase as possible?

I think both CAPITAL_CASE and camelCase are fine for macros.

ericcurtin:
Thanks @freeekanayaka I think I understand now completely.

This should do it:

```
#!/bin/bash

git clean -fdx
git reset --hard origin/master

FIND="find . -name *.c -o -name *.h"
GREP="grep -v 'mu\|dqlite.h'"

autoreconf -i
./configure
CFLAGS=-O0
sed -i "s/\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/#\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/g" Makefile
make -j6 check

for i in $($FIND | $GREP | xargs grep _ | tr -cs '[:alnum:]_#' '\n' | grep _ | sort | uniq);  do
  new_i=$(echo $i | sed "s/__/_/g")
  echo "'$i' '$new_i'"
  $FIND | $GREP | xargs grep -l "$i" | xargs sed -i "s/$i/$new_i/g"
  CFLAGS=-O0
  if ! make -j6 check; then
    $FIND | $GREP | xargs grep -l "$new_i" | xargs sed -i "s/$new_i/$i/g"
    if ! make -j6 check; then
      echo "__ '$i' '$new_i'"
      exit 1
    fi
  fi

  new_i=$(echo $i | sed -r 's/(_)([a-z])/\U\2/g')
  $FIND | $GREP | xargs grep -l "$i" | xargs sed -i "s/$i/$new_i/g"
  if ! make -j6 check; then
    $FIND | $GREP | xargs grep -l "$new_i" | xargs sed -i "s/$new_i/$i/g"
    if ! make -j6 check; then
      echo "camelCase '$i' '$new_i'"
      exit 2
    fi
  fi
done
```

seems like a wastes of builds, but this should only take about an hour on my machine and my rooms a little cold :smile:

It's simpler than figuring out all the exceptions to the rules we need to take into account and gets the job done.

I'll manually review and run the tests then to be sure.

ericcurtin:
And a `git-clang-format -f` of course

ericcurtin:
I wonder is it worth adding a just build the tests option to the make file, I kinda hacked that option together here.

freeekanayaka:
> I wonder is it worth adding a just build the tests option to the make file, I kinda hacked that option together here.

Sorry, I don't get this one, mind elaborating?

ericcurtin:
Sorry that wasn't clear at all

I do this (I basically comment out the execution of tests):

`sed -i "s/\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/#\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/g" Makefile`

so when I run `make check` I can build the tests without running them, handy for refactoring, checking if tests build and other tasks if you want to just run all the tests at the end on completion.

ericcurtin:
```
#!/bin/bash
  
git clean -fdx
git reset --hard origin/master

FIND="find . -name *.c -o -name *.h"
GREP="grep -v 'mu\|dqlite.h'"

autoreconf -i
./configure
CFLAGS=-O0
sed -i "s/\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/#\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/g" Makefile
make -j6 check

sed_change() {
  if [ "$i" != "$new_i" ]; then
    $FIND | $GREP | xargs grep -l "$i" | xargs sed -i "s/$i/$new_i/g"
  fi
}

sed_revert() {
  if [ "$i" != "$new_i" ]; then
    $FIND | $GREP | xargs grep -l "$new_i" | xargs sed -i "s/$new_i/$i/g"
  fi
}

start
for i in $($FIND | $GREP | xargs grep _ | tr -cs '[:alnum:]_#' '\n' | grep _ | sort | uniq);  do
  new_i=$(echo $i | sed "s/__/_/g")
  echo "'$i' '$new_i'"
  if  [[ "$i" =~ ^COMMAND_*|^conn_close_cb ]]; then
    continue
  fi

  if  [ "$i" != "$new_i" ]; then
    sed_change
    if ! make -j6 check; then
      sed_revert
      if ! make -j6 check; then
        echo "__ '$i' '$new_i'"
        exit 1
      fi
    fi
  fi

  i="$new_i"
  new_i=$(echo $i | sed -r 's/(_)([a-z])/\U\2/g')
  echo "'$i' '$new_i'"
  if  [ "$i" != "$new_i" ]; then
    sed_change
    if ! make -j6 check; then
      sed_revert
      if ! make -j6 check; then
        echo "camelCase '$i' '$new_i'"
        exit 2
      fi
    fi
  fi
done

git-clang-format -f
```

freeekanayaka:
> Sorry that wasn't clear at all
> 
> I do this (I basically comment out the execution of tests):
> 
> `sed -i "s/\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/#\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/g" Makefile`
> 
> so when I run `make check` I can build the tests without running them, handy for refactoring, checking if tests build and other tasks if you want to just run all the tests at the end on completion.

Oh, I see. Tests should be relatively fast to run, and they highlight any issue that might happen. `make -j8 check` should be fast. Alternatively you can have your own bash alias for `make -j8 test/unit/core test/integration/core test/integration/uv` or something like that (the `-j8` is what I use because I have 8 processors). What you are doing is also very peculiar, not a typical change/workflow. Normally I don't see a problem.

ericcurtin:
@freeekanayaka I think it's as you stated now :smile: 

The script that did the work below

```
#!/bin/bash

FIND="find . -name *.c -o -name *.h"
GREP="grep -v mu\|dqlite.h"

autoreconf -i
./configure
CFLAGS=-O0
sed -i "s/\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/#\$(MAKE) \$(AM_MAKEFLAGS) check-TESTS/g" Makefile
sed -i "s/-Wall/-Wall -Werror/g" Makefile
sed -i "s/-Wno-conversion/-Wno-conversion -Werror/g" Makefile
make -j16 check

sed_change() {
  if [ "$i" != "$new_i" ]; then
    $FIND | $GREP | xargs grep -l "$i" | xargs sed -i "s/$i/$new_i/g"
  fi

  git-clang-format -f
  git commit -a -m "s/$i/$new_i/g"
}

for i in $($FIND | $GREP | xargs grep _ | tr -cs '[:alnum:]_#' '\n' | grep _ | sort | uniq); do
  new_i=$(echo $i | sed "s/__/_/g")
  echo "'$i' '$new_i'"
  # execeptions don't change these, leading underscore used for many system
  # defines, page_size literal string not to be changed
  if  [[ "$i" =~ ^_|^page_size ]]; then
    continue
  fi

  if  [ "$i" != "$new_i" ]; then
    sed_change
    if make -j16 check; then
      git clang-format -f HEAD~1
      git commit -a --amend --no-edit
    else
      git reset --hard HEAD~1
      if ! make -j16 check; then
        echo "__ '$i' '$new_i'"
        exit 1
      fi
    fi
  fi

  i="$new_i"
  new_i=$(echo $i | sed -r 's/(_)([a-z])/\U\2/g')
  echo "'$i' '$new_i'"
  if  [ "$i" != "$new_i" ]; then
    sed_change
    if make -j16 check; then
      git clang-format -f HEAD~1
      git commit -a --amend --no-edit
    else
      git reset --hard HEAD~1
      if ! make -j16 check; then
        echo "camelCase '$i' '$new_i'"
        exit 2
      fi
    fi
  fi
done
```

ericcurtin:
We now have passing tests, these are the exceptions (not including anything that broke builds or tests):

`^_|^page_size|^sqlite|^foreign_keys|^journal_mode|^cache_size|^wal_autocheckpoint`

freeekanayaka:
There are still a few cases with a mixed style like:

```diff
-	rv = request_connect__decode(cursor, &request);
+	rv = requestConnect__decode(cursor, &request);
```

however these could be probably converted later on.

freeekanayaka:
Another case with mixed style is `struct dqlite_node` defined in `server.h`:

```C
struct dqlite_node
{
	pthread_t thread;                        /* Main run loop thread. */
	struct config config;                    /* Config values */
	struct sqlite3_vfs vfs;                  /* In-memory VFS */
	struct registry registry;                /* Databases */
	struct uv_loop_s loop;                   /* UV loop */
	struct raft_uv_transport raftTransport;  /* Raft libuv transport */
	struct raft_io raft_io;                  /* libuv I/O */
	struct raft_fsm raft_fsm;                /* dqlite FSM */
	sem_t ready;                             /* Server is ready */
	sem_t stopped;                           /* Notifiy loop stopped */
	pthread_mutex_t mutex;                   /* Access incoming queue */
	queue queue;                             /* Incoming connections */
	queue conns;                             /* Active connections */
	bool running;                            /* Loop is running */
	struct raft raft;                        /* Raft instance */
	struct uv_stream_s *listener;            /* Listening socket */
	struct uv_async_s stop;                  /* Trigger UV loop stop */
	struct uv_timer_s startup;               /* Unblock ready sem */
	struct uv_prepare_s monitor;             /* Raft state change monitor */
	int raft_state;                          /* Previous raft state */
	char *bind_address;                      /* Listen address */
	char errmsg[RAFT_ERRMSG_BUF_SIZE];       /* Last error occurred */
};
```

Some attributes are camelCase some other are snake_case. There are other similar occurrences in the branch I think. The downside that I see is that it might confuse a random reader.

ericcurtin:
> ```diff
> requestConnect__decode
> ```

Yeah these tokens must be treated differently to all the others in the script. I must grep for "##__" macros and ignore everything preceding and including the "##" and then my script can treat these tokens as normal and do another mass replace, while testing to ensure no builds get broken.

Ignoring ##__VA_ARGS__ also of course.

ericcurtin:
> Another case with mixed style is `struct dqlite_node` defined in `server.h`:
> 
> ```c
> struct dqlite_node
> {
> 	pthread_t thread;                        /* Main run loop thread. */
> 	struct config config;                    /* Config values */
> 	struct sqlite3_vfs vfs;                  /* In-memory VFS */
> 	struct registry registry;                /* Databases */
> 	struct uv_loop_s loop;                   /* UV loop */
> 	struct raft_uv_transport raftTransport;  /* Raft libuv transport */
> 	struct raft_io raft_io;                  /* libuv I/O */
> 	struct raft_fsm raft_fsm;                /* dqlite FSM */
> 	sem_t ready;                             /* Server is ready */
> 	sem_t stopped;                           /* Notifiy loop stopped */
> 	pthread_mutex_t mutex;                   /* Access incoming queue */
> 	queue queue;                             /* Incoming connections */
> 	queue conns;                             /* Active connections */
> 	bool running;                            /* Loop is running */
> 	struct raft raft;                        /* Raft instance */
> 	struct uv_stream_s *listener;            /* Listening socket */
> 	struct uv_async_s stop;                  /* Trigger UV loop stop */
> 	struct uv_timer_s startup;               /* Unblock ready sem */
> 	struct uv_prepare_s monitor;             /* Raft state change monitor */
> 	int raft_state;                          /* Previous raft state */
> 	char *bind_address;                      /* Listen address */
> 	char errmsg[RAFT_ERRMSG_BUF_SIZE];       /* Last error occurred */
> };
> ```
> 
> Some attributes are camelCase some other are snake_case. There are other similar occurrences in the branch I think. The downside that I see is that it might confuse a random reader.

This tends to occur when variable names match external datatype names too closely (although there are other cases, shadowing variables is another), these are a good example:

```
	struct raft_io raft_io;                  /* libuv I/O */
	struct raft_fsm raft_fsm;                /* dqlite FSM */
```

Probably best to manually change these kind of cases. If in doubt should we always favor camelCase?

freeekanayaka:
Yeah I think a manual pass would probably work best.

I don't have any opinion about camelCase vs snake_case when it comes to struct attributes, since those don't risk collisions, unless that struct is exported in a public `.h` file (e.g. dqlite.h or raft.h), in that case I'd stick to snake_case for consistency with classic C style. At this point maybe it would make sense to convert to camelCase everything which is not public (including struct attributes and local variables).

ericcurtin:
The changes related to macros with ## were added

ericcurtin:
Fixed that struct with the (partial) commits. Will look for structs in the overall diff when I get some time tomorrow.

ericcurtin:
ok I think it's pretty good, got rid of any inconsistencies in structs I saw in the:

`git diff origin/master`

output

src/client.c
```diff
@@ -56,31 +56,31 @@ int clientSendHandshake(struct client *c)
...
+#define REQUEST(LOWER, UPPER)                                     \
+	{                                                         \
+		struct message message;                           \
+		size_t n;                                         \
+		size_t n1;                                        \
+		size_t n2;                                        \
+		void *cursor;                                     \
+		ssize_t rv;                                       \
+		n1 = messageSizeof(&message);                     \
+		n2 = request_##LOWER##Sizeof(&request);           \
```
freeekanayaka, line 68: This is still mixed between snake_case and camelCase, it should be something like `request##CAPITAL##Sizeof(&request)`, where `CAPITAL` is the capital version of the message type name, replacing the `LOWER` version.

src/client.c
```diff
@@ -119,16 +119,16 @@ int clientSendHandshake(struct client *c)
...
-		if (rv != 0) {                                       \
-			return DQLITE_ERROR;                         \
-		}                                                    \
+#define DECODE(LOWER)                                              \
+	{                                                          \
+		int rv;                                            \
+		struct cursor cursor;                              \
+		cursor.p = bufferCursor(&c->read, 0);              \
+		cursor.cap = bufferOffset(&c->read);               \
+		rv = response_##LOWER##Decode(&cursor, &response); \
```
freeekanayaka, line 128: Same here, mixed snake_case and camelCase, probably needs to change `LOWER` to `CAPITAL` or something like that.

src/query.h
```diff
@@ -15,6 +15,6 @@
  * tuples, either until #SQLITE_DONE is returned or a full page of the given
  * buffer is filled.
  */
-int query__batch(sqlite3_stmt *stmt, struct buffer *buffer);
+int query_batch(sqlite3_stmt *stmt, struct buffer *buffer);
```
freeekanayaka, line 18: Since this is not a public symbol (exported in `dqlite.h`), but rather an internal one it should be `queryBatch`.

freeekanayaka:
Looks mostly good to me, I think I caught a couple of things, see above.

ericcurtin:
Thanks very much @freeekanayaka I appreciate it, I'll get to it this evening after work, or whenever I get a spare moment :+1:

src/client.c
ericcurtin: All these kind of ones were just fixed the the last 6 commits

src/query.h
ericcurtin: Fixed, there was a namespace clash here I had to fix first with s/queryBatch/gatewayQueryBatch/g

src/client.c
ericcurtin: Fixed

ericcurtin:
As a validation check, I checked to see if we can compile with g++/clang++, ignoring the type-safety warnings, we eventually hit errors:

```
In file included from src/registry.h:8,
                 from src/server.h:8,
                 from src/server.c:1:
src/server.c:453:17: error: ‘struct dqlite_node’ has no member named ‘conns’
  453 |  QUEUE_PUSH(&t->conns, &conn->queue);
      |                 ^~~~~
src/lib/queue.h:34:20: note: in definition of macro ‘QUEUE_PUSH’
   34 |   QUEUE_NEXT(e) = (q);           \
      |                    ^
src/server.c:453:17: error: ‘struct dqlite_node’ has no member named ‘conns’
  453 |  QUEUE_PUSH(&t->conns, &conn->queue);
      |                 ^~~~~
src/lib/queue.h:10:40: note: in definition of macro ‘QUEUE_PREV’
   10 | #define QUEUE_PREV(q) (*(queue **)&((*(q))[1]))
      |                                        ^
src/server.c:453:2: note: in expansion of macro ‘QUEUE_PUSH’
  453 |  QUEUE_PUSH(&t->conns, &conn->queue);
      |  ^~~~~~~~~~
src/server.c:453:17: error: ‘struct dqlite_node’ has no member named ‘conns’
  453 |  QUEUE_PUSH(&t->conns, &conn->queue);
      |                 ^~~~~
src/lib/queue.h:10:40: note: in definition of macro ‘QUEUE_PREV’
   10 | #define QUEUE_PREV(q) (*(queue **)&((*(q))[1]))
      |                                        ^
src/server.c:453:2: note: in expansion of macro ‘QUEUE_PUSH’
  453 |  QUEUE_PUSH(&t->conns, &conn->queue);
```

I think this is fine though, we only need the public header to be C++ compilable and the library to be C++ linkable, which it should be. We have a really low chance of namespace collision now. This is a slightly different issue anyway, if you even call it an issue at all.

ericcurtin:
Loads of files compiled with g++/clang++ fine also, I think this was in the later stages of compilation.

src/client.c
```diff
@@ -148,90 +148,90 @@ int clientSendOpen(struct client *c, const char *name)
...
 	request.sql = sql;
 	REQUEST(prepare, PREPARE);
 	return 0;
 }
 
-int clientRecvStmt(struct client *c, unsigned *stmt_id)
+int clientRecvStmt(struct client *c, unsigned *stmtId)
 {
-	struct response_stmt response;
+	struct responsestmt response;
```
freeekanayaka, line 168: This should be responseStmt.

src/client.c
```diff
@@ -148,90 +148,90 @@ int clientSendOpen(struct client *c, const char *name)
...
-	c->db_id = response.id;
+	c->dbId = response.id;
 	return 0;
 }
 
 int clientSendPrepare(struct client *c, const char *sql)
 {
-	struct request_prepare request;
-	request.db_id = c->db_id;
+	struct requestprepare request;
```
freeekanayaka, line 159: This should be requestPrepare.

src/client.c
```diff
@@ -314,15 +313,15 @@ int clientSendAssign(struct client *c, unsigned id, int role)
...
+	struct requestremove request;
 	request.id = id;
 	REQUEST(remove, REMOVE);
 	return 0;
 }
 
 int clientRecvEmpty(struct client *c)
 {
-	struct response_empty response;
+	struct responseempty response;
```
freeekanayaka, line 324: This should be responseEmpty.

src/client.c
```diff
@@ -314,15 +313,15 @@ int clientSendAssign(struct client *c, unsigned id, int role)
 
 int clientSendRemove(struct client *c, unsigned id)
 {
-	struct request_remove request;
+	struct requestremove request;
```
freeekanayaka, line 316: This should be requestRemove.

src/client.c
```diff
@@ -304,7 +303,7 @@ int clientSendAdd(struct client *c, unsigned id, const char *address)
 
 int clientSendAssign(struct client *c, unsigned id, int role)
 {
-	struct request_assign request;
+	struct requestassign request;
```
freeekanayaka, line 306: requestAssign

src/gateway.c
```diff
@@ -881,9 +882,9 @@ static int handle_cluster(struct handle *req, struct cursor *cursor)
 	START(cluster, servers);
 
 	response.n = g->raft->configuration.n;
-	cur = buffer__advance(req->buffer, response_servers__sizeof(&response));
+	cur = bufferAdvance(req->buffer, responseserversSizeof(&response));
```
freeekanayaka, line 885: Should be responseServerSizeOf

src/gateway.c
```diff
@@ -881,9 +882,9 @@ static int handle_cluster(struct handle *req, struct cursor *cursor)
...
 	assert(cur != NULL);
-	response_servers__encode(&response, &cur);
+	responseserversEncode(&response, &cur);
```
freeekanayaka, line 887: Should be responseServerEncode

src/gateway.c
```diff
@@ -902,7 +903,7 @@ void raftTransferCb(struct raft_transfer *r)
 {
 	struct gateway *g = r->data;
 	struct handle *req = g->req;
-	struct response_empty response;
+	struct responseempty response;
```
freeekanayaka, line 906: Should be responseEmpty

freeekanayaka (commented):
Hello, I went through the diff again, and I found a few more renames that would need a tweak.

ericcurtin:
I'm not 100% sure how to change to camelCase with request*, response*, etc. @freeekanayaka what prevents me from changing those is the ##LOWER macros. I'll certainly fix the conflicts though :smile:

ericcurtin:
Fixed the conflicts, I hope this branch doesn't cause any conflicts with any changes @freeekanayaka @brauner or others may have in progress :disappointed_relieved: 

After I get this complete, I'll probably focus my spare time back on my thesis, inotify-tools and a charity effort I'm involved in. Who knows maybe in a years time if I have more free time I may be able to contribute again, if I can find a neat use for dqlite, I'm sure I will eventually :smile:

ericcurtin:
More conflicts, I can fix again. Any thoughts on what it would take to get this merged?

PR #275, opened 2020-12-02
bugfixes

brauner:
Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

### DIFF
--- a/src/query.c
+++ b/src/query.c
@@ -112,6 +112,7 @@ int query__batch(sqlite3_stmt *stmt, struct buffer *buffer) {
 
 	/* Insert the column count */
 	cursor = buffer__advance(buffer, sizeof(uint64_t));
+	assert(cursor != NULL);
 	uint64__encode(&n64, &cursor);
 
 	/* Insert the column names */
--- a/src/server.c
+++ b/src/server.c
@@ -240,7 +240,7 @@ int dqlite_node_set_bind_address(dqlite_node *t, const char *address)
 	if (domain == AF_INET) {
 		t->bind_address = sqlite3_malloc((int)strlen(address));
 		if (t->bind_address == NULL) {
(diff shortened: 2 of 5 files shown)


src/query.c
```diff
@@ -112,6 +112,9 @@ int query__batch(sqlite3_stmt *stmt, struct buffer *buffer) {
 
 	/* Insert the column count */
 	cursor = buffer__advance(buffer, sizeof(uint64_t));
+	if (cursor == NULL) {
```
freeekanayaka, line 115: In this case I think the buffer passed here is guaranteed to have at least 8 bytes available, so perhaps this could be turned into:

```
assert(cursor != NULL);
```

but it's also fine this way.

src/server.c
```diff
@@ -596,6 +600,7 @@ int dqlite_node_start(dqlite_node *t)
 	}
 
 	if (!taskReady(t)) {
+		dqlite_node_stop(t);
```
freeekanayaka, line 603: Why do we need this call? As far as I can tell `taskReady()` can only return `false` if `taskRun` calls `sem_post` and in that case it will return and make the thread exit. In that case the UV loop won't be running and the call to `uv_async_send` inside `dqlite_node_stop` will fail.

Maybe instead of calling `dqlite_node_stop` you could simply call `pthread_join`, although it shouldn't be strictly necessary.

src/query.c
brauner: No, I'm fine with the assert.

src/server.c
brauner: I can drop this commit for now so we're not blocking other stuff in here which might be valuable.

freeekanayaka:
Looks good, just waiting for Travis to run the tests again.

PR #285, opened 2021-02-17
Support compile with clang on BSD

rabits:
A couple of changes to make the code compile on BSD systems with clang


stgraber:
Needs a few tweaks to make travis happy :)

stgraber:
Can you rebase this into a single commit?

rabits:
Oh, sure)

stgraber:
What's defining those variables?
I'm not seeing matching autotools changes which would detect and set:
 - HAVE_SYS_UCRED_H
 - HAVE_DECL_LOCAL_PEERCRED
 - HAVE_DECL_SO_PEERCRED

rabits:
Good catch! My bad, will adjust the configs

rabits:
Yep, actually need to rewrite the bsd-part, it compiled just because definitions was not set...

rabits:
@stgraber could you please suggest how to test it properly? I use go-dqlite, so where I will see it's working or not?

src/server.c
```diff
@@ -211,10 +211,11 @@ int dqlite_node_set_bind_address(dqlite_node *t, const char *address)
 		len += sizeof(sa_family_t);
 		addr = (struct sockaddr *)&addr_un;
 	}
-	fd = socket(domain, SOCK_STREAM | SOCK_CLOEXEC, 0);
+	fd = socket(domain, SOCK_STREAM, 0);
 	if (fd == -1) {
 		return DQLITE_ERROR;
 	}
+	fcntl(fd, O_CLOEXEC);
```
stgraber, line 218: Should add error checking for this.

stgraber:
I'm not all that familiar with the internals of dqlite, I certainly understand the goal of this logic but am not sure what's the best way to test it. I tried connecting to a number of the pipes held by LXD here, but that's not going anywhere, presumably because those are already connected, so I'm never hitting the code that would reject the connection (just stuck waiting for accept basically).

@freeekanayaka would definitely be able to assist there, maybe @MathieuBordere somehow already looked at that part of the codebase too but he's mostly been focused on raft issues so far ;)

src/server.c
rabits: Yep, fixed

MathieuBordere:
> I'm not all that familiar with the internals of dqlite, I certainly understand the goal of this logic but am not sure what's the best way to test it. I tried connecting to a number of the pipes held by LXD here, but that's not going anywhere, presumably because those are already connected, so I'm never hitting the code that would reject the connection (just stuck waiting for accept basically).
> 
> @freeekanayaka would definitely be able to assist there, maybe @MathieuBordere somehow already looked at that part of the codebase too but he's mostly been focused on raft issues so far ;)

Not too familiar with this part of the codebase, will have a look eventually if this stays open :-)

freeekanayaka:
If you run the `go-dqlite` tests (in particular the ones in the `go-dqlite/app` sub-package`), that should be enough to test the part that you are modifying. Essentially the code path that you touch in this PR gets triggered if you tell the dqlite engine to listen to an abstract unix socket rather to a TCP port (so for example you can proxy incoming connections with TLS). In that case you want to make sure that incoming connections are coming from the same process (e.g. a Go process performing the TLS proxy dance), otherwise any process on the host would be able to connect.

rabits:
Thanks, folks - figured out there is other changes required after I compiled go-dqlite demo. Will continue to test and come back with a solution)

rabits:
Ok, that was a journey. To run the node and walk through demo it will also require libraft changes, and some other adjustments, but ended up with the networking issues - seems the custom aligned_alloc is not working well and causing a mess in the communication protocol... Will prepare the changes and try to find some help with debugging.

PR #290, opened 2021-04-11
Fixed a couple of issues with unix sockets on MacOS

rabits:
During testing of TLS I found that MacOS don't support abstract sockets and there is a bug I introduced in https://github.com/canonical/dqlite/pull/285 - so moved to the usual sockets for Mac and fixed checking of the client process PID.

The other patches in series:
* canonical/go-dqlite#132
* #285
* #288
* canonical/raft#173

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -220,7 +220,14 @@ int dqlite_node_set_bind_address(dqlite_node *t, const char *address)
 			/* Auto bind */
 			len = 0;
 		} else {
-			strcpy(addr_un.sun_path + 1, address + 1);
+#if defined(__linux__)
+			// Linux abstract socket requires \0 in sun_path[0]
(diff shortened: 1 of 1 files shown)


src/server.c
```diff
@@ -220,7 +220,15 @@ int dqlite_node_set_bind_address(dqlite_node *t, const char *address)
 			/* Auto bind */
 			len = 0;
 		} else {
+#if defined(__linux__)
+			// Linux abstract socket requires \0 in sun_path[0]
 			strcpy(addr_un.sun_path + 1, address + 1);
```
rabits, line 225: Folks, maybe to use strncpy/strcat here? For now I left it as is, but maybe we need little more fortification?

MathieuBordere (approved):

Issue #292, opened 2021-04-22
Assertion `db->follower == NULL' failed when database reaches certain size


MathieuBordere:
src/db.c:40: db__open_follower: Assertion `db->follower == NULL' failed

Encountered while doing a lot of writes and the raft snapshot reaches a certain size, in this case the last raft snapshot taken is
`-rw------- 1 mathieu mathieu  1,1G apr 22 14:28 snapshot-1-261120-10839878`

https://github.com/canonical/dqlite/blob/761706678edc42f9bd163e31b933ace72576883b/src/fsm.c#L172 This line fails.

MathieuBordere:
Looks to be caused by https://github.com/canonical/dqlite/blob/761706678edc42f9bd163e31b933ace72576883b/src/fsm.c#L187
returning `778` that is translated to  `disk I/O error` by `sqlite3_errmsg`

MathieuBordere:
@freeekanayaka Just picking your mind here, are there any hard limits on DB size around 1024MB that you are aware of?

freeekanayaka:
> @freeekanayaka Just picking your brain here, are there any hard limits on DB size around 1024MB that you are aware of?

Nothing comes to my mind, but there's surely something fishy going on..

MathieuBordere:
https://github.com/canonical/dqlite/blob/761706678edc42f9bd163e31b933ace72576883b/src/vfs.c#L356 fails with these parameters.
vfsDatabaseGetPage pgno:262146 page_size:4096 n_pages:262144

MathieuBordere:
@freeekanayaka I have a fix the problem if I drop this assumption.
https://github.com/canonical/dqlite/blob/761706678edc42f9bd163e31b933ace72576883b/src/vfs.c#L354
Do you have some more insight into why Sqlite appears to violate that assumption?

freeekanayaka:
> @freeekanayaka I have a fix for the problem if I drop this assumption.
> 
> https://github.com/canonical/dqlite/blob/761706678edc42f9bd163e31b933ace72576883b/src/vfs.c#L354
> 
> Do you have some more insight into why SQLite appears to violate that assumption?

I can't claim I'm 100% sure, but I think the reason is that SQLite knows what pages are there and at most it creates a new page by getting a new one past the last one. I don't quite see a reason for skipping pages. However, since you say that relaxing that assumption seems to fix the problem, I'd suggest to create a C-only reproducer and use gdb to see where SQLite is doing that and why.

[...]

PR #294, opened 2021-04-26
vfs: Allow skipping a page when writing past PENDING_BYTE

MathieuBordere:
Fixes https://github.com/canonical/dqlite/issues/292.

@freeekanayaka you have an idea how I could fit this in the test suite? https://www.sqlite.org/c3ref/test_control.html can be used to adapt the `pending_byte` value but it is not a stable API, so am a bit reluctant to use it.

### DIFF
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -340,6 +340,32 @@ static void vfsDatabaseDestroy(struct vfsDatabase *d)
 	sqlite3_free(d);
 }
 
+/*
+ * Comment copied entirely for sqlite source code, it is safe to assume
+ * the value 0x40000000 will never change. dq_sqlite_pending_byte is global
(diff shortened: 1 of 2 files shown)


freeekanayaka:
I'd just use that unstable API. It is unlikely to change, and even if it does, it will at most break the tests and in that case you will probably be able to fix it (I would guess that test API just get changed slightly but the functionality still be there in some way).

MathieuBordere:
@freeekanayaka The test I added covers the code that creates a page for the pending_byte, however I'm a bit surprised that I have to insert 65536 entries in order to hit it, 32768 is not enough.
`vfsDatabaseGetPage` is only called once with parameters `page_size:512 pgno:1` in that case. Do you know why that happens?

freeekanayaka (approved):
LGTM

PR #316, opened 2021-07-26
Recover v2

MathieuBordere:
Add v2 of recover command that includes the server roles.

### DIFF
--- a/include/dqlite.h
+++ b/include/dqlite.h
@@ -3,6 +3,7 @@
 
 #include <sqlite3.h>
 #include <stddef.h>
+#include <stdint.h>
 
 /**
  * Version.
@@ -187,7 +188,22 @@ struct dqlite_node_info
 };
 typedef struct dqlite_node_info dqlite_node_info;
 
+/* Defined to be an extensible struct, future additions to this struct should be
+ * 64-bits wide and 0 should not be used as a valid value. */
+struct dqlite_node_info_ext
+{
+        uint64_t size; /* The size of this struct */
+        uint64_t id; /* dqlite_node_id */
+        uint64_t address;
+        uint64_t dqlite_role;
+};
+typedef struct dqlite_node_info_ext dqlite_node_info_ext;
+#define DQLITE_NODE_INFO_EXT_SZ_ORIG 32U /* (4 * 64) / 8 */
+
 /**
+ * !!! Deprecated, use `dqlite_node_recover_ext` instead which also includes
+ * dqlite roles. !!!
+ *
  * Force recovering a dqlite node which is part of a cluster whose majority of
  * nodes have died, and therefore has become unavailable.
  *
@@ -212,6 +228,32 @@ typedef struct dqlite_node_info dqlite_node_info;
  * 6. Restart all nodes.
  */
 int dqlite_node_recover(dqlite_node *n, dqlite_node_info infos[], int n_info);
+
+/**
+ * Force recovering a dqlite node which is part of a cluster whose majority of
+ * nodes have died, and therefore has become unavailable.
+ *
+ * In order for this operation to be safe you must follow these steps:
+ *
+ * 1. Make sure no dqlite node in the cluster is running.
+ *
+ * 2. Identify all dqlite nodes that have survived and that you want to be part
+ *    of the recovered cluster.
+ *
+ * 3. Among the survived dqlite nodes, find the one with the most up-to-date
+ *    raft term and log.
+ *
+ * 4. Invoke @dqlite_node_recover_ext exactly one time, on the node you found in
+ *    step 3, and pass it an array of #dqlite_node_info filled with the IDs,
+ *    addresses and roles of the survived nodes, including the one being recovered.
+ *
+ * 5. Copy the data directory of the node you ran @dqlite_node_recover_ext on to all
+ *    other non-dead nodes in the cluster, replacing their current data
+ *    directory.
+ *
+ * 6. Restart all nodes.
+ */
+int dqlite_node_recover_ext(dqlite_node *n, dqlite_node_info_ext infos[], int n_info);
 
 /**
  * Return a human-readable description of the last error occurred.
--- a/src/server.c
+++ b/src/server.c
@@ -9,7 +9,10 @@
 #include "fsm.h"
 #include "lib/assert.h"
 #include "logger.h"
+#include "protocol.h"
+#include "translate.h"
 #include "transport.h"
+#include "utils.h"
 #include "vfs.h"
 
 /* Special ID for the bootstrap node. Equals to raft_digest("1", 0). */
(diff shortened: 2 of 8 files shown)


include/dqlite.h
```diff
@@ -187,7 +187,18 @@ struct dqlite_node_info
 };
 typedef struct dqlite_node_info dqlite_node_info;
 
+struct dqlite_node_info_v2
+{
+	dqlite_node_id id;
+	const char *address;
+	int dqlite_role;
+};
+typedef struct dqlite_node_info_v2 dqlite_node_info_v2;
```
stgraber, line 196: Can't you add the field to the existing `dqlite_node_info` struct and put a size check or similar in place to detect someone passing you the older struct?

I think @brauner did a bunch of that kind of stuff in liblxc where we need to expand structs and don't want to needlessly have to fork them and duplicate symbols.

stgraber: We'll want @brauner to comment here. For liblxc I believe it's a mix of memset and some checks so if we're fed a smaller struct, we're not reading garbage and can tell if the data was provided or not.

If this isn't doable with the current symbol here, I'd at least argue that the new symbol we're adding should allow for further changes to the struct without then needing a v3.

brauner: It is possible to version a C struct by size and is pretty common. Aleksa and I introduced a dedicated API for this into the kernel a while back. For this to work nicely the struct should usually be aligned at a 64 bit boundary. That ensures that the compiler doesn't add padding. I'm quoting from the document a document we wrote back then:

```rst
Advanced extensibility: extensible structs
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

For more sophisticated system calls that involve a larger number of arguments,
it's preferred to encapsulate the majority of the arguments into an extensible
structure that is passed in by pointer.

Extensible structs are versioned by their size.  For any new non-flag based
extension a new field has to be added to the end of the extensible struct.  The
zero value of the new field must not have any meaning so the system call can
continue to display the old behavior.

Extensible struct system calls can and should use the dedicated
``copy_struct_from_user`` API which enforces the following behavior:

 - The kernel will reject any size that is smaller than the initially supported
   size of the extensible struct.
 - If the extensible struct size the kernel knows about is equal to the size
   passed in from userspace then ``copy_struct_from_user`` will copy the struct
   verbatim.
 - If the extensible struct size the kernel knows about is larger than the size
   passed in from userspace the kernel will copy the size userspace indicated
   and treat all additional extensions it knows about as zero.
 - If the extensible struct size the kernel knows about is smaller than the
   size passed in from userspace the kernel will copy the number of bytes it
   knows about and verify that all trailing bytes are zero.  If non-zero bytes
   are present the kernel returns ``E2BIG``.  While not an intuitive error code
   at first, ``E2BIG`` means that the argument list is too long.

Early examples for extensible struct system calls include
:manpage:`perf_event_open(2)` and :manpage:`sched_setattr(2)`.  These system
calls implement mostly similar behavior even before the introduction of
``copy_struct_from_user`` but have since been switched over to it.  Newer
examples include :manpage:`clone3(2)` and :manpage:`openat2(2)`.

The size associated with an extensible struct can either be passed as
a separate argument in the system call as is e.g. done for :manpage:`clone3(2)`
and :manpage:`openat2(2)`.  Alternatively, the size can be passed as the first
field in the extensible struct as is e.g. done for :manpage:`sched_setattr(2)`.

Any struct passed from userspace to the kernel and especially extensible
structs must ensure that they are correctly padded.  This ensures that no data
can be leaked on accident or on purpose by an attacker from the kernel.  The
easiest way to ensure that a struct is correctly padded is to always use 64 bit
fields::

    struct sys_foo_args {
        __aligned_u64 arg1;
        __aligned_u64 arg2;
        __aligned_u64 arg3;
        __aligned_u64 arg4;
        __aligned_u64 arg5;
    };

System calls that need to worry about the size of their extensible structs or
need fields to be of a specific size can rely on careful manual struct
packing::

    struct sys_foo_args {
        __u32 arg1;
        __u16 arg2;
        __u16 arg3;
        __u32 arg4;
        __u32 arg5;
        __u64 arg6;
    };

(There are tools such as ``pahole`` available that allow to check whether
a struct is correctly padded!)

Note that in contrast to flag arguments passed as register arguments flag
arguments in extensible structures can be 64 bit wide.  As with simple
flag-only system calls, the system call needs to verify any unknown values for
flag-like fields in the passed struct are zeroed.
```

stgraber (approved):
Looks good to me, but I'll rely on @brauner on this one!

include/dqlite.h
```diff
@@ -187,7 +188,22 @@ struct dqlite_node_info
...
 typedef struct dqlite_node_info dqlite_node_info;
 
+/* Defined to be an extensible struct, future additions to this struct should be
+ * 64-bits wide and 0 should not be used as a valid value. */
+struct dqlite_node_info_ext
+{
+        uint64_t size; /* The size of this struct */
+        uint64_t id; /* dqlite_node_id */
+        uint64_t address;
+        uint64_t dqlite_role;
```
brauner, line 198: I usually prefer to use `__u64` kernel types but this should be fine too.

src/server.c
```diff
@@ -726,16 +728,75 @@ int dqlite_node_stop(dqlite_node *d)
...
+    int ret;
+
+    struct dqlite_node_info_ext *infos_ext = calloc((size_t)n_info, sizeof(*infos_ext));
+    if (infos_ext == NULL) {
+        return DQLITE_NOMEM;
+    }
+    for (i = 0; i < n_info; i++) {
+        infos_ext[i].size = sizeof(*infos_ext);
+        infos_ext[i].id = infos[i].id;
+        infos_ext[i].address = (uint64_t)(uintptr_t)infos[i].address;
```
brauner, line 742: In LXC we have macros for this like `#define PTR_TO_UINT64(p) ((uint64_t)((uintptr_t)(p)))`. Might come in handy here.

src/server.c
```diff
@@ -726,16 +728,75 @@ int dqlite_node_stop(dqlite_node *d)
...
+{
+    /* Reject any size smaller than the original definition of the extensible
+     * struct. */
+    if (info->size < DQLITE_NODE_INFO_EXT_SZ_ORIG) {
+        return false;
+    }
+
+    /* Require 8 byte allignment */
+    if (info->size % sizeof(uint64_t)) {
+        return false;
```
brauner, line 761: Fair. :)

brauner (approved):
Looks good overall.

src/server.c
MathieuBordere: Thanks, that should indeed be more readable!

MathieuBordere:
Checking why the tests hung on Ubuntu 20.04 with clang, probably not related to this PR, but checking nonetheless.

Issue #327, opened 2021-10-11
Removing LXD leader from 2 member cluster sometimes causes dqlite assert


tomponline:
This occurs sometimes during the `clustering config remove leader` test:

https://jenkins.linuxcontainers.org/job/lxd-github-pull-test/4890/arch=arm64,backend=dir,compiler=golang-1.13/consoleFull#console-section-34

```
lxc cluster rm node1 --verbose
INFO[10-11|11:51:11] Changing local dqlite raft role          address=10.1.1.102:8443 role=voter id=2
INFO[10-11|11:51:11] Sending member change notification heartbeat to all members address=10.1.1.102:8443
INFO[10-11|11:51:11] Partial member list heartbeat received, skipping full update 
INFO[10-11|11:51:11] Deleting member from cluster             name=node1 force=0
INFO[10-11|11:51:11] Remove node from dqlite raft cluster     address=10.1.1.101:8443 id=1
lxd: src/leader.c:143: leader__close: Assertion `rc == 0' failed.
Error: Delete https://10.1.1.101:8443/1.0/cluster/members/node1: EOF
WARN[10-11|11:51:11] Dqlite: attempt 1: server 10.1.1.101:8443: dial: Failed to connect to HTTP endpoint: dial tcp 10.1.1.101:8443: connect: connection refused 
```

The `Deleting member from cluster` log line shown is here:

https://github.com/lxc/lxd/blob/0b5562262a05ab994a43909e39e30fee1755d6d3/lxd/api_cluster.go#L1616

PR #328, opened 2021-10-11
gateway: Finalize stmt when `raft_barrier` fails

MathieuBordere:
`g->stmt` is prepared here https://github.com/canonical/dqlite/blob/d5694bb937c98591911e84407ec77d64e96e04a0/src/gateway.c#L546, but never finalized when `leader__barrier` fails. This could lead to resource leaks and possibly triggering the assert in https://github.com/canonical/dqlite/blob/d5694bb937c98591911e84407ec77d64e96e04a0/src/leader.c#L143.

Possibly resolves https://github.com/canonical/dqlite/issues/327


freeekanayaka:
Good catch, it's probably that.

Issue #349, opened 2022-01-10
Segmentation fault when running invalid queries


neoaggelos:
### Summary

Empty query inputs can cause the dqlite server to segfault.

### Description

Found out about this while digging around a dqlite cluster. I am using [go-dqlite](https://github.com/canonical/go-dqlite), but the problem seems unrelated to the client language.

When attempting to execute a bad query, the dqlite server crashes with the following output:

```bash
main: src/gateway.c:398: query_barrier_cb: Assertion `stmt != NULL' failed.
Aborted (core dumped)
```

### Replicate

- Start a dqlite cluster
- Connect using go-client.
- Execute `db.Query("")`. Actually, with some quick testing I found that multiple inputs can cause the segmentation fault, e.g. `" "`, `"    "`, `" --"`.

This is on an empty database, so no idea if the crash corrupts any data. 

### Expected Behaviour

Obviously, a correct application would not ever submit a query like this. However, a single client submitting queries should not be able to crash the database.

PR #353, opened 2022-02-16
gateway: Handle empty query statements

MathieuBordere:
Fixes https://github.com/canonical/dqlite/issues/349.

After this PR, when we receive empty SQL statements, we will return a failure when a `prepare` or `query_sql` is requested.
`exec_sql` will succeed when presented with an empty SQL statement.

I could have returned success for a `prepare` request containing an empty statement, but there is no real convention what the `stmt_id` should be in that case, I thought a failure is more reasonable in this case. Same for `query_sql`, I could have returned an empty ROWS response instead of a failure.

What do you think @freeekanayaka ?


freeekanayaka:
I think your approach is fine. It's kind of an edge-case that shouldn't really happen in a "normal" situation, so as long as we don't blow up badly I'd say everything is reasonable.

PR #354, opened 2022-03-01
Close leader connection

### LINKED ISSUE #355: src/vfs.c:2371: vfsInvalidateWalIndexHeader: Assertion `shm->shared[i] == 0' failed
See this in a LXD jenkins test `clustering handover` run:

https://jenkins.linuxcontainers.org/job/lxd-github-pull-test/6445/arch=amd64,backend=random,compiler=golang-1.18,section=cluster/consoleFull#console-section-18

```
Started members 1 and 2
Shutting down cluster members 2 and 3...
INFO[03-24|11:25:40] Asked to shutdown by API                 force=false
INFO[03-24|11:25:40] Starting shutdown sequence               signal="power failure"
INFO[03-24|11:25:40] Asked to shutdown by API                 force=false
INFO[03-24|11:25:40] Starting shutdown sequence               signal="power failure"
INFO[03-24|11:25:40] Removed inactive member event listener client local=10.1.1.104:8443 remote=10.1.1.102:8443
INFO[03-24|11:25:40] Transferring leadership                  address=10.1.1.103:8443
WARN[03-24|11:25:40] Rejecting heartbeat request as shutting down 
WARN[03-24|11:25:40] Failed heartbeat                         remote=10.1.1.102:8443 err="Heartbeat request failed with status: 503 Service Unavailable"
INFO[03-24|11:25:40] Handing over cluster member role         address=10.1.1.103:8443
INFO[03-24|11:25:40] Dqlite proxy stopped                     name=dqlite local=10.1.1.103:8443 remote=10.1.1.103:55050
INFO[03-24|11:25:40] Dqlite connected outbound                name=dqlite local=10.1.1.103:39524 remote=10.1.1.101:8443
INFO[03-24|11:25:40] Dqlite connected outbound                name=raft local=10.1.1.102:48952 remote=10.1.1.104:8443
INFO[03-24|11:25:40] Dqlite proxy started                     name=dqlite local=10.1.1.101:8443 remote=10.1.1.1:39524
INFO[03-24|11:25:40] Dqlite proxy started                     name=dqlite local=10.1.1.104:8443 remote=10.1.1.1:48952
INFO[03-24|11:25:40] Dqlite proxy started                     name=raft local=10.1.1.102:48952 remote=10.1.1.104:8443
INFO[03-24|11:25:40] Dqlite proxy stopped                     name=dqlite local=10.1.1.101:8443 remote=10.1.1.1:39524
WARN[03-24|11:25:40] Dqlite: attempt 1: server 10.1.1.101:8443: no known leader 
INFO[03-24|11:25:40] Dqlite connected outbound                remote=10.1.1.102:8443 name=dqlite local=10.1.1.103:55392
INFO[03-24|11:25:40] Dqlite proxy started                     name=dqlite local=10.1.1.102:8443 remote=10.1.1.1:55392
INFO[03-24|11:25:40] Dqlite connected outbound                local=10.1.1.104:54662 remote=10.1.1.102:8443 name=raft
INFO[03-24|11:25:40] Dqlite proxy started                     local=10.1.1.102:8443 remote=10.1.1.1:54662 name=dqlite
INFO[03-24|11:25:40] Dqlite proxy started                     name=raft local=10.1.1.104:54662 remote=10.1.1.102:8443
lxd: src/vfs.c:2371: vfsInvalidateWalIndexHeader: Assertion `shm->shared[i] == 0' failed.
Error: Put "http://unix.socket/internal/shutdown?force=false": EOF
```

MathieuBordere:
I'm trying to resolve an issue where an unfinished transaction is causing issues after a leadership transfer, the scenario is more or less as follows:

- Node 1 is leader, Node 2 is follower 
- Node 1 receives BEGIN - QUERY XXXX - ...  (notice that it doesn't receive a COMMIT message)
- Node 1 transfers leadership to Node 2, or loses leadership in some way
- Node 2 writes some data to the db
- Node 1 receives the frames, from Node 2, tries to apply them to the DB and triggers an Assert due to the pending transaction in the leader connection from step 2 that is holding locks. https://github.com/canonical/dqlite/blob/f20941be3b2a86f83ebaec1aa7dabfc4e60ac170/src/vfs.c#L2371

The `transferPendingTransaction` test tests this scenario.

This PR is just to ask for a bit of feedback on the possible solutions:
- Close the `conn` after the `monitor_cb` notices that a node has lost leadership, this will also close the sqlite3 leader database connection, this will look to the user like the node went down somehow while it isn't, but the go sql driver will detect this and just open a new connection.
- edit: opted for closing of `conn`, it leads to cleaner code. ~~This PR: Close the sqlite3 leader database connection only in the `monitor_cb`, this requires a user to reissue an `Open` call. I could also reopen the leader connection myself but I'm a bit hesitant to do it. If somehow the thread that was responsible for the incomplete transaction wakes up again and Node 1 has again obtained leadership, and the rest of the transaction would contain writes, than those writes will be applied to the DB but the final COMMIT message would fail because the new leader connection does not know about the BEGIN message.~~

Not reopening the sqlite3 leader connection ourselves will break some clients I think because they don't expect to reopen the DB connection.

Can you think of any other approaches @freeekanayaka that would be viable ?

Fixes #355

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -572,27 +572,21 @@ static void monitor_cb(uv_prepare_t *monitor)
 {
 	struct dqlite_node *d = monitor->data;
 	int state = raft_state(&d->raft);
-	/*
 	queue *head;
 	struct conn *conn;
-	*/
 
 	if (state == RAFT_UNAVAILABLE) {
 		return;
 	}
 
-	/* TODO: we should shutdown clients that are performing SQL requests,
-	 * but not the ones which are doing management-requests, such as
-	 * transfer leadership.  */
-	/*
 	if (d->raft_state == RAFT_LEADER && state != RAFT_LEADER) {
+		tracef("leadership lost");
 		QUEUE__FOREACH(head, &d->conns)
 		{
 			conn = QUEUE__DATA(head, struct conn, queue);
-			conn__stop(conn);
+			conn__leadership_lost(conn);
 		}
 	}
-	*/
 
 	d->raft_state = state;
 }
--- a/src/client.c
+++ b/src/client.c
@@ -365,6 +365,15 @@ int clientSendRemove(struct client *c, unsigned id)
 	return 0;
 }
 
+int clientSendTransfer(struct client *c, unsigned id)
+{
+	tracef("client send transfer fd %d id %u", c->fd, id);
+	struct request_transfer request;
+	request.id = id;
+	REQUEST(transfer, TRANSFER);
+	return 0;
+}
+
 int clientRecvEmpty(struct client *c)
 {
 	tracef("client recv empty fd %d", c->fd);
--- a/src/client.h
+++ b/src/client.h
@@ -79,6 +79,9 @@ int clientSendAssign(struct client *c, unsigned id, int role);
 /* Send a request to remove a server from the cluster. */
 int clientSendRemove(struct client *c, unsigned id);
 
+/* Send a request to transfer leadership to node with id `id`. */
+int clientSendTransfer(struct client *c, unsigned id);
+
 /* Receive an empty response. */
 int clientRecvEmpty(struct client *c);
 
--- a/src/conn.c
+++ b/src/conn.c
@@ -333,3 +333,19 @@ void conn__stop(struct conn *c)
 	gateway__close(&c->gateway);
 	transport__close(&c->transport, closeCb);
 }
+
+/**
+ * Close any connection that has an opened SQLite leader connection after
+ * leadership has been lost.
+ */
+void conn__leadership_lost(struct conn *c)
+{
+	tracef("conn leadership lost");
+	struct gateway *g;
+	g = &c->gateway;
+	if (g == NULL || g->leader == NULL) {
+		return;
+	}
+
+	conn__stop(c);
+}
--- a/src/conn.h
+++ b/src/conn.h
@@ -58,4 +58,9 @@ int conn__start(struct conn *c,
  */
 void conn__stop(struct conn *c);
 
+/**
+ * Called in case Raft leadership is lost.
+ */
+void conn__leadership_lost(struct conn *c);
+
 #endif /* DQLITE_CONN_H_ */
--- a/test/integration/test_fsm.c
+++ b/test/integration/test_fsm.c
@@ -188,8 +188,7 @@ TEST(fsm, snapshotHeapFaultTwoDB, setUp, tearDown, 0, NULL)
 	EXEC(stmt_id, &last_insert_id, &rows_affected);
 
 	/* Close and reopen the client and open a second database */
-	rv = test_server_client_reconnect(&f->servers[0]);
-	munit_assert_int(rv, ==, 0);
+	test_server_client_reconnect(&f->servers[0], &f->servers[0].client);
 
 	HANDSHAKE;
 	OPEN_NAME("test2");
@@ -253,8 +252,7 @@ TEST(fsm, snapshotNewDbAddedBeforeFinalize, setUp, tearDown, 0, NULL)
 
 	/* Close and reopen the client and open a second database,
 	 * and ensure finalize succeeds. */
-	rv = test_server_client_reconnect(&f->servers[0]);
-	munit_assert_int(rv, ==, 0);
+	test_server_client_reconnect(&f->servers[0], &f->servers[0].client);
 
 	HANDSHAKE;
 	OPEN_NAME("test2");
@@ -391,8 +389,7 @@ TEST(fsm, snapshotRestore, setUp, tearDown, 0, num_writes_params)
 	munit_assert_int(rv, ==, 0);
 
 	/* Table is there on fresh connection. */
-	rv = test_server_client_reconnect(&f->servers[0]);
-	munit_assert_int(rv, ==, 0);
+	test_server_client_reconnect(&f->servers[0], &f->servers[0].client);
 	HANDSHAKE;
 	OPEN;
 	PREPARE("SELECT COUNT(*) from test", &stmt_id);
@@ -469,8 +466,7 @@ TEST(fsm, snapshotRestoreMultipleDBs, setUp, tearDown, 0, NULL)
 	PREPARE("INSERT INTO test(n) VALUES(1)", &stmt_id);
 	EXEC(stmt_id, &last_insert_id, &rows_affected);
 
-	rv = test_server_client_reconnect(&f->servers[0]);
-	munit_assert_int(rv, ==, 0);
+	test_server_client_reconnect(&f->servers[0], &f->servers[0].client);
 	HANDSHAKE;
 	OPEN_NAME("test2");
 	PREPARE("CREATE TABLE test2a (n INT)", &stmt_id);
@@ -498,8 +494,7 @@ TEST(fsm, snapshotRestoreMultipleDBs, setUp, tearDown, 0, NULL)
 	munit_assert_int(rv, ==, 0);
 
 	/* Reopen connection */
-	rv = test_server_client_reconnect(&f->servers[0]);
-	munit_assert_int(rv, ==, 0);
+	test_server_client_reconnect(&f->servers[0], &f->servers[0].client);
 	HANDSHAKE;
 	OPEN_NAME("test2");
 
@@ -514,8 +509,7 @@ TEST(fsm, snapshotRestoreMultipleDBs, setUp, tearDown, 0, NULL)
 	munit_assert_string_equal(msg, "no such table: test2b");
 
 	/* Table is there on first DB */
-	rv = test_server_client_reconnect(&f->servers[0]);
-	munit_assert_int(rv, ==, 0);
+	test_server_client_reconnect(&f->servers[0], &f->servers[0].client);
 	HANDSHAKE;
 	OPEN_NAME("test");
 	PREPARE("SELECT * from test", &stmt_id);
--- a/test/integration/test_membership.c
+++ b/test/integration/test_membership.c
@@ -17,7 +17,8 @@
 #define N_SERVERS 3
 #define FIXTURE                                \
 	struct test_server servers[N_SERVERS]; \
-	struct client *client
+	struct client *client;                 \
+	struct rows rows;
 
 #define SETUP                                                 \
 	unsigned i_;                                          \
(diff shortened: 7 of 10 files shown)


freeekanayaka:
I don't recall exactly the details, but I believe in this situation Node 1 should "rollback" the pending transaction when it receives frames from node 2 (noticing that it's not the leader anymore). I thought there was code for this situation already.

stgraber:
Going to let @freeekanayaka take a look into this one

src/server.c
```diff
@@ -572,27 +572,33 @@ static void monitor_cb(uv_prepare_t *monitor)
...
 		QUEUE__FOREACH(head, &d->conns)
 		{
 			conn = QUEUE__DATA(head, struct conn, queue);
-			conn__stop(conn);
+			g = &conn->gateway;
+			if (g == NULL || g->leader == NULL) {
+				continue;
+			}
+
+			if (g->req != NULL && g->stmt == NULL) {
```
freeekanayaka, line 593: Are we really sure that we don't want to have this check read:

```
if (g->req != NULL && g->req->type == REQUEST_TRANSFER) {
```

or add a sort of "whitelist" of request types that we don't want to kill immediately.

I didn't really think too much about the implications of not shutting down other management requests, but perhaps just going through all of them and considering the implications of this change would be good (if you didn't already).

freeekanayaka (approved):
It seems good to me, just a little note about possibly turning the request type check into a whitelist instead of a blacklist, but it's minor.

src/server.c
MathieuBordere: Yes, I think you have a point, the only request type we would like to continue is a `REQUEST_TRANSFER`. I'll do some tests and update the PR.

src/server.c
```diff
@@ -572,27 +572,33 @@ static void monitor_cb(uv_prepare_t *monitor)
...
+				tracef("delayed close of ongoing non-sql request");
+				g->leadership_lost = true;
+			} else {
+				tracef("immediate close of ongoing sql request or idle connection g->sql:%p", g->sql);
+				conn__stop(conn);
+			}
```
freeekanayaka, line 599: Another little nit: perhaps you could encapsulate the logic inside this loop into a public ```conn``` method, so details about the ```gateway``` object are not exposed (if you notice, there are no references to the ```gateway``` in ```server.c```).

For example in ```conn.h``` there could be something like:

```C
/**
 * Called in case Raft leadership gets lost.
 *
 * If the connection is serving a SQL request, it will be closed immediately,
 * otherwise it will be closed as soon as the request completes.
 */
void conn__leadership_lost(struct conn *c);
```

which basically contains the logic inside the loop body.

src/server.c
```diff
@@ -572,27 +572,33 @@ static void monitor_cb(uv_prepare_t *monitor)
...
-	/*
 	if (d->raft_state == RAFT_LEADER && state != RAFT_LEADER) {
+		tracef("leadership lost, close leader connections");
 		QUEUE__FOREACH(head, &d->conns)
 		{
 			conn = QUEUE__DATA(head, struct conn, queue);
-			conn__stop(conn);
+			g = &conn->gateway;
+			if (g == NULL || g->leader == NULL) {
+				continue;
```
freeekanayaka, line 590: Why are you checking for ```g->leader``` also? In case ```g->leader``` is ```NULL``` this should be a connection that is handling management requests no? So in that case we treat the connection differently and not close it? Or perhaps I'm confused.

src/gateway.h
```diff
@@ -29,6 +29,7 @@ struct gateway
 	struct registry *registry;   /* Register of existing databases */
 	struct raft *raft;           /* Raft instance */
 	struct leader *leader;       /* Leader connection to the database */
+	bool leadership_lost;        /* Flag if node lost leadership */
```
freeekanayaka, line 32: Looking closer, it seems that the ```gateway``` object doesn't really make use of this field on its own, while it's the ```conn``` object that uses it. I'd suggest to move the field to the ```conn``` object instead.

freeekanayaka:
Added a few more minor notes.

src/server.c
MathieuBordere: Ok, makes sense.

src/server.c
MathieuBordere: The main point of this PR was to close the sqlite leader connection so that sqlite cleans up its state from an ongoing sql request resulting from a leadership loss. I reasoned that connections without a `g->leader` could stay open as they don't have an open sqlite leader connection and are safe to be reused by the sql driver.

src/gateway.h
MathieuBordere: okay

src/server.c
freeekanayaka: I see.

I believe that there are no cases where we mix using a connection both for SQL and management requests. So perhaps for sake of simplicity you could just check that and simplify the code a bit.

freeekanayaka:
After having thought about this a bit more, another thing that comes to my mind is what I said in the comment I had made [above](https://github.com/canonical/dqlite/pull/354#issuecomment-1055323682).

I believe it's fine to land this PR as it makes things work.

However I'd like to understand why we need the somehow indirect approach of closing connections. Ideally a request should always get a response, if the request (e.g. a SQL request) fails because leadership has been lost, the client should get back that error. If there are pending transactions when frames get applied, they should be rolled back. In other world it feels that we're applying a brute force solution because we can't quite handle state more gracefully. This might be something to look at in the future.

MathieuBordere:
> After having thought about this a bit more, another thing that comes to my mind is what I said in the comment I had made [above](https://github.com/canonical/dqlite/pull/354#issuecomment-1055323682).
> 
> I believe it's fine to land this PR as it makes things work.
> 
> However I'd like to understand why we need the somehow indirect approach of closing connections. Ideally a request should always get a response, if the request (e.g. a SQL request) fails because leadership has been lost, the client should get back that error. If there are pending transactions when frames get applied, they should be rolled back. In other world it feels that we're applying a brute force solution because we can't quite handle state more gracefully. This might be something to look at in the future.

I thought about this too, this is indeed a bit of a brute force solution. When leadership is lost, ideally the client should still be able to send "ABORT" or "ROLLBACK" sql statements or execute prepared "ABORT"/"ROLLBACK" statements, today that's not possible because `exec` or `exec sql` requests will fail when a node is no longer leader. 
It would be cleaner but it would be a fair bit more complex, I agree we should look at it in the future.

src/server.c
MathieuBordere: Hmm yes, you're right, the go client doesn't mix them. They are mixed in the C unit tests though, will adapt and just close any connection that has a `gateway->leader`, then we don't need that delayed close logic at all.

MathieuBordere:
Simplified logic a bit, any connection where `gateway->leader != NULL` can be closed.

freeekanayaka (approved):
Looks good to me, thanks for the cleanups!

freeekanayaka:
> > After having thought about this a bit more, another thing that comes to my mind is what I said in the comment I had made [above](https://github.com/canonical/dqlite/pull/354#issuecomment-1055323682).
> > I believe it's fine to land this PR as it makes things work.
> > However I'd like to understand why we need the somehow indirect approach of closing connections. Ideally a request should always get a response, if the request (e.g. a SQL request) fails because leadership has been lost, the client should get back that error. If there are pending transactions when frames get applied, they should be rolled back. In other world it feels that we're applying a brute force solution because we can't quite handle state more gracefully. This might be something to look at in the future.
> 
> I thought about this too, this is indeed a bit of a brute force solution. When leadership is lost, ideally the client should still be able to send "ABORT" or "ROLLBACK" sql statements or execute prepared "ABORT"/"ROLLBACK" statements (this one will be a bit ugly), today that's not possible because `exec` or `exec sql` requests will fail when a node is no longer leader. It would be cleaner but it would be a fair bit more complex, I agree we should look at it in the future.

FWIW they way I intended it to work originally was that the client simply receives a "leadership-lost" error code and would not need to issue a rollback. The client would typically just close the connection (or use it for requests that don't require a leader) and look for a new leader. That's what is already happening in the Go driver code [here](https://github.com/canonical/go-dqlite/blob/master/driver/driver.go#L752). The problem is that apparently the leadership-lost error doesn't quite get propagated, plus you report there is some leftover state on the server side which triggers an assertion once that node receives frames from the new leader. Those are the two things that should be looked at I believe.

MathieuBordere:
> > > After having thought about this a bit more, another thing that comes to my mind is what I said in the comment I had made [above](https://github.com/canonical/dqlite/pull/354#issuecomment-1055323682).
> > > I believe it's fine to land this PR as it makes things work.
> > > However I'd like to understand why we need the somehow indirect approach of closing connections. Ideally a request should always get a response, if the request (e.g. a SQL request) fails because leadership has been lost, the client should get back that error. If there are pending transactions when frames get applied, they should be rolled back. In other world it feels that we're applying a brute force solution because we can't quite handle state more gracefully. This might be something to look at in the future.
> > 
> > 
> > I thought about this too, this is indeed a bit of a brute force solution. When leadership is lost, ideally the client should still be able to send "ABORT" or "ROLLBACK" sql statements or execute prepared "ABORT"/"ROLLBACK" statements (this one will be a bit ugly), today that's not possible because `exec` or `exec sql` requests will fail when a node is no longer leader. It would be cleaner but it would be a fair bit more complex, I agree we should look at it in the future.
> 
> FWIW they way I intended it to work originally was that the client simply receives a "leadership-lost" error code and would not need to issue a rollback. The client would typically just close the connection (or use it for requests that don't require a leader) and look for a new leader. That's what is already happening in the Go driver code [here](https://github.com/canonical/go-dqlite/blob/master/driver/driver.go#L752). The problem is that apparently the leadership-lost error doesn't quite get propagated, plus you report there is some leftover state on the server side which triggers an assertion once that node receives frames from the new leader. Those are the two things that should be looked at I believe.

I had a variant of this PR that closes the leader connection and basically always reports "LEADERSHIP_LOST" on a request see https://github.com/MathieuBordere/dqlite/commit/86c7b8b276488811ef91e0272de9817aa7fadd27.
It's pretty quick to change it to that implementation if you prefer.

freeekanayaka:
> > > > After having thought about this a bit more, another thing that comes to my mind is what I said in the comment I had made [above](https://github.com/canonical/dqlite/pull/354#issuecomment-1055323682).
> > > > I believe it's fine to land this PR as it makes things work.
> > > > However I'd like to understand why we need the somehow indirect approach of closing connections. Ideally a request should always get a response, if the request (e.g. a SQL request) fails because leadership has been lost, the client should get back that error. If there are pending transactions when frames get applied, they should be rolled back. In other world it feels that we're applying a brute force solution because we can't quite handle state more gracefully. This might be something to look at in the future.
> > > 
> > > 
> > > I thought about this too, this is indeed a bit of a brute force solution. When leadership is lost, ideally the client should still be able to send "ABORT" or "ROLLBACK" sql statements or execute prepared "ABORT"/"ROLLBACK" statements (this one will be a bit ugly), today that's not possible because `exec` or `exec sql` requests will fail when a node is no longer leader. It would be cleaner but it would be a fair bit more complex, I agree we should look at it in the future.
> > 
> > 
> > FWIW they way I intended it to work originally was that the client simply receives a "leadership-lost" error code and would not need to issue a rollback. The client would typically just close the connection (or use it for requests that don't require a leader) and look for a new leader. That's what is already happening in the Go driver code [here](https://github.com/canonical/go-dqlite/blob/master/driver/driver.go#L752). The problem is that apparently the leadership-lost error doesn't quite get propagated, plus you report there is some leftover state on the server side which triggers an assertion once that node receives frames from the new leader. Those are the two things that should be looked at I believe.
> 
> I have a variant of this PR that only closes the leader connection, leaves open the network connection and basically always reports "LEADERSHIP_LOST" on a request see [MathieuBordere@86c7b8b](https://github.com/MathieuBordere/dqlite/commit/86c7b8b276488811ef91e0272de9817aa7fadd27). It's pretty quick to change it to that implementation if you prefer.

Since it this PR seems reasonably easy, I think it's fine to merge it. Perhaps we could look at your alternative PR separately, since I guess it will require some more work and testing on your side. I'd surely prefer to see a more graceful way of error reporting and rollback, but it can be a follow-up step.

PR #356, opened 2022-03-24
Shallow snapshot

MathieuBordere:
@freeekanayaka 

This PR is WIP, it is just to have a discussion on the methodology, don't review the code yet. As discussed in the raft PR on the async snapshots, the trick to have faster snapshots in dqlite is to: 
- copy the WAL
- disallow checkpoints while taking a snapshot
- pass pointers to the database pages to raft to eliminate the copy of the whole database
- allow checkpoints again after raft is done accessing the page pointers

I currently set a flag `read_lock` on the db object to indicate that a snapshot or a checkpoint is busy and allow neither to start when the other is running.

The problem I encountered is that, in the current implementation, the dqlite leader decides when to checkpoint, it then issues a checkpoint command to the whole cluster and when a node applies the checkpoint command, the sqlite database will be checkpointed.

Because nodes independently decide when to snapshot, it can happen that the application of that checkpoint command can fail due to a follower that is taking a snapshot (holding the `read_lock` while compressing and writing to disk) at that time, and the checkpoint command will not be executed again, because we only apply raft logs once.
My solution to this is to also let the nodes decide independently when to checkpoint, this happens at the end of `apply_frames` instead of as a result of a checkpoint command.

Now, here comes the question, do you see any obvious issues with that approach?

### DIFF
--- a/src/leader.c
+++ b/src/leader.c
@@ -122,7 +122,6 @@ int leader__init(struct leader *l, struct db *db, struct raft *raft)
 	}
 
 	l->exec = NULL;
-	l->apply.data = l;
 	l->inflight = NULL;
 	QUEUE__PUSH(&db->leaders, &l->queue);
 	return 0;
@@ -155,8 +154,8 @@ static void leaderCheckpointApplyCb(struct raft_apply *req,
 				    int status,
 				    void *result)
 {
-	(void)req;
 	(void)result;
+	raft_free(req);
 	if (status != 0) {
                 tracef("checkpoint apply failed %d", status);
 	}
@@ -202,11 +201,22 @@ static void leaderMaybeCheckpointLegacy(struct leader *l)
 		return;
 	}
 
-	rv = raft_apply(l->raft, &l->apply, &buf, 1, leaderCheckpointApplyCb);
+	struct raft_apply *apply = raft_malloc(sizeof(*apply));
+	if (apply == NULL) {
+		tracef("raft_malloc - no mem");
+		goto err_after_buf_alloc;
+	}
+	rv = raft_apply(l->raft, apply, &buf, 1, leaderCheckpointApplyCb);
 	if (rv != 0) {
 		tracef("raft_apply failed %d", rv);
-		raft_free(buf.base);
+		raft_free(apply);
+		goto err_after_buf_alloc;
 	}
+
+	return;
+
+err_after_buf_alloc:
+	raft_free(buf.base);
 }
 
 static void leaderApplyFramesCb(struct raft_apply *req,
--- a/src/leader.h
+++ b/src/leader.h
@@ -43,7 +43,6 @@ struct leader
 	sqlite3 *conn;           /* Underlying SQLite connection. */
 	struct raft *raft;       /* Raft instance. */
 	struct exec *exec;       /* Exec request in progress, if any. */
-	struct raft_apply apply; /* To apply checkpoint commands */
 	queue queue;             /* Prev/next leader, used by struct db. */
 	struct apply *inflight;  /* TODO: make leader__close async */
 };


freeekanayaka:
> @freeekanayaka
> 
> This PR is WIP, it is just to have a discussion on the methodology, don't review the code yet. As discussed in the raft PR on the async snapshots, the trick to have faster snapshots in dqlite is to:
> 
>     * copy the WAL
> 
>     * disallow checkpoints while taking a snapshot
> 
>     * pass pointers to the database pages to raft to eliminate the copy of the whole database
> 
>     * allow checkpoints again after raft is done accessing the page pointers
> 
> 
> I currently set a flag `read_lock` on the db object to indicate that a snapshot or a checkpoint is busy and allow neither to start when the other is running.
> 
> The problem I encountered is that, in the current implementation, the dqlite leader decides when to checkpoint, it then issues a checkpoint command to the whole cluster and when a node applies the checkpoint command, the sqlite database will be checkpointed.
> 
> Because nodes independently decide when to snapshot, it can happen that the application of that checkpoint command can fail due to a follower that is taking a snapshot (holding the `read_lock` while compressing and writing to disk) at that time, and the checkpoint command will not be executed again, because we only apply raft logs once. My solution to this is to also let the nodes decide independently when to checkpoint, this happens at the end of `apply_frames` instead of as a result of a checkpoint command.
> 
> Now, here comes the question, do you see any obvious issues with that approach?

I looked very briefly at the change, and at first sight, yes I do see a potential problem. Although I'd need to refresh my memory on some aspects of the code to confirm. Currently the leader decides whether to issue a checkpoint command in `leaderMaybeCheckpoint` (in `leader.c`), and basically the main check is whether there are active readers: if we're above the threshold and there are no readers, then a checkpoint FSM is issued. At that point we're guaranteed that the checkpoint operation will succeed on all FSMs (leader and followers), since the leader will acquire a lock (in `leaderMaybeCheckpoint`) and followers don't currently support readers. This makes the FSM totally deterministic. Instead, if we perform the "are the active readers?" check in the apply-frames command, then if we are on the leader it might turn out that there are active readers and the checkpoint can't be performed. That makes the FSM non deterministic, and you'd end up with a different state on the leader than on followers. At that point I believe that further apply-frames commands will be inconsistent because the WAL will be different. The current design (i.e. the checkpoint command) was done with that goal in mind: fully deterministic FSMs, as the Raft paper requires.

This is a bit of a tricky problem and it seems equivalent to the problem of allowing readers on followers. One obvious idea would be to "throttle" checkpoint commands and subsequent frames commands, but it feels complicated.

MathieuBordere:
> > @freeekanayaka
> > This PR is WIP, it is just to have a discussion on the methodology, don't review the code yet. As discussed in the raft PR on the async snapshots, the trick to have faster snapshots in dqlite is to:
> > ```
> > * copy the WAL
> > 
> > * disallow checkpoints while taking a snapshot
> > 
> > * pass pointers to the database pages to raft to eliminate the copy of the whole database
> > 
> > * allow checkpoints again after raft is done accessing the page pointers
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I currently set a flag `read_lock` on the db object to indicate that a snapshot or a checkpoint is busy and allow neither to start when the other is running.
> > The problem I encountered is that, in the current implementation, the dqlite leader decides when to checkpoint, it then issues a checkpoint command to the whole cluster and when a node applies the checkpoint command, the sqlite database will be checkpointed.
> > Because nodes independently decide when to snapshot, it can happen that the application of that checkpoint command can fail due to a follower that is taking a snapshot (holding the `read_lock` while compressing and writing to disk) at that time, and the checkpoint command will not be executed again, because we only apply raft logs once. My solution to this is to also let the nodes decide independently when to checkpoint, this happens at the end of `apply_frames` instead of as a result of a checkpoint command.
> > Now, here comes the question, do you see any obvious issues with that approach?
> 
> I looked very briefly at the change, and at first sight, yes I do see a potential problem. Although I'd need to refresh my memory on some aspects of the code to confirm. Currently the leader decides whether to issue a checkpoint command in `leaderMaybeCheckpoint` (in `leader.c`), and basically the main check is whether there are active readers: if we're above the threshold and there are no readers, then a checkpoint FSM command is issued. At that point we're guaranteed that the checkpoint operation will succeed on all FSMs (leader and followers), since the leader will acquire a lock (in `leaderMaybeCheckpoint`) and followers don't currently support readers. This makes the FSM totally deterministic. Instead, if we perform the "are there active readers?" check in the apply-frames command, then if we are on the leader it might turn out that there are active readers and the checkpoint can't be performed. That makes the FSM non deterministic, and you'd end up with a different state on the leader than on followers. At that point I believe that further apply-frames commands will be inconsistent because the WAL will be different. The current design (i.e. the checkpoint command) was done with that goal in mind: fully deterministic FSMs, as the Raft paper requires.
> 
> This is a bit of a tricky problem and it seems equivalent to the problem of allowing readers on followers. One obvious idea would be to "throttle" checkpoint commands and subsequent frames commands, but it feels complicated.

It will be inconsistent in how the information spread between the database and the WAL but the content should be identical no?

freeekanayaka:
Unless there's something obvious that I'm missing, I have the sensation this is going to be a bit of a tough problem :/ Or at least I can't think of any straightforward solution right now. What do you think?

freeekanayaka:
> It will be inconsistent in how the information is spread between the database and the WAL but the content should be identical no?

Say that at apply-frames command N the followers perform the checkpoint but leader does not. Won't the apply-frames command N+1 be screwed? (i.e. the followers will not be able to apply it correctly because the WAL is different). Or maybe it was a problem only in earlier implementations. I'm checking the code now to refresh my memory.

freeekanayaka:
Okay, probably I'm confusing the situation with an earlier implementation of dqlite where we were actually sending WAL frames and not merely database pages and so there was a need for the checkpoint command. Now we're only shipping page numbers, and there seems to be no dependency on the WAL state. So it might be fine, I'd just be extra careful and check if there is any subtle ramification with the fact that although the information in the database is the same, the physical bytes can be different and so the "FSM is deterministic" assumption gets relaxed. I don't know if there are consequences of that.

MathieuBordere:
I think this is now ready for review, please take your time to go through this and let me clarify anything that comes to your minds, I'm out for the evening, will comment back in the morning.

Changes:
- Nodes no longer checkpoint as a result of a checkpoint command, but independently decide to checkpoint the database based on the size of the WAL as a result of running `apply_frames`. Have added tests in `test_vfs` to test some scenarios where nodes checkpoint at different times.
- dqlite now uses the `snapshot_finalize` method introduced in raft `0.13.0`. In `fsm__snapshot` dqlite will set a flag, `read_lock` on the database indicating that a snapshot is active, this flag disallows checkpoints from running at the same time as a snapshot. dqlite will then deep-copy the WAL and copy pointers to the database pages. These pointers can be safely accessed by raft to write the snapshot to disk as long as the WAL is not checkpointed into the database, this is prevented by the aforementioned `read_lock`. The snapshot format on disk has not been changed. Finally, after the snapshot has reached disk or has erred out, raft will call `fsm__snapshot_finalize` that will release the `read_lock`, allowing checkpoints to be performed again and will free any allocated buffers.

Remarks:
- Legacy (pre this PR) nodes in a cluster ran by a Modern (this PR) node will still receive checkpoint commands, the checkpoint commands are no-ops on Modern nodes. A leader will issue checkpoint commands once its WAL size has 0-size, so once it has performed a checkpoint itself. This is because the callback that contains this logic is called after the WAL has been potentially checkpointed.
- Modern nodes in a cluster run by a Legacy leader will perform checkpoints autonomously and will perform the no-op checkpoint commands.
- snapshots that would take ages to reach disk will block the checkpoint from running, however once the snapshot eventually reaches disk, the checkpoint will have a chance to run before a new snapshot is triggered by raft. This is because raft applies commands before taking a snapshot & dqlite checkpoints during the execution of a command. So there's a minor risk that the WAL size can grow largeish during a snapshot write, but I believe this is minor.

src/fsm.c
```diff
@@ -61,6 +61,123 @@ static int add_pending_pages(struct fsm *f,
...
+		goto cleanup;
+	}
+	tracef("sqlite3_wal_checkpoint_v2 success");
+
+	/* Since no reader transaction is in progress, we must be able to
+	 * checkpoint the entire WAL */
+	assert(wal_size == 0);
+	assert(ckpt == 0);
+
+cleanup:
```
freeekanayaka, line 173: FWIW I believe that both in dqlite and raft the current naming convention for goto labels that rollback state after an error is:

```C
err_after_<name of the operation that has to be rolled back>
```

e.g.

```
goto err_after_database_read_lock
goto err_after_database_open_follower
err_after_open_follower:
   // ...
err_after_database_read_lock:
   // ...
```

The rational being that it should be a tad more clear what labels reverts what operation, plus names like "cleanup" and "unlock" might become ambiguous if more reverts are added.

YMMV

src/fsm.c
```diff
@@ -61,6 +61,123 @@ static int add_pending_pages(struct fsm *f,
...
+
+		/* Not locked. Let's release the lock we just
+		 * acquired. */
+		flags = SQLITE_SHM_UNLOCK | SQLITE_SHM_EXCLUSIVE;
+		main_f->pMethods->xShmLock(main_f, i, 1, flags);
+	}
+
+	rv = sqlite3_wal_checkpoint_v2(
+	     db->follower, "main", SQLITE_CHECKPOINT_TRUNCATE, &wal_size, &ckpt);
+	if (rv != 0) {
```
freeekanayaka, line 162: I would have expected this to be an `assert(rv == 0)`, but I noticed the old code removed just below does the same. It might be interesting to dig in git history to see if it has always been like this or was added for a specific reason. If it was always like this, perhaps it'd be worth checking the SQLite code if there are actually valid failure modes that we have to guard from.

This is optional, since basically it doesn't change current behavior.

src/fsm.c
```diff
@@ -395,37 +479,115 @@ static int decodeDatabase(struct fsm *f, struct cursor *cursor)
...
+	QUEUE__FOREACH(head, &f->registry->dbs)
+	{
+		if (i == n_bufs) {
+		    break;
+		}
+		db = QUEUE__DATA(head, struct db, queue);
+		/* i is the index of the database header */
+		sqlite3_free(bufs[i].base);
+		/* i is now the index of the next database header (if any) */
+		i += 1 + dbNumPages(db) + 1;
```
freeekanayaka, line 547: Would be nice to add a comment about what those `1`s are (I guess the two bueffers for the database header and WAL). E.g.

```C
i += 1 /* db header */ + dbNumPages(db) + 1 /* WAL */;
```

src/fsm.c
```diff
@@ -435,27 +597,81 @@ static int fsm__snapshot(struct raft_fsm *fsm,
...
+	}
+
+	/* Free allocated buffers */
+	freeSnapshotBufs(f, *bufs, *n_bufs);
+	sqlite3_free(*bufs);
+	*bufs = NULL;
+	*n_bufs = 0;
+
+	/* Unlock all databases that were locked for the snapshot, this is safe
+	 * because DB's are only ever added at the back of the queue. */
```
freeekanayaka, line 660: Are there are cases where DBs are removed? I guess currently not.

src/leader.c
```diff
@@ -145,41 +145,36 @@ void leader__close(struct leader *l)
...
 
-/* Attempt to perform a checkpoint if possible. */
-static bool leaderMaybeCheckpoint(struct leader *l)
+/* Attempt to perform a checkpoint on nodes running a version of dqlite that
+ * doens't perform autonomous checkpoints. For recent nodes, the checkpoint
+ * command will just be a no-op.
+ * This function will run after the WAL might have been checkpointed during a call
+ * to `apply_frames`.
+ * */
+static void leaderMaybeCheckpointLegacy(struct leader *l)
```
freeekanayaka, line 171: Is this really necessary? It sounds like it will generate a checkpoint command even on clusters having all new nodes, where that's not needed. I believe that the situation where the leader is newer and there are older followers should be generally short lived (since those older followers should get updated soon), and maybe it makes more sense to optimize for the common case where all nodes are new.

freeekanayaka (approved):
Looks good to me, minus nitpicks. Good work with VFS-level tests.

src/fsm.c
MathieuBordere: Thanks, I'll add a `TODO` for now.

src/fsm.c
MathieuBordere: Indeed, currently not.

src/leader.c
MathieuBordere: I know it's not ideal but on a write heave database the WAL can quickly grow out of control I think. I could leave this in for a couple of months and remove it after say 4-6 months.

MathieuBordere:
Found a bug, will reopen soon.

stgraber:
@freeekanayaka got a few minutes to do another quick review on this before we hit merge?

PR #365, opened 2022-04-26
Revert "Close leader connection"

MathieuBordere:
Reverts canonical/dqlite#354

The LXD proxy is not liking the fact that I close the network connections, already have a fix for this, will revert for now.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -365,15 +365,6 @@ int clientSendRemove(struct client *c, unsigned id)
 	return 0;
 }
 
-int clientSendTransfer(struct client *c, unsigned id)
-{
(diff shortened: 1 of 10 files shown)


freeekanayaka:
Oh, that's unfortunate. What fix are you planning? To fix the LXD proxy or change approach and implement better error propagation?

MathieuBordere:
> Oh, that's unfortunate. What fix are you planning? To fix the LXD proxy or change approach and implement better error propagation?

It's not the proxy, it's the driver that doesn't recognize EOF, also see https://github.com/canonical/go-dqlite/issues/182

PR #366, opened 2022-04-26
Close leader sql connection

MathieuBordere:
As discussed in https://github.com/canonical/go-dqlite/pull/186, similar to https://github.com/canonical/dqlite/pull/354, but instead of closing the network connection, it only closes the `gateway->leader` and the SQLite connection it contains. Only the last commit contains unreviewed changes.

Ongoing requests will be canceled with `LEADERSHIP_LOST`, future requests will err out with `SQLITE_NOTFOUND` because the database is not opened. This should be viewed in conjunction with https://github.com/canonical/go-dqlite/pull/187 that will mark the connection as bad resulting in the driver to reopen a new connection.

`LOOKUP_DB` at the start of the gateway functions that deal with the SQLite leader connection will err out with `SQLITE_NOTFOUND` after the node has lost leadership, so a leader connection that has lost leadership once will remain broken.

fixes https://github.com/canonical/dqlite/issues/355

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -572,27 +572,21 @@ static void monitor_cb(uv_prepare_t *monitor)
 {
 	struct dqlite_node *d = monitor->data;
 	int state = raft_state(&d->raft);
-	/*
 	queue *head;
 	struct conn *conn;
-	*/
 
 	if (state == RAFT_UNAVAILABLE) {
 		return;
 	}
 
-	/* TODO: we should shutdown clients that are performing SQL requests,
-	 * but not the ones which are doing management-requests, such as
-	 * transfer leadership.  */
-	/*
 	if (d->raft_state == RAFT_LEADER && state != RAFT_LEADER) {
+		tracef("node %llu@%s: leadership lost", d->raft.id, d->raft.address);
 		QUEUE__FOREACH(head, &d->conns)
 		{
 			conn = QUEUE__DATA(head, struct conn, queue);
-			conn__stop(conn);
+			gateway__leader_close(&conn->gateway, RAFT_LEADERSHIPLOST);
 		}
 	}
-	*/
 
 	d->raft_state = state;
 }
--- a/src/client.c
+++ b/src/client.c
@@ -365,6 +365,15 @@ int clientSendRemove(struct client *c, unsigned id)
 	return 0;
 }
 
+int clientSendTransfer(struct client *c, unsigned id)
+{
+	tracef("client send transfer fd %d id %u", c->fd, id);
+	struct request_transfer request;
+	request.id = id;
+	REQUEST(transfer, TRANSFER);
+	return 0;
+}
+
 int clientRecvEmpty(struct client *c)
 {
 	tracef("client recv empty fd %d", c->fd);
--- a/src/client.h
+++ b/src/client.h
@@ -79,6 +79,9 @@ int clientSendAssign(struct client *c, unsigned id, int role);
 /* Send a request to remove a server from the cluster. */
 int clientSendRemove(struct client *c, unsigned id);
 
+/* Send a request to transfer leadership to node with id `id`. */
+int clientSendTransfer(struct client *c, unsigned id);
+
 /* Receive an empty response. */
 int clientRecvEmpty(struct client *c);
 
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -31,11 +31,10 @@ void gateway__init(struct gateway *g,
 	g->protocol = DQLITE_PROTOCOL_VERSION;
 }
 
-void gateway__close(struct gateway *g)
+void gateway__leader_close(struct gateway *g, int reason)
 {
-	tracef("gateway close");
-	if (g->leader == NULL) {
-		stmt__registry_close(&g->stmts);
+	if (g == NULL || g->leader == NULL) {
+		tracef("gateway:%p or gateway->leader are NULL", g);
 		return;
 	}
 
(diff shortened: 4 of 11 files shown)


src/server.c
```diff
@@ -572,27 +572,21 @@ static void monitor_cb(uv_prepare_t *monitor)
...
-	 * but not the ones which are doing management-requests, such as
-	 * transfer leadership.  */
-	/*
 	if (d->raft_state == RAFT_LEADER && state != RAFT_LEADER) {
+		tracef("node %llu@%s: leadership lost", d->raft.id, d->raft.address);
 		QUEUE__FOREACH(head, &d->conns)
 		{
 			conn = QUEUE__DATA(head, struct conn, queue);
-			conn__stop(conn);
+			gateway__leader_close(&conn->gateway, RAFT_LEADERSHIPLOST);
```
freeekanayaka, line 587: The `RAFT_LEADERSHIPLOST` error should actually be generated only by the raft library.

To put it shortly, by the time `monitor_cb` gets called, the raft library should have already triggered a cascade of callbacks that terminates any pending dqlite client request with `LEADERSHIP_LOST`. So a good part of the cleanups happening in `gateway__close` should not be needed, and it would be a bug if they are.

Basically the detailed sequence of events should be:

1. The raft library decides that the node needs to step down from leader state. As part of converting from leader to follower it will terminate all pending apply requests with `RAFT_LEADERSHIPLOST` (see `convertClearLeader()` in `convert.c`).
2. The `leaderApplyFramesCb()` callback in dqlite's `leader.c` gets called for inflight commit requests, which will set `g->leader->inflight` to `NULL` and call `leader_exec_cb()` in `gateway.c` via `maybeExecDone()` .
3. The `leader_exec_cb()` sets `g->req` and `g->stmt` to NULL
4. At the point there should be no state left regarding inflight write client requests, however I guess there might be an pending query statement if it's being batched (see below).
5. The `monitor_cb` gets called after all this has already happened and should not deal with terminating write requests.

So one thing that most probably needs to be fixed is that `gateway__leader_close()` should assert that there are no inflight write requests (and any associated state).

However, one issue of the overall approach that maybe we haven't considered (or at least I didn't) is what happens when the leadership is lost in the middle of batched query response. In that case I believe the server will never send to the client the next batch and the client will be stuck waiting for it. There should be a timeout for that, but it feels a bit sub-optimal.

Probably we need to reason a bit about this case.

MathieuBordere: Good to know that those barrier cleanups should not be needed, they are triggered by tests though so should be easy enough to fix in raft (I hope) when they are removed.

Thanks for laying out what the planned sequence of events is/was, makes things clear.

I will let this one rest then for a while though and see when I can continue work on it, because I need to work on other things related to the cycle.

freeekanayaka: Sounds good, there are a number of details involved, more than we envisioned. I was also thinking about the implications for a future proper handling of `LEADERSHIP_LOST` via client sessions and request IDs, and I'm not totally sure about how the whole mechanics work with respect to this issue.

Just to be clear, those cleanups **are** needed when `gateway__close()` gets called "normally", they are not needed only when called in this case. However, I'll add that those cleanups are still basically hacks and the proper fix would be to make `gateway__close()` asynchronous, so all callbacks have a chance to get triggered instead of having them triggered "by hand" (hope it's clear what I mean). There's a comment about that in [`leaderApplyFramesCb`](https://github.com/canonical/dqlite/blob/master/src/leader.c#L254):

MathieuBordere: > The `RAFT_LEADERSHIPLOST` error should actually be generated only by the raft library.
> 
> To put it shortly, by the time `monitor_cb` gets called, the raft library should have already triggered a cascade of callbacks that terminates any pending dqlite client write request with `LEADERSHIP_LOST`. So a good part of the cleanups happening in `gateway__close` should not be needed, and it would be a bug if they are.
> 
> Basically the detailed sequence of events should be:
> 
>     1. The raft library decides that the node needs to step down from leader state. As part of converting from leader to follower it will terminate all pending apply requests with `RAFT_LEADERSHIPLOST` (see `convertClearLeader()` in `convert.c`).
> 
>     2. The `leaderApplyFramesCb()` callback in dqlite's `leader.c` gets called for inflight commit requests, which will set `g->leader->inflight` to `NULL` and call `leader_exec_cb()` in `gateway.c` via `maybeExecDone()` .
> 
>     3. The `leader_exec_cb()` sets `g->req` and `g->stmt` to NULL
> 
>     4. At the point there should be no state left regarding inflight write client requests, however I guess there might be an pending query statement if it's being batched (see below).
> 
>     5. The `monitor_cb` gets called after all this has already happened and should not deal with terminating write requests.
> 
> 
> So one thing that most probably needs to be fixed is that `gateway__leader_close()` should assert that there are no inflight write requests (and any associated state).
> 
> However, one issue of the overall approach that maybe we haven't considered (or at least I didn't) is what happens when the leadership is lost in the middle of batched query response. In that case I believe the server will never send to the client the next batch and the client will be stuck waiting for it. There should be a timeout for that, but it feels a bit sub-optimal.
> 
> Probably we need to reason a bit about this case.

I would be inclined to merge it anyhow as it fixes https://github.com/canonical/dqlite/issues/355 that pops up quite regularly in the LXD testsuite. Unless you think it makes the current implementation worse.

freeekanayaka: I don't think it makes things worse indeed, so I'm ok with that. We can revisit this later on.

Issue #379, opened 2022-08-23
dqlite error codes alias raft error codes


cole-miller:
There is some overlap between error codes used by dqlite and by raft:

https://github.com/canonical/raft/blob/17ce02fa378f650fc6a02d21355c37e684da1167/include/raft.h#L14

https://github.com/canonical/dqlite/blob/20e7ab1d8ed69505d8739e9b4de416e467c185aa/include/dqlite.h#L21

This is a problem because some functions in the dqlite public API can return either a raft error code or a dqlite error code -- dqlite_node_create is an example. An error return from these functions is ambiguous.

I realized this while doing some followup on https://github.com/canonical/go-dqlite/pull/196. The obvious way to fix it is to change the values of the dqlite (or raft) error codes -- unfortunately that's a breaking change.

cole-miller:
Of the functions in dqlite.h, these can return either a raft error code or a dqlite code:

- dqlite_node_create
- dqlite_node_start

It's probably best to patch these to always return a dqlite code. Also, it looks like many of the dqlite_vfs_* functions can return either an sqlite error code (`SQLITE_*`) or a dqlite code, so those should be patched as well.

cole-miller:
@stgraber I figured the functions in src/vfs.h should use sqlite return codes, and then the wrappers in include/dqlite.h should convert those to dqlite return codes -- does that sound reasonable?

PR #380, opened 2022-08-23
Fix non-dqlite return codes in public API

cole-miller:
This fixes a couple of cases where a function belonging to the dqlite public API could return with a non-dqlite error code. See #379 -- I've left src/vfs.c alone for now because I'm not sure whether the functions in src/vfs.h should be using dqlite return codes or sqlite codes, but I can add a commit addressing those functions once the convention is established.


stgraber:
Those two look fine to me. I'll let @MathieuBordere and you chat about the others when he's back.

PR #381, opened 2022-08-25
Expose string errors from dqlite_node_create

cole-miller:
This implements roughly @freeekanayaka's suggestion in https://github.com/canonical/raft/issues/228. If we agree this is the way to go, I'll make a PR to go-dqlite to adjust its behavior accordingly.

Note that unlike Free's sample code in that thread, I don't differentiate in the public API between DQLITE_NOMEM errors and other kinds of errors. Instead, dqlite_node_errmsg and dqlite_node_destroy just handle NULL arguments gracefully. I went this route because it makes callers' lives simpler and because DQLITE_NOMEM is an ambiguous signal -- was it the `struct dqlite_node` allocation that failed, or some other allocation along the way?

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -1,5 +1,6 @@
 #include "server.h"
 
+#include <errno.h>
 #include <stdlib.h>
 #include <sys/un.h>
 #include <time.h>
@@ -39,7 +40,8 @@ int dqlite__init(struct dqlite_node *d,
 	registry__init(&d->registry, &d->config);
 	rv = uv_loop_init(&d->loop);
 	if (rv != 0) {
-		/* TODO: better error reporting */
+		snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "uv_loop_init(): %s",
+			 uv_strerror(rv));
 		rv = DQLITE_ERROR;
 		goto err_after_vfs_init;
 	}
@@ -49,7 +51,8 @@ int dqlite__init(struct dqlite_node *d,
 	}
 	rv = raft_uv_init(&d->raft_io, &d->loop, dir, &d->raft_transport);
 	if (rv != 0) {
-		/* TODO: better error reporting */
+		snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "raft_uv_init(): %s",
+			 d->raft_io.errmsg);
 		rv = DQLITE_ERROR;
 		goto err_after_raft_transport_init;
 	}
@@ -62,6 +65,8 @@ int dqlite__init(struct dqlite_node *d,
 	rv = raft_init(&d->raft, &d->raft_io, &d->raft_fsm, d->config.id,
 		       d->config.address);
 	if (rv != 0) {
+		snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "raft_init(): %s",
+			 raft_errmsg(&d->raft));
 		return DQLITE_ERROR;
 	}
 	/* TODO: expose these values through some API */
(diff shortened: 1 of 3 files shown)


stgraber:
I'll let @freeekanayaka and @MathieuBordere weight in on this one.

src/server.c
```diff
@@ -38,7 +39,7 @@ int dqlite__init(struct dqlite_node *d,
 	registry__init(&d->registry, &d->config);
 	rv = uv_loop_init(&d->loop);
 	if (rv != 0) {
-		/* TODO: better error reporting */
+		snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "uv_loop_init(): failed");
```
freeekanayaka, line 42: It'd be nice to include the error message from libuv, e.g.:

```C
snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "uv_loop_init(): %s", uv_strerror(rv));
```

See http://docs.libuv.org/en/v1.x/errors.html#c.uv_strerror

src/server.c
```diff
@@ -48,7 +49,7 @@ int dqlite__init(struct dqlite_node *d,
 	}
 	rv = raft_uv_init(&d->raft_io, &d->loop, dir, &d->raft_transport);
 	if (rv != 0) {
-		/* TODO: better error reporting */
+		snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "raft_uv_init(): failed");
```
freeekanayaka, line 52: Similarly, we could provide more details using something like:

```C
snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "raft_uv_init(): %s", d->raft_io.errmsg);
```

See the implementation of `raft_uv_init()` for details.

src/server.c
```diff
@@ -77,13 +80,13 @@ int dqlite__init(struct dqlite_node *d,
 #else
 	rv = sem_init(&d->ready, 0, 0);
 	if (rv != 0) {
-		/* TODO: better error reporting */
+		snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "sem_init(): failed");
```
freeekanayaka, line 83: Here as well, something like:

```C
snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "sem_init(): %s", strerror(rv));
```

src/server.c
```diff
@@ -77,13 +80,13 @@ int dqlite__init(struct dqlite_node *d,
...
 		rv = DQLITE_ERROR;
 		goto err_after_raft_fsm_init;
 	}
 	rv = sem_init(&d->stopped, 0, 0);
 	if (rv != 0) {
-		/* TODO: better error reporting */
+		snprintf(d->errmsg, DQLITE_ERRMSG_BUF_SIZE, "sem_init(): failed");
```
freeekanayaka, line 89: Same as above.

freeekanayaka (approved):
Just a few suggestions.

I agree with the idea of not special-casing `DQLITE_NOMEM`, since it is indeed ambiguous, and the approach in the PR actually provides more details.

src/server.c
cole-miller: strerror is not necessarily thread-safe, which seems like a concern here, so maybe strerror_r would be better? I see strerror is already used elsewhere.

freeekanayaka: > strerror is not necessarily thread-safe, which seems like a concern here, so maybe strerror_r would be better? I see strerror is already used elsewhere.

In theory we could hit some races, if for example you run this code concurrently with other parts of your application, but that seems unlikely. That being said, using `strerror_r` would be certainly better.

cole-miller: I'll use strerror for now and look into switching everything to strerror_r in a separate PR.

MathieuBordere (commented):
Looks good!

MathieuBordere:
@cole-miller do you expect go-dqlite to behave weirdly if I merge this?

cole-miller:
We'll want to update go-dqlite to revert my https://github.com/canonical/go-dqlite/pull/196, but it should be okay to merge this first -- it won't change go-dqlite's externally visible behavior.

PR #384, opened 2022-08-25
Unify address parsing and support IPv6

### LINKED ISSUE #312: IPv6 support
Trying to bind to `[::1]:9000` or `[fdaa:0:1e1d:b8b:ce2:0:a:2]:9000` results in:
```
create node: failed to set bind address "[fdaa:0:1e1d:b8b:ce2:0:a:2]:9000": 2
```
It seems only IPv4 addresses are [parsed](https://github.com/canonical/dqlite/blob/66338cd6a38e42f328aaeebea2189a1e96681054/src/server.c#L187):
```
rv = uv_ip4_addr(host, atoi(port), addr)
```
Otherwise everything seems IPv6 ready. I might try myself, but my C is a little rusty. Any plans to support IPv6? I'd appreciate it very much and could help with testing.

cole-miller:
This is a draft for now, I'll push more commits later that replace the old address-parsing code and add some tests. Feedback welcome in the meantime!

Closes #312.


cole-miller:
I kept this inside dqlite (instead of exposing a utility function from libraft) for now because I wanted to focus on the smaller implementation-only change. If it works out, we could of course move it into raft at some later point.

The DQLITE_ADDR_PARSE_UNIX thing also warrants explaining. Currently `@`-formatted Unix domain addresses are accepted by dqlite_node_set_bind_address but not by the default callbacks in src/transport.c, and I didn't want to expand what's accepted without discussion. If we want transport.c to handle Unix domain addresses as well, I can adjust the PR.

freeekanayaka:
> I kept this inside dqlite (instead of exposing a utility function from libraft) for now because I wanted to focus on the smaller implementation-only change. If it works out, we could of course move it into raft at some later point.
> 
> The DQLITE_ADDR_PARSE_UNIX thing also warrants explaining. Currently `@`-formatted Unix domain addresses are accepted by dqlite_node_set_bind_address but not by the default callbacks in src/transport.c, and I didn't want to expand what's accepted without discussion. If we want transport.c to handle Unix domain addresses as well, I can adjust the PR.

Setting Unix domain address as bind address is useful only if you are going to proxy incoming connections: that's what go-dqlite and LXD do in order to support TLS. Basically you accept new network TLS connections in Go and then proxy the data back and forth from dqlite by connecting to the Unix socket. Similarly, for outgoing connections, go-dqlite and LXD use `dqlite_node_set_connect_func` to essentially implement the same proxy dance but for outgoing connections.

This means that there's no need to support Unix domain address in src/transport.c, because no real dqlite node would ever have a Unix domain address as real address.

Hope that's clear enough, happy to provide more details if that's confusing :)

src/lib/addr.h
```diff
@@ -0,0 +1,29 @@
...
+ * @addr_len) to hold the result, DQLITE_ERROR is returned.
+ *
+ * @service should be a string representing a port number, e.g. "8080".
+ *
+ * @flags customizes the behavior of the function. Currently the only flag is
+ * DQLITE_ADDR_PARSE_UNIX: when this is ORed in @flags, parseAddr will also
+ * parse Unix socket addresses in the form `@NAME`, where NAME may be empty.
+ * This creates a socket address in the (Linux-specific) "abstract namespace".
+ */
+int parseAddr(const char *input,
```
freeekanayaka, line 23: Nit: please name this `AddrParse()`.

The convention should be CamelCase capitalized names for internally exported function, with the module name as prefix. Private functions declared inside modules as `static` should be camelCase, without capital letter.  See `vfs.h`/`vfs.c` in dqlite for examples of this convention, and the whole code base in libraft.

Unfortunately most of dqlite code base does not yet follow this convention, but it'd be nice to gradually migrate to it as we modify code.

YMMV

freeekanayaka (approved):

cole-miller:
@freeekanayaka Thanks for the explanation -- I guess I had in mind a single-host setup (for testing, say) where all IPC happens over Unix sockets; is that something we want to support? Of course you can also just do the proxy thing with localhost in that situation.

cole-miller:
Okay, ready for final review -- I've replaced the old parsing code, updated the API documentation in dqlite.h, and added some unit tests.

stgraber:
Going to let @freeekanayaka or @MathieuBordere weight in on this one too :)

freeekanayaka:
Looks good to me.

PR #387, opened 2022-09-01
lib/byte: Rewrite to avoid triggering -Wconversion

cole-miller:
Should fix the Launchpad build failures we saw. I'm not sure why -Wconversion didn't fire in our GHA CI -- different compiler versions, maybe?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -45,7 +45,7 @@ TESTS = unit-test integration-test
 
 check_LTLIBRARIES = libtest.la
 
-libtest_la_CFLAGS = $(AM_CFLAGS) -DMUNIT_TEST_NAME_LEN=60 -Wno-unused-result -Wno-conversion -Wno-maybe-uninitialized -Wno-strict-prototypes -Wno-old-style-definition
+libtest_la_CFLAGS = $(AM_CFLAGS) -DMUNIT_TEST_NAME_LEN=60 -Wno-unknown-warning-option -Wno-unused-result -Wno-conversion -Wno-maybe-uninitialized -Wno-strict-prototypes -Wno-old-style-definition
 libtest_la_SOURCES = \
   test/lib/endpoint.c \
   test/lib/fault.c \
@@ -78,7 +78,7 @@ unit_test_SOURCES += \
   test/unit/test_tuple.c \
   test/unit/test_vfs.c \
   test/unit/main.c
-unit_test_CFLAGS = $(AM_CFLAGS) -Wno-maybe-uninitialized -Wno-float-equal -Wno-conversion
+unit_test_CFLAGS = $(AM_CFLAGS) -Wno-unknown-warning-option -Wno-maybe-uninitialized -Wno-float-equal -Wno-conversion
 unit_test_LDFLAGS = $(AM_LDFLAGS)
 unit_test_LDADD = libtest.la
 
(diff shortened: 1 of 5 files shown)


cole-miller:
Ah, we get warnings in CI, but it still passes because we're not running with -Werror.

cole-miller:
I think this is a diagnostics false positive. Looking at this [on Compiler Explorer](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:20,endLineNumber:1,positionColumn:20,positionLineNumber:1,selectionStartColumn:20,selectionStartLineNumber:1,startColumn:20,startLineNumber:1),source:'%23include+%3Cstdint.h%3E%0A%0Auint16_t+ByteGetBe16(const+uint8_t+*buf)%0A%7B%0A%09return+((uint16_t)(buf%5B0%5D)+%3C%3C+8)+%7C+(uint16_t)(buf%5B1%5D)%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:54.6916890080429,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:cg104,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-Wconversion',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+10.4+(C,+Editor+%231,+Compiler+%231)',t:'0')),k:45.3083109919571,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+12.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+10.4+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:45.3083109919571,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4), -Wconversion fires for GCC before 10.4 but not after. The GCC release notes [say](https://gcc.gnu.org/gcc-10/changes.html)

> * Enhancements to existing warnings: 
>   * [-Warith-conversion](https://gcc.gnu.org/onlinedocs/gcc-10.1.0/gcc/Warning-Options.html#index-Warith-conversion) re-enables warnings from -Wconversion, -Wfloat-conversion, and -Wsign-conversion that are now off by default for an expression where the result of an arithmetic operation will not fit in the target type due to promotion, but the operands of the expression do fit in the target type.

(The new -Warith-conversion does not fire for ByteGetBe16 and co.) Note that with pre-10.1 GCC -Wconversion will fire even for something like

```
uint16_t f(uint16_t x) {
    x <<= 8;
    return x;
}
```

So the diagnostic is definitely overly sensitive. Some of the CI jobs (GHA and Launchpad) are running older versions of GCC and therefore we get warnings.

cole-miller:
The latest code in this branch doesn't trigger -Wconversion with any version of Clang (afaict), but it does trigger the false positive -Wconversion with GCC before 10.1, which I don't know how to avoid short of conditionally disabling the diagnostic for those versions.

cole-miller:
Forcing CI to use -Werror uncovered some other minor problems, but I think I've fixed them all now. Hopefully nothing unexpected will turn up in the next Launchpad run.

PR #390, opened 2022-09-02
buffer: Don't unnecessarily realloc in loop.

MathieuBordere:
Small optimization in order not to `realloc` the data buffer multiple times in the loop.


cole-miller (approved):
Looks good to me -- nice catch.

PR #395, opened 2022-09-09
Add barriers before calling sqlite3_prepare_v2

### LINKED ISSUE #210: A raft barrier should be run before preparing stements
If the raft log is not yet replayed by the time a query-sql or prepare command is run, then ```sqlite3_prepare``` might fail because the table is not there. We should run a barrier as appropriate.

cole-miller:
WIP, tests will need fixing. See https://github.com/canonical/raft/issues/302.

- [x] Fix compile errors in tests
- [x] Update failing tests
- [ ] New tests:
  - [x] Check calling gateway__close when the barrier from handle_prepare is in flight
  - [x] Other tests introduced in https://github.com/MathieuBordere/dqlite/tree/commit-previous-term
  - [x] inject failure of `leader__barrier`
  - [x] closing early while exec barrier is in flight
  - [x] prepare request to non-leader
  - [ ] ~~handling of malformed requests~~ lock down how different kinds of gateway errors are handled?
- [x] Remove redundant barrier from `querySqlBarrierCb`
- [x] Delete `handle->registered_stmt` on gateway close

Closes #210
Closes https://github.com/canonical/go-dqlite/issues/78

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -618,21 +618,48 @@ done:
 	g->sql = NULL;
 }
 
+static void execSqlBarrierCb(struct barrier *barrier, int status)
+{
+	tracef("exec sql barrier cb status:%d", status);
+	struct gateway *g = barrier->data;
+	struct handle *req = g->req;
+	const char *sql = g->sql;
+
+	assert(req != NULL);
+	g->req = NULL;
+	g->sql = NULL;
+	if (status != 0) {
+		failure(req, status, "barrier error");
+		return;
+	}
+
+	assert(g->req == NULL);
+	assert(g->sql == NULL);
+	assert(g->stmt == NULL);
+	req->gateway->sql = sql;
+	handle_exec_sql_next(req, false);
+}
+
 static int handle_exec_sql(struct handle *req)
 {
-        tracef("handle exec sql");
+	tracef("handle exec sql");
 	struct cursor *cursor = &req->cursor;
 	struct gateway *g = req->gateway;
+	int rc;
 	START(exec_sql, result);
 	CHECK_LEADER(req);
 	LOOKUP_DB(request.db_id);
 	FAIL_IF_CHECKPOINTING;
 	(void)response;
-	assert(g->req == NULL);
-	assert(g->sql == NULL);
-	assert(g->stmt == NULL);
-	req->gateway->sql = request.sql;
-	handle_exec_sql_next(req, false);
+	g->req = req;
+	g->sql = request.sql;
+	rc = leader__barrier(g->leader, &g->barrier, execSqlBarrierCb);
+	if (rc != 0) {
+		tracef("handle exec sql barrier failed %d", rc);
+		g->req = NULL;
+		g->sql = NULL;
+		return rc;
+	}
 	return 0;
 }
 
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -38,7 +38,7 @@ void gateway__leader_close(struct gateway *g, int reason)
 		return;
 	}
 
-	if (g->stmt != NULL) {
+	if (g->req != NULL) {
 		if (g->leader->inflight != NULL) {
 			tracef("finish inflight apply request");
 			struct raft_apply *req = &g->leader->inflight->req;
@@ -46,7 +46,7 @@ void gateway__leader_close(struct gateway *g, int reason)
 			assert(g->req == NULL);
 			assert(g->stmt == NULL);
 		} else if (g->barrier.cb != NULL) {
-			tracef("finish inflight query barrier");
+			tracef("finish inflight barrier");
 			/* This is not a typo, g->barrier.req.cb is a wrapper
 			 * around g->barrier.cb and when called, will set g->barrier.cb to NULL.
 			 * */
@@ -58,8 +58,8 @@ void gateway__leader_close(struct gateway *g, int reason)
 			struct raft_barrier *b = &g->leader->exec->barrier.req;
 			b->cb(b, reason);
 			assert(g->leader->exec == NULL);
-		} else if (g->req != NULL && g->req->type != DQLITE_REQUEST_QUERY
-					  && g->req->type != DQLITE_REQUEST_EXEC) {
+		} else if (g->stmt != NULL && g->req->type != DQLITE_REQUEST_QUERY
+					   && g->req->type != DQLITE_REQUEST_EXEC) {
 			/* Regular exec and query stmt's will be closed when the
 			 * registry is closed below. */
 			tracef("finalize exec_sql or query_sql type:%d", g->req->type);
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -655,6 +655,7 @@ static int handle_exec_sql(struct handle *req)
 	g->sql = request.sql;
 	rc = leader__barrier(g->leader, &g->barrier, execSqlBarrierCb);
 	if (rc != 0) {
+		g->req = NULL;
 		tracef("handle exec sql barrier failed %d", rc);
 		g->req = NULL;
 		g->sql = NULL;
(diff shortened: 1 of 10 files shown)


cole-miller:
Tests are fixed except for a use-after-free reported by ASAN.

cole-miller:
Tests are passing, but there's at least one problem with the current code -- CHECK_LEADER is deferred until callback invocation, where it really should be run as soon as the request is received. Working on fixing that now.

cole-miller:
Now that tests are passing, I think this is ready for review (commit-by-commit should be easiest). Here's an overview of the changes to make it easier to see what's going on in the diffs.

We update the following handlers in `src/gateway.c`:

- `handle_prepare`: This previously didn't run any kind of barrier, so we add one.
- `handle_exec_sql`: This already runs a barrier indirectly (via `leader__exec`), but only after calling `sqlite3_prepare_v2`, so we update it to run another barrier earlier in execution. I experimented with removing the second barrier since it's now redundant, but it would be a bit messy since the other use of `leader__exec` *does* need it to run a barrier first; we'd have to split out a separate function or add an option to `leader__exec`. The performance cost of the redundancy should be minimal, since the first barrier will ensure that the second call to `needsBarrier` quickly returns false.
- `handle_query_sql`: This previously had a `leader__barrier` call after `sqlite3_prepare_v2`; I removed that barrier and added one earlier on.

We also update the logic in `gateway__leader_close` to reflect the new barrier from `handle_prepare`, which needs to be properly cancelled on gateway close.

In all cases, we run `CHECK_LEADER` before installing the barrier, since `leader__barrier` requires the calling node to be the leader. It's worth noting that this means an ill-formed (unparseable) PREPARE request to a non-leader node will fail with `SQLITE_IOERR_NOT_LEADER` instead of dropping the request.

src/gateway.c
```diff
@@ -288,20 +299,41 @@ static int handle_prepare(struct handle *req)
...
+static int handle_prepare(struct handle *req)
+{
+	tracef("handle prepare");
+	struct gateway *g = req->gateway;
+	int rc;
+	CHECK_LEADER(req);
+	g->req = req;
+	rc = leader__barrier(g->leader, &g->barrier, prepareBarrierCb);
+	if (rc != 0) {
+		tracef("handle prepare barrier failed %d", rc);
```
MathieuBordere, line 334: I think you need to set `g->req` to `NULL` here. Is this line covered by a test? Would be great to add one if it isn't.

cole-miller: Good call -- for testing, how could we inject a failure of `leader__barrier`?

src/gateway.c
```diff
@@ -571,31 +612,72 @@ static void handle_exec_sql_next(struct handle *req, struct cursor *cursor)
...
+static int handle_exec_sql(struct handle *req)
+{
+	tracef("handle exec sql");
 	struct gateway *g = req->gateway;
+	int rc;
+	CHECK_LEADER(req);
+	g->req = req;
+	rc = leader__barrier(g->leader, &g->barrier, execSqlBarrierCb);
+	if (rc != 0) {
+		tracef("handle exec sql barrier failed %d", rc);
```
MathieuBordere, line 657: same cleanup needed as `handle_prepare`

test/integration/test_membership.c
```diff
@@ -111,14 +111,7 @@ TEST(membership, join, setUp, tearDown, 0, NULL)
 	PREPARE("INSERT INTO test(n) VALUES(1)", &stmt_id);
 	EXEC(stmt_id, &last_insert_id, &rows_affected);
 
-	/* The table is visible from the new node */
```
MathieuBordere, line 114: Can we try to elect node 2 here, so that this functionality is still tested? Or is it tested elsewhere?

src/gateway.c
```diff
@@ -288,20 +299,41 @@ static int handle_prepare(struct handle *req)
...
 	response.id = (uint32_t)stmt->id;
 	response.params = (uint64_t)sqlite3_bind_parameter_count(stmt->stmt);
 	SUCCESS(stmt, STMT);
+	return 0; // ignored
+}
+
+static void prepareBarrierCb(struct barrier *barrier, int status)
+{
+	tracef("prepare barrier cb status:%d", status);
+	(void)prepareBarrierCbInner(barrier, status);
```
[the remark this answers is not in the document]
freeekanayaka: Oh I see. Besides the fact that it feels a bit of a smell, it also means that actually the checks performed by macros like `START` will be effectively ignored no?

For example if `rv_ = request_##REQ##__decode(cursor, &request);` returns non-0 then `prepareBarrierCbInner` will silently return without providing any response to the client, and effectively hanging the request forever. Unless I'm missing something.

It'd be nice to keep using those macros at the very beginning of the request processing chain, possibly saving on the `gateway` structure any data that you need to propagate (like it's done with `gateway->sql`). Or something along these lines, I didn't really think a lot about this, just pointing that this trick feels/is probably suboptimal.

test/integration/test_membership.c
cole-miller: Yes, I just did the simplest thing here, but modifying it to elect and talk to node 2 should be straightforward.

src/gateway.c
MathieuBordere: There's this `test_heap_fault_enable` e.g. in `test/unit/test_vfs.c`. We could look for a simple mocking framework too.

cole-miller: Afaict the stuff in `src/lib/heap.{c,h}` (used by `test_vfs.c`) only deals with the sqlite3 allocator, but it should be straightforward to set up what we need using `raft_heap_set` (already public API).

src/gateway.c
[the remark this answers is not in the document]
cole-miller: > For example if rv_ = request_##REQ##__decode(cursor, &request); returns non-0 then prepareBarrierCbInner will silently return without providing any response to the client, and effectively hanging the request forever. Unless I'm missing something.

Yes, that's a good point. I thought this was the same as the current behavior, but it's not -- currently, if (say) `START` fails in `handle_prepare` we end up closing the gateway/connection:

https://github.com/canonical/dqlite/blob/3a9bb7abc316507ad77e80a9df22713dafb72731/src/conn.c#L150-L155

Which is maybe not ideal, but better than silently dropping the message.

> It'd be nice to keep using those macros at the very beginning of the request processing chain, possibly saving on the gateway structure any data that you need to propagate (like it's done with gateway->sql).

Yes, I avoided this on the first pass for the sake of keeping the change simple, but it would allow solving this problem in a less hacky way. I'll take a look at implementing it.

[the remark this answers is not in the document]
freeekanayaka: > > For example if rv_ = request_##REQ##__decode(cursor, &request); returns non-0 then prepareBarrierCbInner will silently return without providing any response to the client, and effectively hanging the request forever. Unless I'm missing something.
> 
> Yes, that's a good point. I thought this was the same as the current behavior, but it's not -- currently, if (say) `START` fails in `handle_prepare` we end up closing the gateway/connection:
> 
> https://github.com/canonical/dqlite/blob/3a9bb7abc316507ad77e80a9df22713dafb72731/src/conn.c#L150-L155
> 
> Which is maybe not ideal, but better than silently dropping the message.

In this specific case closing the connection might be the best you can do since if decoding fails it means the client sent some invalid data and it would be difficult to recover the communication from that. It should basically happen only if a client is buggy.

> 
> > It'd be nice to keep using those macros at the very beginning of the request processing chain, possibly saving on the gateway structure any data that you need to propagate (like it's done with gateway->sql).
> 
> Yes, I avoided this on the first pass for the sake of keeping the change simple, but it would allow solving this problem in a less hacky way. I'll take a look at implementing it.

Perhaps this data should be added to the request object instead of the gateway object, but I'm just hand-waving, as I did really look at the details.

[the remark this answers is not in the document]
cole-miller: Okay, the latest series of commits stores the new `uint64_t db_id` and `struct stmt *registered_stmt` fields in the handle, but continues to use existing gateway fields where they're available. I could do a separate refactoring PR to move some request-scoped gateway fields to the handle.

cole-miller:
This is ready for a second round of review -- I've implemented feedback from the first round and added an additional unit test.

src/gateway.c
```diff
@@ -58,8 +58,8 @@ void gateway__leader_close(struct gateway *g, int reason)
 			struct raft_barrier *b = &g->leader->exec->barrier.req;
```
MathieuBordere, line 58: This branch is no longer covered by the tests, can you check why it's the case? Maybe it's no longer needed. I'm being nitty about it because there were issues with the cleanup in the past.

MathieuBordere (commented):
Looks good! Just a couple of questions

src/gateway.h
```diff
@@ -30,9 +30,12 @@ struct gateway
 	struct raft *raft;           /* Raft instance */
 	struct leader *leader;       /* Leader connection to the database */
 	struct handle *req;          /* Asynchronous request being handled */
+	/* FIXME store this in the req */
 	sqlite3_stmt *stmt;          /* Statement being processed */
+	/* FIXME store this in the req */
 	bool stmt_finalize;          /* Whether to finalize the statement */
```
freeekanayaka, line 36: Note that both `stmt` and `stmt_finalize` are cross-request values, basically because we need batch rows yielded by a query (otherwise the might be to many and the client has no way to cancel). What happens in those cases is that the client sends a first request containing the query statement (either an existing prepared statement or plain SQL), then it gets back immediately a response with the first batch of resulting rows, and then the client has to submit additional requests for each additional batch for rows, which will be delivered as individual response. Hope that's clear.

src/gateway.h
```diff
@@ -54,15 +57,20 @@ void gateway__leader_close(struct gateway *g, int reason);
...
 {
-	void *data; /* User data */
-	int type;   /* Request type */
+	void *data;                   /* User data */
+	int type;                     /* Request type */
 	struct gateway *gateway;
 	struct buffer *buffer;
 	struct cursor cursor;
+	uint64_t db_id;               /* For use by prepare callback */
+	struct stmt *registered_stmt; /* For use by prepare callback */
```
freeekanayaka, line 73: I would probably change this to `size_t stmt_id`, since once a statement gets added to the registry it is normally accessed by ID, rather than by direct memory access. This has the advantage of ensuring consistency, in that you know for sure that the object is actually registered. It is also more consistent with the `db_id` field.

src/gateway.h
```diff
@@ -54,15 +57,20 @@ void gateway__leader_close(struct gateway *g, int reason);
...
 struct handle
 {
-	void *data; /* User data */
-	int type;   /* Request type */
+	void *data;                   /* User data */
+	int type;                     /* Request type */
 	struct gateway *gateway;
 	struct buffer *buffer;
 	struct cursor cursor;
+	uint64_t db_id;               /* For use by prepare callback */
```
freeekanayaka, line 72: Nit: I'd use `size_t` as type here, which is what is used internally for IDs of registered objects. The fixed size types are used when serializing those values for sending them over the wire.

src/gateway.c
freeekanayaka: This change seems indeed a bit fishy. I'd like to understand why it's done.

I'm realizing that putting stuff in the `req` object is a bit more subtle than I thought. We probably need a bit more clarity around that, but I didn't think about it too much yet.

test/unit/test_gateway.c
```diff
@@ -182,6 +188,7 @@ static void handleCb(struct handle *req, int status, int type)
 		prepare.sql = SQL;              \
 		ENCODE(&prepare, prepare);      \
 		HANDLE(PREPARE);                \
```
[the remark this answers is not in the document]
cole-miller: At least on my machine, the following unit tests fail, due to the first assertion in `ASSERT_CALLBACK` triggering:

- `gateway/query/barrier`
- `gateway/query/barrierInFlightQuery`
- `gateway/query/barrierInFlightExec`

What happens in `gateway/query/barrier`, for example, is that the re-elected leader comes to power with an uncommitted entry in its log, so it wants to commit a barrier before answering the second `PREPARE` request, and that means the callback isn't invoked immediately.

src/gateway.h
cole-miller: Thanks for pointing that out, I'll remove the FIXMEs.

src/gateway.c
cole-miller: @MathieuBordere So the gateway uses `leader__exec` in two places: `handle_exec` and `handle_exec_sql`. With this change, the `leader__exec` call from `handle_exec_sql` never installs a barrier because we have a new call to `leader__barrier` just before `sqlite3_prepare_v2`. It sounds like we don't have any test coverage of `handle_exec` that results in actually installing a barrier, but I can quickly fix that.

MathieuBordere: Okay, would be nice to have that!

cole-miller: @freeekanayaka Just to clarify "this change", you mean the commit updating `gateway__leader_close`?

freeekanayaka: Exactly.

cole-miller: Added that unit test in the latest commit -- I've confirmed with `make check-code-coverage` that our tests now cover all the branches in `gateway__leader_close`.

@freeekanayaka My reasoning is that with the other changes in this PR, it's now possible to have an in-flight barrier even when `g->stmt == NULL` (for example, if `handle_prepare` was called). The check in `gateway__leader_close` needs to be updated so that the barrier will be properly cancelled in that situation. The `g->req != NULL` check determines whether there's a request in flight, and then if there is we check which fields (regular barrier, exec barrier, prepared statement) need to be cleaned up. I think the only way this change could cause problems is if it's possible to have `g->req == NULL` but `g->stmt != NULL` -- I haven't found a way for that to happen, but it's possible I missed something.

MathieuBordere (approved):

src/gateway.c
freeekanayaka: Ok, I'm not sure either, but since you've looked at it and you added test I believe it's reasonably safe.

This part of the code is unfortunate because the real solution would be to make `gateway__close()` asynchronous, since that's what it is. We've talked about it in the past, but it's a change that requires a bit of rework at various layers, so for now it's still pending.

freeekanayaka (approved):

PR #398, opened 2022-09-13
vfs fnctl fix

MathieuBordere:
Avoids memory corruption uncovered by the added tests.


cole-miller:
For others' reference, the relevant requirement is documented [here](https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlpragma):

> The handler for an [SQLITE_FCNTL_PRAGMA](https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlpragma) file control can optionally make the first element of the char** argument point to a string obtained from [sqlite3_mprintf()](https://sqlite.org/c3ref/mprintf.html) or the equivalent and that string will become the result of the pragma or the error message if the pragma fails.

PR #401, opened 2022-09-15
Initial disk mode support

MathieuBordere:
~WIP - don't review~

- Have tried to run all existing tests with the new disk vfs. Most failures come from the fsm snapshot functionality not being in place for the on-disk case.
- Not yet 100% sure about the abstraction, I feel like a user should be able to choose the VFS per database, but currently I just put the whole of dqlite in disk-mode, i.e. every database will be stored on disk. If the snapshotting behaviour is different for in-memory/on-disk, different VFS's for different databases could lead to complex snapshotting behaviour in raft, where we have to mix different methods to snapshot. Would like to avoid that.
- SYNCs are still turned off, the real transaction is the raft log being stored to disk, the database writes to disk from SQLite don't have to be synced I think.
- Still needs a bunch of cleanup.

### DIFF
--- a/src/lib/fs.c
+++ b/src/lib/fs.c
@@ -1,0 +1,61 @@
+#include <ftw.h>
+#include <stdio.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "fs.h"
+#include "../tracing.h"
+
+int FsEnsureDir(const char *path)
+{
+	int rv;
+	struct stat st = {0};
+
+	rv = stat(path, &st);
+	if (rv == 0) {
+		if (!S_ISDIR(st.st_mode)) {
+			tracef("%s is not a directory", path);
+			return -1;
+		}
+	}
+
+	/* Directory does not exist */
+	if (rv == -1) {
+		return mkdir(path, 0755);
+	}
+
+	return 0;
+}
+
+
+static int fsRemoveDirFilesNftwFn(const char *       path,
+                                  const struct stat *sb,
+                                  int                type,
+                                  struct FTW *       ftwb)
+{
+	int rv;
+
+	(void)sb;
+	(void)type;
+	(void)ftwb;
+
+	rv = 0;
+
+	/* Don't remove directory */
+	if (S_ISREG(sb->st_mode)) {
+		rv = remove(path);
+	}
+
+	return rv;
+}
+
+int FsRemoveDirFiles(const char *path)
+{
+	int rv;
+
+	rv = nftw(path, fsRemoveDirFilesNftwFn,
+	          10, FTW_DEPTH | FTW_MOUNT | FTW_PHYS);
+	return rv;
+
+}
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -1198,10 +1198,7 @@ static int vfsFileControlPragma(struct vfsFile *f, char **fnctl)
 	assert(left != NULL);
 
 	if (strcmp(left, "page_size") == 0 && right) {
-		/* When the user executes 'PRAGMA page_size=N' we save the
-		 * size internally.
-		 *
-		 * The page size must be between 512 and 65536, and be a
+		/* The page size must be between 512 and 65536, and be a
 		 * power of two. The check below was copied from
 		 * sqlite3BtreeSetPageSize in btree.c.
 		 *
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -1,8 +1,13 @@
 #include <errno.h>
 #include <pthread.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
 #include <sys/time.h>
+#include <unistd.h>
 
 #include <raft.h>
 
@@ -94,6 +99,7 @@ struct vfsDatabase
 {
 	char *name;        /* Database name. */
 	void **pages;      /* All database. */
+	unsigned page_size;/* Only used for on-disk db */
 	unsigned n_pages;  /* Number of pages. */
 	struct vfsShm shm; /* Shared memory. */
 	struct vfsWal wal; /* Associated WAL. */
@@ -236,6 +242,7 @@ static void vfsDatabaseInit(struct vfsDatabase *d)
 {
 	d->pages = NULL;
 	d->n_pages = 0;
+	d->page_size = 0;
 	vfsShmInit(&d->shm);
 	vfsWalInit(&d->wal);
 }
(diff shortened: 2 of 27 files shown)


MathieuBordere:
This PR introduces a "disk-mode" for dqlite where the main SQLite database is
stored in a regular SQLite file on disk instead of storing the pages of the 
database in memory.

disk-mode is considered **experimental**. Users can try it out and report any 
issues with the new SQLite VFS or raft fsm.

### Design considerations:
- The SQLite database files live in a folder `database` in the raft dir. This
  folder is emptied on every startup. The user is urged not to interact 
  directly with the database file while the system is running.
  The database file is not the complete picture of the SQLite database. The 
  WAL still lives in memory and is not persisted to disk.
- Snapshots are taken using the async snapshot interface of raft and are taken
  in 3 (simplified) steps.
  1. copy the WAL and disable checkpoints, this guarantees the contents of the
  disk file will not be overwritten during snapshotting.
  2. `mmap` the database files and write the snapshot to disk
  3. enable checkpoints again and clean up resources.

### Known limitations:
- The blocking SQLite API calls are executed in the raft/dqlite main loop, 
  which can lead to blocking the main event loop. One solution to fix this 
  could involve the use of coroutines (to be investigated).
- Raft snapshots of the whole database are still loaded in memory in 1 chunck
  when restoring the database. This happens on startup or when installing a 
  snapshot after receiving a raft InstallSnapshot RPC. So, while memory usage
  should decrease greatly when using this mode, large memory spikes will still
  occur. `libraft` will have to be adapted to allow.
  a) Sending a snapshot in chunks, with the receiver of the snapshot 
  accumulating the chunks on disk.
  b) Restoring a snapshot in chunks.
  c) Taking a snapshot in chunks.

src/vfs.c
```diff
@@ -13,6 +18,7 @@
 #include "lib/assert.h"
 #include "lib/byte.h"
 
+#include "db.h"
```
freeekanayaka, line 21: I probably didn't look enough to the diff, but why do we need to include ```d.h```?

freeekanayaka (commented):
I just skimmed the diff and left a comment. Overall, I like the clear-cut in `vfs.c` implementation, where the new disk mode does not mix with the the old in-memory mode, so we can experiment without the risk of regression.

I'll try to give a closer look in the next days.

src/vfs.c
MathieuBordere: it's an artifact of some experimentation I think, removed.

freeekanayaka:
I believe supporting disk mode as a global on/off switch (vs a per-database one) is fine. We don't support multiple databases anyway at the moment I think.

src/lib/fs.c
```diff
@@ -0,0 +1,63 @@
+#include <ftw.h>
+#include <stdio.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "fs.h"
+#include "../tracing.h"
+
+int DirEnsure(const char *path)
```
freeekanayaka, line 10: Nit: naming convention should be `<ModuleName><FunctionName>`, in this case something like `FsDirEnsure`, or maybe `FsEnsureDir` would more readable.

src/lib/fs.c
```diff
@@ -0,0 +1,63 @@
...
+	/* Directory does not exist */
+	if (rv == -1) {
+		return mkdir(path, 0700);
+	}
+
+	return 0;
+}
+
+
+static int dir_tear_down_nftw_fn(const char *       path,
```
freeekanayaka, line 32: Nit: naming convention for local unexported functions should be `<moduleName><FunctionName>`, in this case perhaps something like `fsRemoveDirFilesNftwFn`.

src/lib/fs.c
```diff
@@ -0,0 +1,63 @@
...
+
+	/* Don't remove directory */
+	if (S_ISREG(sb->st_mode)) {
+	    rv = remove(path);
+	}
+
+	return rv;
+}
+
+int DirRemoveFiles(const char *path)
```
freeekanayaka, line 53: Nit: naming convention as above, perhaps `FsRemoveDirFiles` ?

freeekanayaka (approved):
Looks good to me, with the caveats already explained in the PR description.

MathieuBordere:
> I believe supporting disk mode as a global on/off switch (vs a per-database one) is fine. We don't support multiple databases anyway at the moment I think.

Thanks, I thought we did support multiple databases, I tested out a few cases and it appeared to be working (might have been a lucky shot, don't know). Will give it a closer look.

MathieuBordere:
I've run a Jepsen test-suite run against this branch and noticed a - very - frequent failure, will try to figure it out first before merging this.

freeekanayaka:
> > I believe supporting disk mode as a global on/off switch (vs a per-database one) is fine. We don't support multiple databases anyway at the moment I think.
> 
> Thanks, I thought we did support multiple databases, I tested out a few cases and it appeared to be working (might have been a lucky shot, don't know). Will give it a closer look.

It's kind of surprising, given the code here:

https://github.com/canonical/dqlite/blob/master/src/gateway.c#L130

but perhaps it works when using different connections.

MathieuBordere:
> > > I believe supporting disk mode as a global on/off switch (vs a per-database one) is fine. We don't support multiple databases anyway at the moment I think.
> > 
> > 
> > Thanks, I thought we did support multiple databases, I tested out a few cases and it appeared to be working (might have been a lucky shot, don't know). Will give it a closer look.
> 
> It's kind of surprising, given the code here:
> 
> https://github.com/canonical/dqlite/blob/master/src/gateway.c#L130
> 
> but perhaps it works when using different connections.

Indeed, when using multiple connections.

freeekanayaka:
> I've run a Jepsen test-suite run against this branch and noticed a - very - frequent failure, will try to figure it out first before merging this, will convert to Draft to inhibit accidental merge.

Could it be that the event look gets blocked for too long? For example when a checkpoint happens I'd expect the event loop to be blocked for a relatively long time, since we might need to write as many as 1000 pages to disk (the default checkpoint threshold is 1000).

Just thinking loud.

MathieuBordere:
> > I've run a Jepsen test-suite run against this branch and noticed a - very - frequent failure, will try to figure it out first before merging this, will convert to Draft to inhibit accidental merge.
> 
> Could it be that the event look gets blocked for too long? For example when a checkpoint happens I'd expect the event loop to be blocked for a relatively long time, since we might need to write as many as 1000 pages to disk (the default checkpoint threshold is 1000).
> 
> Just thinking loud.

I first made the implementation with sync snapshots and later converted to async snapshots, but I did a sloppy job and raft copies the WAL in a uv worker thread while it is actively being written, while it should have copied the WAL in the main thread.

MathieuBordere:
Jepsen tests are looking good, still investigating a failure that occurs with the disk-nemesis, possibly not a bug in this PR but in managing the behaviour when a raft node converts to `RAFT_UNAVAILABLE` after failing to apply a command to the state machine.

MathieuBordere:
The remaining failure in the Jepsen suite is related to https://github.com/canonical/go-dqlite/issues/213 . `apply_frames` fails because `sqlite3_open_v2` fails due to the disk being full and the node becomes unavailable. This happens to every node in the cluster, resulting in an unavailable cluster.

MathieuBordere:
I think this can be merged. The failure looks not to be related to this PR.

Issue #377, opened 2022-07-28
>255 query bindings causes query to silently not return full results


tomponline:
Originally reported from https://github.com/lxc/lxd/issues/10705

@freeekanayaka confirmed the issue is that the query binding ID is limited to 255.

https://github.com/lxc/lxd/pull/10706#issuecomment-1198258908

We should support sqlite's underlying limits:

https://www.sqlite.org/limits.html

See point 9:

>  the maximum value of a host parameter number is SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0.

And if that limit is reached fail the query rather than silently returning partial results.

tomponline:
https://github.com/canonical/go-dqlite/pull/194

cole-miller:
Going to take a stab at this.

tomponline:
@cole-miller I think @MathieuBordere was already working on it last time I asked him. You may want to pick up with him about how far he got with it.

cole-miller:
I was looking at this a bit yesterday, and I think the most straightforward way to implement the fix is to introduce new request types corresponding to `REQUEST_EXEC`, `REQUEST_EXEC_SQL`, `REQUEST_QUERY`, `REQUEST_QUERY_SQL` for which the format of the params tuple has a larger (say, 32 bits) "number of values" field. Alternatives proposed in the other thread that I considered:

- Using the message schema version field: the documentation says that message format changes that accompany a bump to this version must be backwards-compatible, only adding new fields at the end. In this case, we're increasing the size of the "number of params" field in the tuple header, which comes in the middle of the message. As far as I can tell, current dqlite doesn't check the message schema version (`flags` field) at all for requests, so old servers would just try to parse new-style requests in the old format.
- Using a variable-length encoding for the "number of params" field in the tuple header: I don't see how this can be made backwards-compatible. It's perfectly valid for the client to send a `REQUEST_EXEC` with exactly 255 params, so we can't change how dqlite servers parse a request where the "number of params" field has that value -- and similarly for every other 8-bit unsigned integer.

If we introduce new request types, then all old-style requests will continue to work, so clients like go-dqlite can introduce the new requests gradually. If a client sends one of the new requests to an old server that doesn't recognize it, it will be dropped silently, because we fall out of this switch:

[...]

PR #407, opened 2022-09-23
Support more statement parameters

cole-miller:
This implements the strategy @freeekanayaka described in #377: query and exec requests with schema version 1 use a new encoding for the parameters tuple, which stores the number of tuple entries in 4 bytes instead of 1. I'm using a fixed-length 4-byte encoding instead of a variable-length encoding because it's simpler and the size overhead is negligible (at most a couple of bytes *per request*).

Only the EXEC request is modified right now, but if this approach looks good to everyone I'll push more commits updating the EXEC_SQL, QUERY, and QUERY_SQL requests as well.


src/tuple.h
```diff
@@ -100,7 +101,7 @@ int tuple_decoder__init(struct tuple_decoder *d,
  * parameters format this is the value contained in the first byte of the tuple
  * header.
  */
-unsigned tuple_decoder__n(struct tuple_decoder *d);
+unsigned long tuple_decoder__n(struct tuple_decoder *d);
```
cole-miller, line 104: Switched to `unsigned long` here and elsewhere because that's the smallest `int` flavor that's guaranteed to be at least 4 bytes wide.

src/tuple.h
```diff
@@ -49,7 +49,7 @@
 
 #include "protocol.h"
 
-enum { TUPLE__ROW = 1, TUPLE__PARAMS };
+enum { TUPLE__ROW = 1, TUPLE__PARAMS, TUPLE__PARAMS32 };
```
freeekanayaka, line 52: For future readers, please can you include comments explaining what these formats are? In particular the difference between `TUPLE__PARAMS` and `TUPLE__PARAMS32` should be explained here.

freeekanayaka:
Looks a good approach to me (and good call for the fixed length 4-byte slot vs variable length).

* I think it's fine to use `DQLITE_PARSE` for that error.
* No other change to `EXEC` comes to my mind.
* I think it's fine either way, as long as both code paths are sufficiently tested.

cole-miller:
This should be ready to go!

freeekanayaka (approved):

test/unit/test_gateway.c
```diff
@@ -1053,6 +1055,47 @@ TEST_CASE(exec, barrier_closing, NULL)
...
+	uint64_t stmt_id;
+	size_t len = 20000;
+	char *sql = munit_malloc(len);
+	size_t pos;
+	size_t i;
+	size_t num_exec_params = 999;
+	struct value *values = munit_calloc(num_exec_params, sizeof(*values));
+	(void)params;
+
+	pos = snprintf(sql, len, "SELECT (n) FROM test WHERE n = ?");
```
MathieuBordere, line 1072: I would expect a `SELECT` statement when testing `query` not `exec`. I'm not sure if it makes a whole lot of difference in this case, but would change it for consistency with the other tests.

test/unit/test_gateway.c
```diff
@@ -1614,6 +1696,43 @@ TEST_CASE(exec_sql, barrier_error, NULL)
...
+	struct exec_sql_fixture *f = data;
+	size_t len = 20000;
+	char *sql = munit_malloc(len);
+	size_t pos;
+	size_t i;
+	size_t num_exec_params = 999;
+	struct value *values = munit_calloc(num_exec_params, sizeof(*values));
+	(void)params;
+
+	pos = snprintf(sql, len, "SELECT (n) FROM test WHERE n = ?");
```
MathieuBordere, line 1712: same remark as in `exec` case.

test/unit/test_gateway.c
cole-miller: Updated to use `DELETE FROM` instead

PR #411, opened 2022-09-27
tuple: Tweak assertion to avoid warning on LP32 archs

cole-miller:


MathieuBordere:
It probably starts to make sense to have a 32 bit CI step.

cole-miller:
I was thinking the same, but I don't know how to set that up on GHA -- all the built-in runners are x86_64.

MathieuBordere:
There's https://github.com/marketplace/actions/run-on-architecture but @stgraber is not a big fan of it. It might be enough for our use case though, catching obvious errors.

cole-miller:
Discussion continues at #412

PR #415, opened 2022-10-04
Proof-of-concept for instrumentation with bpftrace

cole-miller:
This adds a single, simple script that can be used to run bpftrace against a dqlite app and record a histogram of the time spent by each raft request between enqueueing and dequeueing. Example usage:

```sh
$ make integration-test
$ sudo bt/request -I ../raft/include -c ./integration-test
```

Notes:

- Ubuntu Jammy currently packages bpftrace 0.14.0, which is unable to find some system headers on its own (this is fixed in 0.15.0); so you may have to add `-I /usr/include/x86_64-linux-gnu` to that invocation
- bpftrace knows about some libc headers, but not yet stdbool.h, so we vendor that ourselves for the moment
- raft and dqlite need to be built with debug info (`--enable-debug`)
- The environment variable `LIBRAFT_SO_PATH` is supported to point bpftrace in the right direction in case libraft.so is installed somewhere other than /usr/local/lib (a similar customization point will be needed for libdqlite.so in scripts that probe dqlite function calls)

### DIFF
--- a/bt/request
+++ b/bt/request
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+set -o errexit
+
+libraft_path="${LIBRAFT_SO_PATH:-/usr/local/lib/libraft.so.2}"
+exec bpftrace -I resources -I include $@ - <<EOF
+#include <raft.h>
+
+struct request
+{
+    void *data;
+    int type;
+    raft_index index;
+    void *queue[2];
+};
+
+uprobe:$libraft_path:lifecycleRequestStart
+{
+	\$req = (struct request *)arg1;
+	@start_request[\$req->data, \$req->type, \$req->index] = nsecs;
+}
+
+uprobe:$libraft_path:lifecycleRequestEnd
+{
+	\$req = (struct request *)arg1;
+	\$start = @start_request[\$req->data, \$req->type, \$req->index];
+	\$end = nsecs;
+	@full[\$req->data, \$req->type, \$req->index] = (\$start, \$end);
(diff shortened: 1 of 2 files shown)


cole-miller:
(See https://github.com/canonical/raft/pull/316 for the implementation of `lifecycleRequestStart` and `lifecycleRequestEnd`)

MathieuBordere:
cool! 

I had to install `gcc-multilib` to get rid of this error on jammy:
```
/usr/include/stdio.h:27:10: fatal error: 'bits/libc-header-start.h' file not found
```

MathieuBordere (approved):

cole-miller:
> I had to install gcc-multilib to get rid of this error on jammy:

Ah, that's the same error I was seeing without `-I /usr/include/x86_64-linux-gnu`. I guess the gcc-multilib install copies those header files to somewhere that's on bpftrace's default path (directly under /usr/include?).

MathieuBordere:
> > I had to install gcc-multilib to get rid of this error on jammy:
> 
> That's the same error I was seeing without `-I /usr/include/x86_64-linux-gnu`. I guess the gcc-multilib install copies those header files to somewhere that's on bpftrace's default search path (directly under /usr/include?).

Ow sorry, read over that part of the description.

cole-miller:
I'm currently trying to get bpftrace's support for profiling by tracing to work; having trouble getting useful stack traces, presumably some debug symbols are missing but I haven't figured out where yet.

PR #431, opened 2022-11-15
Implement database drop operation

### LINKED ISSUE #422: Support deletion of database

cole-miller:
WIP (still needs tests) but feedback appreciated.

This PR implements a new wire protocol request for dqlite that allows a client to *drop* or delete an entire database, so that it will appear empty when next opened. Some other SQL databases have a `DROP my_db` statement that does this, but SQLite does not, so we implement the behavior out-of-band by just deleting the backing database file and WAL.

Design notes:

- The new REQUEST_DROP is parametrized by a database ID, not a database name, and so clients can only drop a database that they'd previously opened (with REQUEST_OPEN). This avoids a class of error conditions where the client tries to drop a non-existent database and generally makes the mental model cleaner (for me, at least).
- Leaders will not drop a database if connections other than the requesting client's are open. We set a `dropping` flag on the db object to prevent clients from opening a DB that another client has requested to drop.
- There is a new COMMAND_DROP log entry, and applying this log entry is the trigger for deleting the database "files" (via the VFS xDelete callback which we control). Therefore, dropping is asynchronous, just as though we were executing some SQL.

Questions/places where I'd like feedback:

- Do we need to take additional measures in apply_drop or elsewhere to prevent xDelete from stomping on other database operations? My mental model is that we always apply entries in log order and so there's no concurrency to worry about, but I'm not so clear on the interaction with snapshotting and checkpointing.
- Similarly, is the `dropping` flag sufficient to protect against other stuff interleaving with handle_drop and dropApplyCb? Should it be an atomic? Do we need to take some kind of lock on the database itself?
- Do we need to worry about closing/clearing the db object's follower connection?
- Is this going to have weird interactions with disk mode?

Closes #422


MathieuBordere:
> Do we need to take additional measures in apply_drop or elsewhere to prevent xDelete from stomping on other database operations? My mental model is that we always apply entries in log order and so there's no concurrency to worry about, but I'm not so clear on the interaction with snapshotting and checkpointing.

A snapshot, while it is being taken, holds a copy of the WAL and pointers to database pages. Deleting the database pages while raft is accessing pointers to the DB pages will cause issues.

Checkpointing happens synchronously in the main loop after frames have been applied to the WAL. I think that should be fine.

> Similarly, is the dropping flag sufficient to protect against other stuff interleaving with handle_drop and dropApplyCb? Should it be an atomic? Do we need to take some kind of lock on the database itself?

I don't think it should be atomic as long as all accesses to the variable happen from the same thread. I think we should be able to acquire all locks like in `maybeCheckpoint` just before deleting the DB (as all connections are closed, this should succeed)

> Do we need to worry about closing/clearing the db object's follower connection?

I think it's okay, the follower connection is always immediately closed after usage from what I can see. I think you can assert `db->follower == NULL` where applicable.

> Is this going to have weird interactions with disk mode?

Not that I can think of, but let's make sure to test it.

MathieuBordere:
> Unfortunately, there's another issue: a follower may have a snapshot in progress when it comes time to apply the drop command, and there's no way to back out the entry at that point. I'm not sure what to do about this.

This begins to sound like hackery, but we could mark the database for deletion and garbage-collect it somewhere else once the snapshot finishes.

freeekanayaka:
> > Unfortunately, there's another issue: a follower may have a snapshot in progress when it comes time to apply the drop command, and there's no way to back out the entry at that point. I'm not sure what to do about this.
> 
> This begins to sound like hackery, but we could mark the database for deletion and garbage-collect it somewhere else once the snapshot finishes.
> 
> Or fail the application of the log entry containing the deletion, but in such a way that raft will retry it at some point. It's not a disaster if a follower's FSM is not up-to-date I think, when it becomes leader, it will update it's FSM.
> 
> The second approach is more sane imo, but it still doesn't feel nice.
>

Fail the application of the log entry is going to be complicated IMO, because I don't think the raft library is quite prepared for that. Failing a log entry should be used only as emergency, like an inconsistency is detected or any other hard unresolvable situations (while this one is a transient problem).

Garbage collecting seems a tad better, although also not that nice.
 
> Maybe we could look for other ways to obtain the same desired effect for the user, "dropping a database" could mean completely emptying it, and maybe that's easier.

I'd be in favor of emptying the database instead. Effectively it should be the same thing, because if you try to open it again it does make a difference for the user if it's not there at all (deleted) or if it's just there but empty. I believe we create databases automatically when you try to open them, so it seems emptying the db would be completely transparent from the user point of view.

PR #434, opened 2022-11-17
fsm: Don't ignore return value in fsm__apply.

MathieuBordere:
We were ignoring all `apply_xxx` return values except in the `switch` `default:` branch.


MathieuBordere:
I'll still try to add a test for this one, but it's a pretty obvious bug.

PR #435, opened 2022-11-18
Introduce RESET request

### LINKED ISSUE #422: Support deletion of database

cole-miller:
This PR implements a new RESET request that allows a dqlite client to restore the database it has opened to a pristine state. Instead of deleting the in-memory database "file", we use the SQLITE_DBCONFIG_RESET_DATABASE method described [here](https://sqlite.org/c3ref/c_dbconfig_defensive.html), which goes through the normal VFS channels. This still requires a separate request because of the out-of-band sqlite3_db_config calls. The implementation of the new request mimics what would happen if we did `EXEC_SQL("VACUUM")`, with some simplifications since we don't need to run a barrier or bind parameters.

Closes #422

### DIFF
--- a/src/leader.c
+++ b/src/leader.c
@@ -467,7 +467,7 @@ static void resetBarrierCb(struct barrier *barrier, int status)
 		goto finish;
 	}
 
-#if SQLITE_VERSION_NUMBER >= 302400
+#if SQLITE_VERSION_NUMBER >= 3024000
 	vfs = sqlite3_vfs_find(l->db->config->name);
 	assert(vfs != NULL);
 	usage = VfsDatabaseMemoryUsage(vfs, l->db->filename);
@@ -480,6 +480,7 @@ static void resetBarrierCb(struct barrier *barrier, int status)
 	rv = sqlite3_db_config(l->conn, SQLITE_DBCONFIG_RESET_DATABASE, 1, 0);
 	assert(rv == 0);
 	exec->status = sqlite3_exec(l->conn, "VACUUM", NULL, NULL, NULL);
+	tracef("reset barrier cb exec status:%d", exec->status);
 	rv = sqlite3_db_config(l->conn, SQLITE_DBCONFIG_RESET_DATABASE, 0, 0);
 	assert(rv == 0);
 
@@ -509,6 +510,7 @@ static void resetBarrierCb(struct barrier *barrier, int status)
 	tracef("reset barrier cb not supported");
 	(void)vfs;
 	(void)frames;
+	(void)usage;
 	(void)n;
 	(void)i;
 	rv = DQLITE_ERROR;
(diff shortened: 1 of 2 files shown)


freeekanayaka:
> Ah, it looks like SQLITE_DBCONFIG_RESET_DATABASE was added in SQLite v3.24.0, which is not available on Bionic. We could do some `#ifdef` stuff to detect an old SQLite and just fail RESET requests in this case?

+1

freeekanayaka:
> sqlite3_table_column_metadata

I'm not entirely sure why this requirement is there, but maybe you can use `sqlite3_table_column_metadata` against one of the virtual tables/columns that are guaranteed to be there, e.g. `sqlite3_table_column_metadata(db, NULL, "sqlite_master", "name", ...)`.

MathieuBordere:
> Ah, it looks like SQLITE_DBCONFIG_RESET_DATABASE was added in SQLite v3.24.0, which is not available on Bionic. We could do some `#ifdef` stuff to detect an old SQLite and just fail RESET requests in this case?

I would maybe check it at runtime, with `sqlite3_version` or `sqlite3_libversion`.

src/leader.c
```diff
@@ -450,3 +450,49 @@ int leader__barrier(struct leader *l, struct barrier *barrier, barrier_cb cb)
 	}
 	return 0;
 }
+
+int leader__reset(struct leader *l, struct exec *exec, exec_cb cb)
```
MathieuBordere, line 454: We probably want to go through a barrier before executing operations against the DB?

src/leader.c
```diff
@@ -450,3 +450,49 @@ int leader__barrier(struct leader *l, struct barrier *barrier, barrier_cb cb)
...
+	exec->leader = l;
+	memset(&exec->barrier, 0, sizeof(exec->barrier));
+	exec->stmt = NULL;
+	exec->cb = cb;
+	l->exec = exec;
+
+	rv = sqlite3_db_config(l->conn, SQLITE_DBCONFIG_RESET_DATABASE, 1, 0);
+	assert(rv == 0);
+	exec->status = sqlite3_exec(l->conn, "VACUUM", NULL, NULL, NULL);
+	rv = sqlite3_db_config(l->conn, SQLITE_DBCONFIG_RESET_DATABASE, 0, 0);
```
MathieuBordere, line 471: Out of curiosity, does dqlite's memory footprint shrink after resetting a DB? If it's not shrinking, we should look into it.

MathieuBordere: Does the `VACUUM` actually succeed?

MathieuBordere:
Okay, we should first figure out why the database is not shrinking before merging this.

src/leader.c
freeekanayaka: > Hmm, I don't see any change in number of DB pages, number of WAL frames, or number of shm regions after leaderApplyFrames returns vs. before `VACUUM`. @freeekanayaka, can you think of anything that would be preventing our VFS from dropping allocated memory when SQLite is done using it? (I could also just be looking in the wrong places -- I don't fully understand the VFS subsystem yet.)

Not on the top of my head, but I also don't have a clear picture of the changes that are being made. The SQLite memory allocator has some helper functions to track the amount of allocated memory, perhaps that could be useful to check.

PR #438, opened 2022-11-24
fsm: Explicitly cast uint64_t to size_t

MathieuBordere:
Fixes 32bit build. 

There's a deeper issue in the raft snapshot restore API that limits the database size to what fits in a single buffer of `size_t` bytes.
This would need to be fixed in a separate PR as it requires more thorough work.


MathieuBordere:
I'm just going to merge this once tests succeed as it fixes Launchpad builds.

Issue #432, opened 2022-11-16
row size limit? getting "Error: disk I/O error" from dqlite-demo


nickchen:
is there a row byte size limit? 

libuv=1.44.2, libraft=0.15.0, sqlite3=3.39.4, dqlite=1.11.1

running the dqlite-demo with following

# server
```
CGO_LDFLAGS_ALLOW="-Wl,-z,now" CGO_ENABLED=1 go run  dqlite-demo.go --api 127.0.0.1:8001 --db 127.0.0.1:9001 -v
```

# client

```
# base64 /dev/urandom | head -c 20000000 > file.txt
#  curl -X PUT -T "file.txt" http://127.0.0.1:8001/my-key-1
Error: disk I/O error
```

MathieuBordere:
Thanks for your submission, we'll have a look.

MathieuBordere:
> It's this one, I think:
> 
> https://github.com/canonical/dqlite/blob/fdf7742abbd396204e6faeea30e80bd71f59eec8/src/leader.c#L343
> 
> Trying to step into that call but GDB is being difficult...

The err is `SQLITE_IOERR_SHORT_READ`, try to look for that in dqlite's codebase, might give a clue.

nickchen:
just checking in to see if there are any progress with this?

nickchen:
nice! when's the planned release for this?

MathieuBordere:
I would have liked it to simmer a bit longer, we could do a release early next week?

PR #444, opened 2022-12-05
Support building sqlite3 from amalgamation

cole-miller:
This adds a `--enable-build-sqlite` option to our configure.ac. In this mode, the build will look for sqlite3.c in the build root and just add that to libdqlite_la_SOURCES, instead of linking with `-lsqlite3`.

The build system doesn't take care of fetching the SQLite amalgamation -- the idea is to obtain an SQLite source archive/checkout and do `make sqlite3.c`, or download the amalgamation from sqlite.org. It's important to use a version of the amalgamation that matches the version of sqlite.h that's on your include path.

Our build turns on a bunch of warnings that fire on the SQLite source code; I had to modify the amalgamation by adding

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wsign-conversion"
#pragma GCC diagnostic ignored "-Wconversion"
#pragma GCC diagnostic ignored "-Wfloat-equal"
#pragma GCC diagnostic ignored "-Wfloat-conversion"
#pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
```

at the top and

```
#pragma GCC diagnostic pop
```

at the bottom.

The point is to be able to easily add printfs in SQLite code, turn on SQLITE_DEBUG, etc., to support debugging issues like #432 that arise from the interaction between SQLite and our VFS.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,11 @@
 ACLOCAL_AMFLAGS = -I m4
 AM_CFLAGS += $(CODE_COVERAGE_CFLAGS)
 AM_CFLAGS += $(SQLITE_CFLAGS) $(UV_CFLAGS) $(RAFT_CFLAGS) $(PTHREAD_CFLAGS)
-AM_LDFLAGS = $(SQLITE_LIBS) $(UV_LIBS) $(RAFT_LIBS) $(PTHREAD_LIBS)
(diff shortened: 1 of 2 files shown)


MathieuBordere (approved):
Cool, should come in handy

MathieuBordere:
can you fix the merge conflict please @cole-miller ?

PR #445, opened 2022-12-06
vfs: Fix the short read check in vfsWalRead

### LINKED ISSUE #432: row size limit? getting "Error: disk I/O error" from dqlite-demo

cole-miller:
### Problem

When trying to read a page from the WAL, we bail early with an error if there are no committed pages in `struct vfsWal`. But sometimes a transaction will want to read its own WAL frames that haven't been committed yet, and this check will prevent that if the transaction starts out with an empty WAL (because it's running right after a checkpoint, for example).

### Solution

Remove the separate check for `w->n_frame == 0` and just see whether `vfsWalFrameLookup` returns `NULL`, indicating that there's no frame at the given index. `vfsWalFrameLookup` can retrieve frames from either `w->frames` or `w->tx`.

Closes #432 

### DIFF
--- a/test/integration/test_cluster.c
+++ b/test/integration/test_cluster.c
@@ -170,3 +170,33 @@ TEST(cluster, dataOnNewNode, setUp, tearDown, 0, cluster_params)
 	clientCloseRows(&rows);
 	return MUNIT_OK;
 }
+
+/* Insert a huge row, causing SQLite to allocate overflow pages. Then
+ * insert the same row again. (Reproducer for https://github.com/canonical/raft/issues/432.) */
+TEST(cluster, hugeRow, setUp, tearDown, 0, NULL)
+{
+	struct fixture *f = data;
+	unsigned stmt_id;
+	unsigned last_insert_id;
+	unsigned rows_affected;
+	char *sql;
+	ssize_t n;
+	size_t huge = 20000000;
+	(void)params;
+
+	HANDSHAKE;
+	OPEN;
+	PREPARE("CREATE TABLE IF NOT EXISTS model(key TEXT, value TEXT, UNIQUE(key))", &stmt_id);
+	EXEC(stmt_id, &last_insert_id, &rows_affected);
+	sql = munit_malloc(huge);
+	n = snprintf(sql, huge, "INSERT OR REPLACE INTO model (key, value) VALUES('my-key-1', '");
+	memset(sql + n, 'A', huge - n);
+	memcpy(sql + huge - 3, "')", 3);
+	PREPARE(sql, &stmt_id);
+	free(sql);
+	EXEC(stmt_id, &last_insert_id, &rows_affected);
+	/* Again */
+	EXEC(stmt_id, &last_insert_id, &rows_affected);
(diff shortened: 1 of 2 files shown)


freeekanayaka:
> One concern I had is that if a transaction is pending on one connection, other connections should definitely not be able to read uncommitted WAL data from that transaction; vfsWalRead itself doesn't have any checks around this, before or after this change. I think it should be taken care of by each connection having to hold the appropriate locks before trying to read from the WAL -- is that right?

That's right, it's not a VFS concern. SQLite takes care of it by taking note of the highest committed WAL frame number at the time the transaction started, and never reading past it.

test/integration/test_cluster.c
```diff
@@ -170,3 +170,31 @@ TEST(cluster, dataOnNewNode, setUp, tearDown, 0, cluster_params)
...
+
+TEST(cluster, hugeRow, setUp, tearDown, 0, NULL)
+{
+	struct fixture *f = data;
+	unsigned stmt_id;
+	unsigned last_insert_id;
+	unsigned rows_affected;
+	char *sql;
+	ssize_t n;
+size_t huge = 20000000;
```
freeekanayaka, line 182: Missing indentation.

test/integration/test_cluster.c
```diff
@@ -170,3 +170,31 @@ TEST(cluster, dataOnNewNode, setUp, tearDown, 0, cluster_params)
...
+	sql = munit_malloc(huge);
+	n = snprintf(sql, huge, "INSERT OR REPLACE INTO model (key, value) VALUES('my-key-1', '");
+	memset(sql + n, 'A', huge - n);
+	memcpy(sql + huge - 3, "')", 3);
+	PREPARE(sql, &stmt_id);
+	free(sql);
+	EXEC(stmt_id, &last_insert_id, &rows_affected);
+	/* Again */
+	EXEC(stmt_id, &last_insert_id, &rows_affected);
+	return MUNIT_OK;
```
freeekanayaka, line 198: I'm a bit surprised that the there's no test failure or assertion violation because of the prepared statement above not being finalized.

But probably we have changed that behavior and now we automatically finalize leftover statements.

freeekanayaka (approved):

test/integration/test_cluster.c
MathieuBordere: They're finalized when the statement registry is closed here https://github.com/canonical/dqlite/blob/ee7a3cdb3a996fa798443b9df51391925a12b8d8/src/gateway.c#L72

PR #457, opened 2023-01-12
gateway: Assign IDs to exec and config requests

cole-miller:
Just a first pass at assigning request IDs in the dqlite layer that can be used in tracing. We only use the first 8 bytes of the new 16-byte req_id field from RAFT__REQUEST, and set the last byte of that field to -1 to indicate that the ID is set.

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -1,6 +1,7 @@
 #include "gateway.h"
 
 #include "bind.h"
+#include "id.h"
 #include "protocol.h"
 #include "query.h"
 #include "request.h"
@@ -13,9 +14,10 @@
 void gateway__init(struct gateway *g,
 		   struct config *config,
 		   struct registry *registry,
-		   struct raft *raft)
+		   struct raft *raft,
+		   struct id_state seed)
 {
-        tracef("gateway init");
+	tracef("gateway init");
 	g->config = config;
 	g->registry = registry;
 	g->raft = raft;
@@ -30,6 +32,7 @@ void gateway__init(struct gateway *g,
 	g->barrier.cb = NULL;
 	g->barrier.leader = NULL;
 	g->protocol = DQLITE_PROTOCOL_VERSION;
+	g->random_state = seed;
 }
 
 void gateway__leader_close(struct gateway *g, int reason)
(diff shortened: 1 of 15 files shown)


src/gateway.c
```diff
@@ -30,6 +43,9 @@ void gateway__init(struct gateway *g,
 	g->barrier.cb = NULL;
 	g->barrier.leader = NULL;
 	g->protocol = DQLITE_PROTOCOL_VERSION;
+	rv = getrandom(&g->random_state, sizeof(g->random_state), 0);
```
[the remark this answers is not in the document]
MathieuBordere: > Is it plausible that the urandom entropy pool hasn't already been initialized by the time someone gets around to running dqlite? My understanding (from reading e.g. [this LWN article](https://lwn.net/Articles/800509/)) is that this is only a potential problem while booting up.

yeah, it was just to be on the safe-side.

src/gateway.c
```diff
@@ -30,6 +33,14 @@ void gateway__init(struct gateway *g,
 	g->barrier.cb = NULL;
 	g->barrier.leader = NULL;
 	g->protocol = DQLITE_PROTOCOL_VERSION;
+	g->random_state = 0;
+	/* For seeding the ID generator, there are no security concerns, so read from
+	 * /dev/urandom instead of calling getrandom(2) to avoid potential blocking. */
+	urandom = open("/dev/urandom", O_RDONLY);
+	assert(urandom != -1);
```
MathieuBordere, line 40: Now that I think of it, shouldn't we just initialize the random state once, when the server starts? Now we have to do it for every connection and in real-life a lot of connections get set up and broken down.

src/leader.c
```diff
@@ -321,6 +324,9 @@ static int leaderApplyFrames(struct exec *req,
 	apply->leader = req->leader;
 	apply->req.data = apply;
 	apply->type = COMMAND_FRAMES;
+	memset(&apply->req.req_id, 0, sizeof(apply->req.req_id));
```
MathieuBordere, line 327: `idSet` can be used here

src/id.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include <stdint.h>
+
+struct id_state
+{
+	uint64_t data[4];
+};
+
+
+uint64_t idNext(struct id_state *state);
```
MathieuBordere, line 12: Maybe add some (short) comments on what these functions are supposed to do / used for.

MathieuBordere (approved):
Looks good!

PR #458, opened 2023-01-17
Prepare client.c for use in the client library

cole-miller:
- `src/client.{c,h} -> src/client/protocol.{c,h}` in preparation for more client source files
- Implement remaining request/response types (that are actually used by go-dqlite)
- Use up-to-date protocol/schema versions for all requests
- Improved error handling
- Timeouts for reading from the underlying fd

This will be easiest to review commit-by-commit.


src/client/protocol.c
```diff
@@ -0,0 +1,832 @@
...
+	       int timeout_first_millis,
+	       int timeout_follow_millis)
+{
+	tracef("init client");
+	int rv;
+	c->fd = fd;
+	c->db_name = NULL;
+	c->db_is_init = false;
+
+	c->timeout_first_millis = timeout_first_millis;
```
MathieuBordere, line 75: Why does the distinction between `timeout_first_millis` and `timeout_follow_millis` exist?

src/client/protocol.c
```diff
@@ -0,0 +1,832 @@
...
+/* Read data from fd into buf until one of the following occurs:
+ *
+ * - The full count n of bytes is read.
+ * - A read returns 0 (EOF).
+ * - Polling the fd in preparation for a read times out.
+ *
+ * On error, -1 is returned. Otherwise the return value is the count
+ * of bytes read. This may be less than n if either EOF happened or
+ * polling timed out. */
+static ssize_t doRead(int fd, void *buf, size_t n, int timeout_millis)
```
MathieuBordere, line 26: Maybe we can pass the client struct here instead and do some accounting on the timeout and save e.g. `timeout_left` in the client struct after measuring the time spent in `poll`?

MathieuBordere:
Did a first round of reading through the changes, think it already looks good.

src/client/protocol.c
MathieuBordere: Yes, something like that could work I think. Same then for `Send`?

freeekanayaka:
> @freeekanayaka, could you elucidate what `DQLITE_REQUEST_CONNECT` is for? It seems that it's not used at all by go-dqlite, so I didn't implement it here. There's still a codepath in dqlite to handle such requests -- perhaps it could be removed?

It's basically used as glue between dqlite and raft. If you use dqlite without a custom transport, then when raft needs to establish a connection to another raft node it will use the default dqlite transport defined in `transport.c` which basically connects to the desired dqlite engine running the raft code and sends the `DQLITE_REQUEST_CONNECT` command to tell the dqlite engine to handover the connection to the raft code.

It's implicitly used by go-dqlite as well via `dqlite_node_set_connect_func`.

I hope my memory doesn't fail me and that everything is correct and clear :) Please let me know if something doesn't match.

freeekanayaka:
> @freeekanayaka Got it, thanks! I'll add this to the docs.
> 
> What's the motivation for proxying raft connections through dqlite, instead of having them flow independently? I guess it's convenient to have just one address that both dqlite and raft are listening on.

Sorry, I didn't see your follow up before. Yes, it's to have a single endpoint.

src/client/protocol.c
```diff
@@ -14,67 +14,258 @@
 #include "../tracing.h"
 #include "../tuple.h"
 
+/* Convert a value that potentially borrows data from the client_proto read buffer
+ * into one that owns its data. The owned data must be free with freeOwnedValue. */
+static int makeValueOwned(struct value *val)
+{
+	char *p;
+	switch (val->type) {
```
MathieuBordere, line 22: I like a `default` case in a switch statement, I also think Coverity will complain if it's absent.

src/client/protocol.c
```diff
@@ -14,67 +14,258 @@
...
+			break;
+	}
+	return 0;
+}
+
+/* Free the owned data of a value, which must have had makeValueOwned called
+ * on it previously. */
+static void freeOwnedValue(struct value val)
+{
+	switch (val.type) {
```
MathieuBordere, line 53: Also please add a `default` case.

src/client/protocol.c
```diff
@@ -14,67 +14,258 @@
...
+			break;
+		case DQLITE_ISO8601:
+			p = strdup(val->iso8601);
+			if (p == NULL) {
+				return DQLITE_CLIENT_PROTO_ERROR;
+			}
+			val->iso8601 = p;
+			break;
+		case SQLITE_BLOB:
+			p = malloc(val->blob.len);
```
MathieuBordere, line 38: Just letting up a balloon, an opinion from a personal hero ;-) https://twitter.com/antirez/status/1619369355833405441?s=20&t=nG9lOE2osQLVHC6PtV2E-Q . Not saying we should do anything like that right now.

src/client/protocol.c
```diff
@@ -650,16 +865,19 @@ int clientSendDescribe(struct client_proto *c)
...
 	return 0;
 }
 
-int clientRecvServer(struct client_proto *c, uint64_t *id, char **address)
+int clientRecvServer(struct client_proto *c,
+		     uint64_t *id,
+		     char **address,
+		     int *budget_millis)
 {
 	tracef("client recv server");
```
MathieuBordere, line 882: indentation is off here (not your fault)

src/client/protocol.c
```diff
@@ -675,7 +893,7 @@ int clientRecvServer(struct client_proto *c, uint64_t *id, char **address)
 	return 0;
 }
 
-int clientRecvWelcome(struct client_proto *c)
+int clientRecvWelcome(struct client_proto *c, int *budget_millis)
 {
 	tracef("client recv welcome");
```
MathieuBordere, line 898: indentation off

src/client/protocol.c
```diff
@@ -704,7 +925,10 @@ int clientRecvFailure(struct client_proto *c, uint64_t *code, const char **msg)
 	return 0;
 }
 
-int clientRecvServers(struct client_proto *c, struct client_node_info **servers, size_t *n_servers)
+int clientRecvServers(struct client_proto *c,
+		      struct client_node_info **servers,
+		      size_t *n_servers,
+		      int *budget_millis)
 {
 	tracef("client recv servers");
```
MathieuBordere, line 933: indentation (it's probably off in most of the similar functions)

.gitignore
```diff
@@ -25,3 +25,5 @@ integration-test
 dqlite.pc
 libtool
 stamp-h*
+sqlite3.c
+configure~
```
MathieuBordere, line 29: A user should probably ignore `~` globally instead of here?

MathieuBordere (commented):
Mostly nits :-)

Issue #418, opened 2022-10-18
dqlite max database size is limited to 4GB


MathieuBordere:
dqlite is limited to databases not exceeding 2^32 Bytes in size, possibly related to sqlite's memory allocator.

MathieuBordere:
Each write should check if the resulting database size is not larger than `SIZE_MAX` because otherwise we will not be able to restore the database after that it has been snapshotted.

marco6:
`SIZE_MAX` is `2^64` in 64-bit systems, while `2^32` in 32-bit systems.

I know no real-world processor that can do more than 57-bits of addresses (the upper part usually *must* be a sign extension). Even riscv is going for 57 and does not generally allow pointer tagging. This means that even if that check was performed, the process would exhaust its *address space* long before that limit is reached (likely going OOM way before that).

On 32-bit, you can address the whole space, but in reality the exhaustion will still happen (even on PAE-enabled systems, where you have 32-bit worth of data for the application) as the process and the rest of the application will still map a part of it for the text sections, the static variables, the stack etc etc.

Closing this.

PR #459, opened 2023-01-19
Fix integer overflow issues that limited database size.

MathieuBordere:
Together with https://github.com/canonical/raft/pull/371 fixes the problems that limit the database size to `2^32` bytes. Now the practical database size of dqlite is limited to `SIZE_MAX` or whatever fits in the single `raft_buffer` of
```
struct raft_fsm
{   
    ...      
    int (*restore)(struct raft_fsm *fsm, struct raft_buffer *buf); 
    ...      
};  
```

The `SIZE_MAX` size limitation is tracked in #418 (previously used to track the size limit of 4GB).


src/vfs.c
```diff
@@ -2598,7 +2603,8 @@ static int vfsDatabaseRestore(struct vfsDatabase *d,
 			goto oom_after_pages_alloc;
 		}
 		pages[i] = page;
-		memcpy(page, &data[i * page_size], page_size);
+		offset = (size_t)i * (size_t)page_size;
```
MathieuBordere, line 2606: This indexes into a data buffer of maximum size `SIZE_MAX` and we already check here https://github.com/canonical/dqlite/blob/5d00a628786a4c078e66e4c8dec9e8232e440d9f/src/fsm.c#L464 if the buffer is smaller than `SIZE_MAX`, so this should not overflow.

src/vfs.c
```diff
@@ -2656,7 +2663,8 @@ static int vfsWalRestore(struct vfsWal *w,
 		}
 		frames[i] = frame;
 
-		p = &data[VFS__WAL_HEADER_SIZE + i * vfsFrameSize(page_size)];
+		offset = (size_t)VFS__WAL_HEADER_SIZE + ((size_t)i * (size_t)vfsFrameSize(page_size));
+		p = &data[offset];
```
MathieuBordere, line 2667: This indexes into a data buffer of maximum size `SIZE_MAX` and we already check here https://github.com/canonical/dqlite/blob/5d00a628786a4c078e66e4c8dec9e8232e440d9f/src/fsm.c#L464 if the buffer is smaller than `SIZE_MAX`, so this should not overflow. (same reason as above)

src/vfs.c
```diff
@@ -2711,7 +2719,7 @@ int VfsRestore(sqlite3_vfs *vfs,
 	}
 
 	page_size = vfsDatabaseGetPageSize(database);
-	offset = database->n_pages * page_size;
+	offset = (size_t)database->n_pages * (size_t)page_size;
```
MathieuBordere, line 2722: Same reason as above, this multiplication should not overflow `SIZE_MAX`

PR #473, opened 2023-02-07
github: Remove deprecated Ubuntu Bionic

cole-miller:


stgraber:
There doesn't appear to be 22.10 runners. I guess they only provide LTS releases.

PR #474, opened 2023-02-07
gateway: Error when prepared SQL has ignored tail

### LINKED ISSUE #461: PREPARE request discards "tail" information
sqlite3_prepare_v2 and friends only compile the *first* statement from the input SQL string and store a pointer to the rest of the string in an out parameter `**pzTail`. Right now we implement dqlite's PREPARE request with a single call to sqlite3_prepare_v2, discarding the tail. That means if a client tries to prepare a string like

```sql
INSERT INTO foo (n) VALUES (1); INSERT INTO foo (n) VALUES (2)
```

the actual prepared statement will only cover up to the semicolon, but this will not be signalled to the client in any way. This is kind of a footgun on its own, and also makes implementing our own `dqlite_prepare` porcelain (with an analogous signature to sqlite3_prepare_v2) difficult.

What I think we should do:

- Cause dqlite servers to emit an error when they get a PREPARE request containing more than one statement.
- Introduce a new request/response pair (or use the schema version field -- not my preference) that lets the server pass back to the client how much of an SQL string was prepared.

Edit: Worth noting that sqlite3_prepare_v2 supports passing NULL for `pzTail`, in which case the "tail" information is just discarded. So our current PREPARE request is the equivalent of always passing NULL to the sqlite3 function.

cole-miller:
Closes #461 

### DIFF
--- a/src/client/protocol.c
+++ b/src/client/protocol.c
@@ -502,17 +502,25 @@ int clientSendPrepare(struct client_proto *c, const char *sql, struct client_con
 	struct request_prepare request;
 	request.db_id = c->db_id;
 	request.sql = sql;
-	REQUEST(prepare, PREPARE, 0);
+	REQUEST(prepare, PREPARE, DQLITE_PREPARE_STMT_SCHEMA_V1);
 	return 0;
 }
 
-int clientRecvStmt(struct client_proto *c, uint32_t *stmt_id, struct client_context *context)
+int clientRecvStmt(struct client_proto *c,
+			uint32_t *stmt_id,
+			uint64_t *offset,
+			struct client_context *context)
 {
 	struct cursor cursor;
-	struct response_stmt response;
-	RESPONSE(stmt, STMT);
-	*stmt_id = response.id;
-	tracef("client recv stmt stmt_id %" PRIu32, *stmt_id);
+	struct response_stmt_with_offset response;
+	RESPONSE(stmt_with_offset, STMT_WITH_OFFSET);
+	tracef("client recv stmt stmt_id:%" PRIu32 " offset:%" PRIu64, response.id, response.offset);
+	if (stmt_id != NULL) {
+		*stmt_id = response.id;
+	}
+	if (offset != NULL) {
+		*offset = response.offset;
+	}
 	return 0;
 }
 
--- a/src/conn.c
+++ b/src/conn.c
@@ -49,13 +49,15 @@ abort:
 	conn__stop(c);
 }
 
-static void gateway_handle_cb(struct handle *req, int status, int type)
+static void gateway_handle_cb(struct handle *req, int status, uint8_t type, uint8_t schema)
 {
 	struct conn *c = req->data;
 	size_t n;
 	void *cursor;
 	uv_buf_t buf;
 	int rv;
+
+	assert(schema <= req->schema);
 
 	/* Ignore results firing after we started closing. TODO: instead, we
 	 * should make gateway__close() asynchronous. */
@@ -72,9 +74,9 @@ static void gateway_handle_cb(struct handle *req, int status, int type)
 	n = buffer__offset(&c->write) - message__sizeof(&c->response);
 	assert(n % 8 == 0);
 
-	c->response.type = (uint8_t)type;
+	c->response.type = type;
 	c->response.words = (uint32_t)(n / 8);
-	c->response.schema = 0;
+	c->response.schema = schema;
 	c->response.extra = 0;
 
 	cursor = buffer__cursor(&c->write, 0);
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -299,8 +299,7 @@ static void prepareBarrierCb(struct barrier *barrier, int status)
 		return;
 	}
 
-	rc = sqlite3_prepare_v2(g->leader->conn, sql, -1, &stmt->stmt,
-				&tail);
+	rc = sqlite3_prepare_v2(g->leader->conn, sql, -1, &stmt->stmt, &tail);
 	if (rc != SQLITE_OK) {
 		tracef("prepare barrier cb sqlite prepare failed %d %s", rc, sqlite3_errmsg(g->leader->conn));
 		stmt__registry_del(&g->stmts, stmt);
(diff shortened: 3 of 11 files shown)


src/client/protocol.c
```diff
@@ -502,17 +502,25 @@ int clientSendPrepare(struct client_proto *c, const char *sql, struct client_con
 	struct request_prepare request;
 	request.db_id = c->db_id;
 	request.sql = sql;
-	REQUEST(prepare, PREPARE, 0);
+	REQUEST(prepare, PREPARE, 1);
```
MathieuBordere, line 505: I would probably prefer a macro instead of `1`.

src/conn.c
```diff
@@ -72,9 +72,9 @@ static void gateway_handle_cb(struct handle *req, int status, int type)
 	n = buffer__offset(&c->write) - message__sizeof(&c->response);
 	assert(n % 8 == 0);
 
-	c->response.type = (uint8_t)type;
+	c->response.type = type;
 	c->response.words = (uint32_t)(n / 8);
-	c->response.schema = 0;
+	c->response.schema = schema;
```
MathieuBordere, line 77: Useful to assert that the response schema is equal or less than the `req->schema`?

src/gateway.c
```diff
@@ -115,17 +115,17 @@ void gateway__close(struct gateway *g)
 	}
 
 /* Encode the given success response and invoke the request callback */
-#define SUCCESS(LOWER, UPPER)                                                  \
+#define SUCCESS_V0(LOWER, UPPER)                                               \
```
MathieuBordere, line 118: On the first pass it wasn't clear that the ` _V0` suffix is there to indicate the schema version, it could use some clarification.

Why don't we have a `SUCCESS_V1`?

src/gateway.c
```diff
@@ -115,17 +115,17 @@ void gateway__close(struct gateway *g)
...
 		void *_cursor;                                                 \
 		assert(_n % 8 == 0);                                           \
 		_cursor = buffer__advance(req->buffer, _n);                    \
-		/* Since responses are small and the buffer it's at least 4096 \
+		/* Since responses are small and the buffer it's at least 4096
 		 * bytes, this can't fail. */                                  \
 		assert(_cursor != NULL);                                       \
 		response_##LOWER##__encode(&response, &_cursor);               \
-		req->cb(req, 0, DQLITE_RESPONSE_##UPPER);                      \
+		req->cb(req, 0, DQLITE_RESPONSE_##UPPER, 0);                   \
```
MathieuBordere, line 128: I'd prefer a `SCHEMA_V0` instead of `0` (like above)

src/gateway.c
```diff
@@ -316,10 +318,35 @@ static void prepareBarrierCb(struct barrier *barrier, int status)
...
+			response.params = (uint64_t)sqlite3_bind_parameter_count(stmt->stmt);
+			SUCCESS_V0(stmt, STMT);
+			break;
+		case 1:
+			response_with_offset.db_id = (uint32_t)req->db_id;
+			response_with_offset.id = (uint32_t)stmt->id;
+			response_with_offset.params = (uint64_t)sqlite3_bind_parameter_count(stmt->stmt);
+			response_with_offset.offset = (uint64_t)(tail - sql);
+
+			n = response_stmt_with_offset__sizeof(&response_with_offset);
```
MathieuBordere, line 340: Is it not possible to use a `SUCCES_V1`-like macro here?

MathieuBordere (commented):
Functionality looks good, but I would at least expect some tests with/without tail, that test the actual functionality in this PR and with the different schema versions.

src/gateway.c
MathieuBordere: If you can create some macro that's reusable, we can use it for that too, yes. Now there's some code duplication, and it just feels it could be handled somewhat cleaner.

src/client/protocol.c
```diff
@@ -502,17 +502,25 @@ int clientSendPrepare(struct client_proto *c, const char *sql, struct client_con
 	struct request_prepare request;
 	request.db_id = c->db_id;
 	request.sql = sql;
-	REQUEST(prepare, PREPARE, 0);
+	REQUEST(prepare, PREPARE, DQLITE_PREPARE_STMT_SCHEMA_V1);
```
MathieuBordere, line 505: Maybe something to think about later, but what should happen when a dqlite server doesn't know a certain schema version? Should the client try to downgrade the request? In case of a prepare stmt, the downgrade to `V0` should work in most cases as I expect that most of the SQL that has to be prepared will only contain a single statement.

src/gateway.c
```diff
@@ -784,11 +814,16 @@ static void querySqlBarrierCb(struct barrier *barrier, int status)
 		return;
 	}
 
+	if (*tail != '\0') {
```
MathieuBordere, line 817: Can you test this too please?

test/unit/test_gateway.c
```diff
@@ -592,6 +594,41 @@ TEST_CASE(prepare, non_leader, NULL)
...
+	f->request.db_id = 0;
+	f->request.sql = "CREATE TABLE test (n INT); SELECT * FROM test";
+	CLUSTER_ELECT(0);
+	ENCODE(&f->request, prepare);
+	HANDLE_SCHEMA_STATUS(DQLITE_REQUEST_PREPARE, 1, 0);
+	WAIT;
+	ASSERT_CALLBACK(0, STMT_WITH_OFFSET);
+	DECODE(&response, stmt_with_offset);
+	munit_assert_int(response.id, ==, 0);
+	munit_assert_uint64(response.offset, ==, 26);
```
MathieuBordere, line 628: I'm being nitty here, but it would be nice if we used the offset returned by dqlite to index into the original SQL string and try to successfully prepare the next statement.

src/gateway.c
```diff
@@ -316,10 +319,29 @@ static void prepareBarrierCb(struct barrier *barrier, int status)
 		return;
 	}
 
-	response.db_id = (uint32_t)req->db_id;
-	response.id = (uint32_t)stmt->id;
-	response.params = (uint64_t)sqlite3_bind_parameter_count(stmt->stmt);
-	SUCCESS(stmt, STMT);
+	if (req->schema == 0 && *tail != '\0') {
```
MathieuBordere, line 322: can use `DQLITE_PREPARE_STMT_SCHEMA_V0` here too

MathieuBordere (approved):

PR #476, opened 2023-02-13
Refactor gateway.c

cole-miller:
This PR refactors gateway.c to store several pieces of request-scoped data in the request handle object instead of the gateway itself. In the process, I fixed (?) how we handle parameters to multi-statement EXEC_SQL requests.

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -26,7 +26,6 @@ void gateway__init(struct gateway *g,
 	g->stmt = NULL;
 	g->stmt_finalize = false;
 	g->exec.data = g;
-	g->sql = NULL;
 	stmt__registry_init(&g->stmts);
 	g->barrier.data = g;
 	g->barrier.cb = NULL;
@@ -1354,6 +1353,7 @@ handle:
 	req->buffer = buffer;
 	req->db_id = 0;
 	req->stmt_id = 0;
+	req->sql = NULL;
 
 	switch (type) {
 #define DISPATCH(LOWER, UPPER, _)                 \
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -108,7 +108,7 @@ void gateway__close(struct gateway *g)
 	}
 
 #define CHECK_LEADER(REQ)                                            \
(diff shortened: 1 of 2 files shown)


src/gateway.c
```diff
@@ -522,25 +522,19 @@ static void query_barrier_cb(struct barrier *barrier, int status)
...
 	assert(stmt != NULL);
 
-	g->stmt = NULL;
 	g->req = NULL;
 
 	if (status != 0) {
-		if (g->stmt_finalize) {
-			sqlite3_finalize(stmt);
-			g->stmt_finalize = false;
-		}
```
freeekanayaka, line 548: I'm not entirely what this `if` block is about (I **think** it was added not to long ago to handle some corner case, probably git history will help in figuring that out), however why is it not needed anymore?

Is there a test in the pristine master code that fails when removing the above `if`? (proving that the `if` is tested and needed).

freeekanayaka: Thanks for looking. It's unfortunate that the test passes even with the code commented out. I don't quite remember the details here, but I've seen that in follow-up commits you change things even further, so it's probably not something to worry too much about. There might be issues that were covered by this `if` that are not relevant anymore, if they are we'll hopefully trigger them in jepsen or something like that (given that the unit tests don't show it).

src/gateway.h
```diff
@@ -71,6 +71,7 @@ struct handle
 	size_t db_id;            /* For use by prepare callback */
 	size_t stmt_id;          /* For use by prepare callback */
 	const char *sql;
+	unsigned exec_count;
```
freeekanayaka, line 73: I didn't look to hard, but I find the use of `exec_count` a bit obscure. It seems to somehow replace the previous `gateway->stmt`, which was used to save temporarily prepared statements for `QUERY_SQL` and `EXEC_SQL`, but I don't quite understand `exec_count`.

freeekanayaka: Or maybe I'm confused, I'll try to give another look tomorrow. Don't block on this, I guess it's mostly for my understanding, although some detailed comment in declaration of the `unsigned exec_count` field (and `const char *sql` too) would be great.

PR #477, opened 2023-02-15
Support QUERY and QUERY_SQL that modify the database

### LINKED ISSUE #471: Wire protocol limitations as a backend for the C client

cole-miller:
Closes #471. This is based on #476, I'll rebase onto master once that merges.

As suggested by @freeekanayaka, we don't support modifying statements that also return rows (for now). Modifying statements submitted with QUERY or QUERY_SQL go through exactly the same codepath as if submitted with EXEC or EXEC_SQL, respectively, except that we send an empty ROWS response instead of a RESULT response.

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -186,6 +186,18 @@ static void failure(struct handle *req, int code, const char *message)
 	req->cb(req, 0, DQLITE_RESPONSE_FAILURE, 0);
 }
 
+static void emptyRows(struct handle *req)
+{
+	void *cursor = buffer__advance(req->buffer, 8 + 8);
+	uint64_t val;
+	assert(cursor != NULL);
+	val = 0;
+	uint64__encode(&val, &cursor);
+	val = DQLITE_RESPONSE_ROWS_DONE;
+	uint64__encode(&val, &cursor);
+	req->cb(req, 0, DQLITE_RESPONSE_ROWS, 0);
+}
+
 static int handle_leader_legacy(struct gateway *g, struct handle *req)
 {
         tracef("handle leader legacy");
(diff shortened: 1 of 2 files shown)


src/gateway.c
```diff
@@ -576,9 +608,15 @@ static int handle_query(struct gateway *g, struct handle *req)
 	}
 	req->stmt_id = stmt->id;
 	g->req = req;
-	rv = leader__barrier(g->leader, &g->barrier, query_barrier_cb);
+
+	is_readonly = (bool)sqlite3_stmt_readonly(stmt->stmt);
+	if (is_readonly) {
+		rv = leader__barrier(g->leader, &g->barrier, query_barrier_cb);
```
MathieuBordere, line 614: Don't let it stop you (it's there in original code), but where do we send a `failure` response if this call or `leader__exec` in the other branch fails?

MathieuBordere (approved):

PR #478, opened 2023-02-16
server: Expose uv block_size setting.

MathieuBordere:
Exposes the raft setting to set the `uv->block_size`. I started digging a bit because I had horrible results with the `dqlite-benchmark` tool on my laptop. I noticed that +- 99.5% of the async writes failed with `EAGAIN` on my laptop with 4KB block size. Changing the block size to 64KB increase success rate of the async writes to +- 80% and doubles the write throughput (or halves the average time a default write by the benchmark tool takes) on my 2 machines.

I think it's worth exposing this setting so that users can experiment with it. A lot will depend on the kind of workload I think.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -367,6 +367,32 @@ int dqlite_node_set_snapshot_params(dqlite_node *n,
 	return 0;
 }
 
+#define KB(N) (1024 * N)
+int dqlite_node_set_block_size(dqlite_node *n, size_t size)
+{
+	if (n->running) {
+		return DQLITE_MISUSE;
+	}
+
+	switch (size) {
+		case 512:      // fallthrough
+		case KB(1):    // fallthrough
+		case KB(2):    // fallthrough
+		case KB(4):    // fallthrough
+		case KB(8):    // fallthrough
+		case KB(16):   // fallthrough
+		case KB(32):   // fallthrough
+		case KB(64):   // fallthrough
+		case KB(128):  // fallthrough
+		case KB(256):
+			break;
+		default:
+			return DQLITE_ERROR;
+	}
+
+	raft_uv_set_block_size(&n->raft_io, size);
+	return 0;
+}
 int dqlite_node_enable_disk_mode(dqlite_node *n)
 {
 	int rv;
--- a/include/dqlite.h
+++ b/include/dqlite.h
@@ -200,6 +200,14 @@ int dqlite_node_set_snapshot_params(dqlite_node *n,
 				    unsigned snapshot_trailing);
 
 /**
+ * Set the block size used for performing disk IO when writing raft log segments
+ * to disk. @size is limited to a list of preset values.
+ *
+ * This function must be called before calling dqlite_node_start().
+ */
+int dqlite_node_set_block_size(dqlite_node *n, size_t size);
+
+/**
  * WARNING: This is an experimental API.
  *
  * By default dqlite holds the SQLite database file and WAL in memory. By
--- a/test/integration/test_node.c
+++ b/test/integration/test_node.c
@@ -316,6 +316,43 @@ TEST(node, networkLatencyMsTooLarge, setUp, tearDown, 0, node_params)
 	return MUNIT_OK;
 }
 
+TEST(node, blockSize, setUp, tearDown, 0, NULL)
+{
+	struct fixture *f = data;
+	int rv;
+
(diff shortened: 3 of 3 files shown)


MathieuBordere:
@freeekanayaka Can you share some insight regarding the `uv->block_size` parameter in libraft and how it was chosen? Is the 4KB maximum size chosen for some (extra) data durability guarantees or ... ? Any insight is welcome.

freeekanayaka:
> @freeekanayaka Can you share some insight regarding the `uv->block_size` parameter in libraft and how it was chosen? Is the 4KB maximum size https://github.com/canonical/raft/blob/5243fa2cb568456f58dd6e5852fc11d95cd08b72/src/uv_fs.c#L667 chosen for some (extra) data durability guarantees or ... ? Any insight is welcome.

I think 4KB was the best practical choice at that time, since it was what the kernel recommended, but I might misremember or things might have changed in the meantime, also hardware-wise.

It should be ok to modify it.

freeekanayaka:
Instead of exposing this setting perhaps fixing the auto-detection would be better? Of course both things can be done if you think it's important to have a way to bypass the auto-detection. However, if the auto-detection is reasonably good, perhaps it's ok not expose it, one less knob. If it turns out that there are case where the auto-detection is faulty, it can be fixed (like in this case). Just thinking loud, I'm fine either way,

MathieuBordere:
Yes, I think there's certainly room to improve the auto detection and was planning to eventually do that, this is more of a first step to have an easy way to turn a knob and understand the influence of the parameter.

src/server.c
```diff
@@ -356,6 +356,32 @@ int dqlite_node_set_snapshot_params(dqlite_node *n, unsigned snapshot_threshold,
...
+	case 512:       // fallthrough
+	case KB(1):     // fallthrough
+	case KB(2):     // fallthrough
+	case KB(4):     // fallthrough
+	case KB(8):     // fallthrough
+	case KB(16):    // fallthrough
+	case KB(32):    // fallthrough
+	case KB(64):    // fallthrough
+	case KB(128):   // fallthrough
+	case KB(256):
```
freeekanayaka, line 376: It feels like these preset sizes check should be placed in `raft_uv_set_block_size()` instead, because it really applies there (any consumer of `raft_uv_set_block_size()` will want to have the same check).

MathieuBordere:
I'll still move it to libraft.

calvin2021y:
> Exposes the raft setting to set the `uv->block_size`. I started digging a bit because I had horrible results with the `dqlite-benchmark` tool on my laptop. I noticed that +- 99.5% of the async writes failed with `EAGAIN` on my laptop with 4KB block size. Changing the block size to 64KB increase success rate of the async writes to +- 80% and doubles the write throughput (or halves the average time a default write by the benchmark tool takes) on my 2 machines.
> 
> I think it's worth exposing this setting so that users can experiment with it. A lot will depend on the kind of workload I think.

maybe you can try reform your disk LBA Forma.

freeekanayaka:
FWIW libuv's has now support for `io_uring` for file operations, so it should be relatively easy now to move away from the complicated `aio`-based code that we use now for async file writes. Or alternatively go for native `io_uring`. AFAIK `io_uring` takes care of details like optimal block size.

calvin2021y:
@freeekanayaka 

libuv + io uring is ready : https://github.com/libuv/libuv/commit/96e05543f53b19d9642b4b0dd73b86ad3cea313e

MathieuBordere:
> FWIW libuv's has now support for `io_uring` for file operations, so it should be relatively easy now to move away from the complicated `aio`-based code that we use now for async file writes. Or alternatively go for native `io_uring`. AFAIK `io_uring` takes care of details like optimal block size.

I also would support to depend on a specialized system to detect optimal block size instead ofdoing it ourselves suboptimally.

freeekanayaka:
> > FWIW libuv's has now support for `io_uring` for file operations, so it should be relatively easy now to move away from the complicated `aio`-based code that we use now for async file writes. Or alternatively go for native `io_uring`. AFAIK `io_uring` takes care of details like optimal block size.
> 
> I also would support to depend on a specialized system to detect optimal block size instead ofdoing it ourselves suboptimally.

Is there one?

Switching to `io_uring` via `libuv` might be an easier and longer-term fix (in the sense that the `aio` kernel subsystem is pretty much deprecated right now).

PR #480, opened 2023-02-28
Implement server-side role management

cole-miller:
This is still rough and most error handling is stubbed out, but I think it represents a workable approach to moving role management onto the server.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -87,6 +87,7 @@ unit_test_SOURCES += \
   test/unit/test_registry.c \
   test/unit/test_replication.c \
   test/unit/test_request.c \
+  test/unit/test_role_management.c \
   test/unit/test_tuple.c \
   test/unit/test_vfs.c \
   test/unit/main.c
@@ -100,6 +101,7 @@ integration_test_SOURCES = \
   test/integration/test_fsm.c \
   test/integration/test_membership.c \
   test/integration/test_node.c \
+  test/integration/test_role_management.c \
   test/integration/test_vfs.c \
   test/integration/main.c
 integration_test_CFLAGS = $(AM_CFLAGS) -Wno-conversion
--- a/test/integration/test_role_management.c
+++ b/test/integration/test_role_management.c
@@ -1,0 +1,160 @@
+#include "../../src/client/protocol.h"
+#include "../../src/server.h"
+#include "../lib/client.h"
+#include "../lib/endpoint.h"
+#include "../lib/fs.h"
+#include "../lib/heap.h"
+#include "../lib/runner.h"
+#include "../lib/server.h"
+#include "../lib/sqlite.h"
+#include "../lib/util.h"
+
+#define N_SERVERS 5
+#define FIXTURE                                \
+	struct test_server servers[N_SERVERS]; \
+	struct client_proto *client;           \
+	struct rows rows;
+
+#define SETUP                                                 \
+	unsigned i_;                                          \
+	test_heap_setup(params, user_data);                   \
+	test_sqlite_setup(params);                            \
+	for (i_ = 0; i_ < N_SERVERS; i_++) {                  \
+		struct test_server *server = &f->servers[i_]; \
+		test_server_setup(server, i_ + 1, params);    \
+	}                                                     \
+	test_server_network(f->servers, N_SERVERS);           \
+	for (i_ = 0; i_ < N_SERVERS; i_++) {                  \
+		struct test_server *server = &f->servers[i_]; \
+		test_server_start(server, params);            \
+	}                                                     \
+	SELECT(1)
+
+#define TEAR_DOWN                                        \
+	unsigned i_;                                     \
+	for (i_ = 0; i_ < N_SERVERS; i_++) {             \
+		tracef("test_server_tear_down(%u)", i_); \
+		test_server_tear_down(&f->servers[i_]);  \
+	}                                                \
+	test_sqlite_tear_down();                         \
+	test_heap_tear_down(data)
+
+#define SELECT(ID) f->client = test_server_client(&f->servers[ID - 1])
+
+#define TRIES 5
+
+static char *trueonly[] = {"1", NULL};
+
+static char *threeonly[] = {"3", NULL};
+
+static MunitParameterEnum role_management_params[] = {
+    {"role_management", trueonly},
+    {"target_voters", threeonly},
+    {"target_standbys", threeonly},
+    {NULL, NULL},
+};
+
+SUITE(role_management)
+
+struct fixture
+{
+	FIXTURE;
+};
+
+static void *setUp(const MunitParameter params[], void *user_data)
+{
+	struct fixture *f = munit_malloc(sizeof *f);
+	SETUP;
+	return f;
+}
+
+static void tearDown(void *data)
+{
+	struct fixture *f = data;
+	TEAR_DOWN;
+	free(f);
+}
+
+static bool hasRole(struct fixture *f, dqlite_node_id id, int role)
+{
+	struct client_node_info *servers;
+	size_t n_servers;
+	struct client_context context;
+	unsigned i;
+	bool ret = false;
+	int rv;
+
+	clientContextMillis(&context, 5000);
+	rv = clientSendCluster(f->client, &context);
+	munit_assert_int(rv, ==, 0);
+	rv = clientRecvServers(f->client, &servers, &n_servers, &context);
+	munit_assert_int(rv, ==, 0);
+	for (i = 0; i < n_servers; i += 1) {
+		if (servers[i].id == id) {
+			ret = servers[i].role == role;
+			break;
+		}
+	}
+	for (i = 0; i < n_servers; i += 1) {
+		free(servers[i].addr);
+	}
+	free(servers);
+
+	return ret;
+}
+
+TEST(role_management, promote, setUp, tearDown, 0, role_management_params)
+{
+	struct fixture *f = data;
+	unsigned id = 2;
+	const char *address = "@2";
+	int tries;
+
+	HANDSHAKE;
+
+	id = 2;
+	address = "@2";
+	ADD(id, address);
+	for (tries = 0; tries < TRIES && !hasRole(f, 2, DQLITE_VOTER);
+	     tries += 1) {
+		sleep(1);
+	}
+	if (tries == TRIES) {
+		return MUNIT_FAIL;
+	};
+
+	id = 3;
+	address = "@3";
+	ADD(id, address);
+	for (tries = 0; tries < TRIES && !hasRole(f, 3, DQLITE_VOTER);
+	     tries += 1) {
+		sleep(1);
+	}
+	if (tries == TRIES) {
+		return MUNIT_FAIL;
+	};
+
+	id = 4;
+	address = "@4";
+	ADD(id, address);
+	for (tries = 0; tries < TRIES && !hasRole(f, 4, DQLITE_STANDBY);
+	     tries += 1) {
+		sleep(1);
+	}
+	if (tries == TRIES) {
+		return MUNIT_FAIL;
+	};
+
+	id = 5;
+	address = "@5";
+	ADD(id, address);
+	for (tries = 0; tries < TRIES && !hasRole(f, 5, DQLITE_STANDBY);
+	     tries += 1) {
+		sleep(1);
+	}
+	if (tries == TRIES) {
+		return MUNIT_FAIL;
+	};
+
+	return MUNIT_OK;
+}
--- a/test/lib/server.c
+++ b/test/lib/server.c
@@ -39,8 +39,15 @@ void test_server_stop(struct test_server *s)
 	int rv;
 
 	test_server_client_close(s, &s->client);
-	rv = dqlite_node_stop(s->dqlite);
+
+	if (s->role_management) {
+		dqlite_node_handover(s->dqlite);
+		rv = dqlite_node_stop(s->dqlite);
+	} else {
+		rv = dqlite_node_stop(s->dqlite);
+	}
 	munit_assert_int(rv, ==, 0);
+
 	dqlite_node_destroy(s->dqlite);
 }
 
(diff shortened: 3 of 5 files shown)


MathieuBordere (commented):
Looks good on first pass

src/role_management.c
```diff
@@ -0,0 +1,665 @@
...
+		polling = &polling_objs[j];
+		polling->cb = cb;
+		polling->node = d;
+		polling->cluster = cluster;
+		polling->n_cluster = n;
+		polling->count = count;
+		polling->i = j;
+		work = &work_objs[j];
+		work->data = polling;
+		rv = uv_queue_work(&d->loop, work, pollClusterWorkCb, pollClusterAfterWorkCb);
```
MathieuBordere, line 471: I think we might want to increase the libuv thread pool size, see https://docs.libuv.org/en/v1.x/threadpool.html something like 128 by default?

freeekanayaka:
> I'm wondering whether the assertion failure inside uv_close (resulting from an uninitialized handle) is a problem that only surfaces with specific versions of libuv. When I run the new integration test locally against libuv v1.44.3, it's not triggered.

It's our code that is calling `uv_close()`, and that assertion seems to imply that the handle that we're passing to `uv_close()` is somehow screwed up (either uninitialized, as you point, or maybe with corrupted memory that was overwritten somewhere), so even if for some reason the problem doesn't surface in v1.44.3 it looks like a bug on our side.

src/role_management.c
```diff
@@ -0,0 +1,685 @@
...
+{
+	queue *root = &d->changes;
+	queue *head = QUEUE__NEXT(root);
+	struct change_record *rec;
+	QUEUE__INIT(&d->changes);
+	while (head != root) {
+		rec = QUEUE__DATA(head, struct change_record, queue);
+		head = QUEUE__HEAD(&rec->queue);
+		raft_free(rec);
+	}
```
freeekanayaka, line 684: If I understand the goal of the code correctly, a more idiomatic (and probably more readable) way to do this would be:

```C
	while (!QUEUE__IS_EMPTY(&d->changes)) {
		struct change_record *rec;
		queue *head;
		head = QUEUE__HEAD(&d->changes);
		rec = QUEUE__DATA(head, struct change_record, queue);
		QUEUE__REMOVE(head);
		raft_free(rec);
	}
```

There are a few examples of this in the dqlite and raft code base.

src/role_management.c
```diff
@@ -0,0 +1,706 @@
...
+}
+
+/* When a role change has completed, start the next one. */
+static void changeCb(struct raft_change *change, int status)
+{
+	struct dqlite_node *d = change->data;
+
+	raft_free(change);
+	if (status != 0) {
+		/* TODO */
```
freeekanayaka, line 218: Could it be that when shutting down `status` is `RAFT_CANCELED` or other similar error but nevertheless the code starts a new change just below? That might somehow explain the `uv_close()` assertion failure.

include/dqlite.h
```diff
@@ -218,6 +247,8 @@ int dqlite_node_enable_disk_mode(dqlite_node *n);
  */
 int dqlite_node_start(dqlite_node *n);
 
+int dqlite_node_handover(dqlite_node *n);
```
freeekanayaka, line 250: Missing docstring.

The current implementation seems to not only handover the role, but also to stop the node, effectively makimg `dqlite_node_handover()` an alternative to `dqlite_node_stop()`. I feel this API should just handover the role, without stopping the node and leaving that responsibility to `dqlite_node_stop()`.

One reason is to give users a chance to first try to run `dqlite_node_handover()` and in case it returns non-zero possibly take appropriate measures **before** running `dqlite_node_stop()` and stop the node for good.

Another reason is that it somehow makes modular code harder: imagine that in the future we want to add some other optional behavior that happens before shutdown. We could add a `dqlite_node_some_new_pre_shutdown_behavior_that_also_stops_the_node()` function, which like `dqlite_node_handover` implements some specific pre-shutdown behavior and then stops the node. But what if we want to give users full control about choosing what pre-shutdown behavior to run? For instance they might want to run the new behavior **and** the handover behavior, or just one of them. The user might also have custom logic that they want to run between the handover and the actual stop.

If we separate concerns of optional pre-shutdown behavior and actual shutdown, we can have more control by mix and matching them, e.g.:

```C
int rv;

/* First try to handover our role */
rv = dqlite_node_handover(n);
if (rv != 0) {
    /* Run some custom logic that my application needs 
     * in case the handover fails. */
}

/* Also run the other new dqlite pre-shutdown behavior, in
 * this case we don't care about whether it succeeds or not */
dqlite_node_some_new_pre_shutdown_behavior(n);

/* Run app-specific logic that is required to run just before
* the final stop. */
my_app_pre_shutdown_behavior();

/* Finally stop the node for real */
dqlite_node_stop(n);
```

We should also make the timeout for `dqlite_node_handover()` configurable (it's kind if hard-coded right now): basically the function would return an error if it could not successfully handover the role within the specified timeout. We could have a reasonable default timeout, and let the users change it with something like `dqlite_node_set_handover_timeout(n, msecs)`. This could be done in a separate PR too.

freeekanayaka: > Oh, I guess this isn't insurmountable, since the thread where dqlite_node_handover is called is not the same as the libuv main thread, so we can block in the "handover thread" on some signal from the libuv thread that the handover has wrapped up. Using pthread_condvar_t I suppose, or a semaphore (which is what dqlite_node_start does).

Exactly, the same approach implemented by `dqlite_node_start()` can be used here.

src/role_management.c
freeekanayaka: Note that if the capacity of the thread pool gets exhausted, then further calls to `uv_queue_work` will not fail, but rather queued and executed when threads become available. In general the size of the threadpool should ideally match the hardware capacity (e.g. the number of cores of the host), in order to reduce contention and context switches.

I think we can safely leave libuv's default.

src/role_management.c
```diff
@@ -35,6 +35,29 @@
  * thread pool (see pollClusterAfterWorkCb). We don't start a new round of
  * adjustment if a "tick" occurs while the queue of changes from the last
  * round is still nonempty.
+ *
+ * A handover is triggered when we call dqlite_node_stop on a node that's
```
freeekanayaka, line 39: s/dqlite_node_stop/dqlite_node_handover/

include/dqlite.h
freeekanayaka: Looks great, thanks!

src/server.c
```diff
@@ -447,11 +459,40 @@ static void destroy_conn(struct conn *conn)
...
+	struct dqlite_node *d = handover->data;
+	int rv;
+
+	/* Nothing to do. */
+	if (d->status != DQLITE_SERVER_RUNNING) {
+		return;
+	}
+
+	/* TODO how should the server respond to client requests while in this
+	 * state? */
```
freeekanayaka, line 480: During a `dqlite_node_handover()` request there's nothing special we need to do. If this node was not the leader, it is not serving client requests. If it was the leader, it will try to transfer leadership, and after that's done client requests will start failing and the client will figure out who the new leader is. Note that both clients and the `gateway.c` code that servers requests are prepared for losing leadership at any time anyways, regardless of `dqlite_node_handover()`, so `dqlite_node_handover()` doesn't quite change anything for them.

freeekanayaka:
> Broken up into commits for easier review :)

Much easier to review indeed!

src/role_management.c
```diff
@@ -0,0 +1,722 @@
...
+		return;
+	}
+	/* If a series of role adjustments is already in progress, don't kick
+	 * off another one. */
+	if (!QUEUE__IS_EMPTY(&d->changes)) {
+		return;
+	}
+	/* If a handover is in progress or the node has shut down, do nothing.
+	 */
+	if (d->status != DQLITE_SERVER_RUNNING) {
```
freeekanayaka, line 675: I think this can be turned into an assertion: `assert(d->status == DQLITE_SERVER_RUNNING`, or rather `assert(d->running)` considering the other comment. There should be no way that `RoleManagementAdjust()` can be called while the server is not running.

This will also eliminate the "line not covered by tests" warning emitted by Codecov below.

src/role_management.h
```diff
@@ -0,0 +1,49 @@
...
+ * be transferred to other nodes if necessary, and then the callback will be
+ * invoked on the loop thread. The callback's second argument will be 0 if the
+ * handover succeeded and nonzero otherwise. */
+void RoleManagementHandover(struct dqlite_node *d,
+			    void (*cb)(struct dqlite_node *, int));
+
+/* Drain the queue of changes computed by RoleManagementAdjust. This should be
+ * done when the node is shutting down, to avoid a memory leak. */
+void RoleManagementDrain(struct dqlite_node *d);
+
```
freeekanayaka, line 48: Nit: what about renaming:

- `roles_management.{c,h}` to  `roles.{c.h}`
- `RoleManagementAdjust()` to `RolesAdjust()`
- `RoleManagementHandover()` to `RolesHandover()`
- `RolesManagementDrain()` to `RolesCancelPendingChanges()`
- `RoleManagementComputeChanges` to `RolesComputeChanges()`
- `struct dqlite_node->changes` to `struct dqlite_node->roles_changes`

It's a bit more terse, and the "management" word is a bit redundant. We have many other things that are "management" (actually mostly everything :), but we don't tag them as such for sake of brevity.

freeekanayaka: Right, what I meant is that it should be fairly obvious that `roles.c` is where role management code lives and that `RolesXXX` APIs are the ones dealing with role management.

src/roles.c
```diff
@@ -0,0 +1,525 @@
...
+	for (i = 0; i < n_cluster && standby_count > standbys; i += 1) {
+		if (cluster[i].role != DQLITE_STANDBY) {
+			continue;
+		}
+		cb(cluster[i].id, DQLITE_SPARE, arg);
+		cluster[i].role = DQLITE_SPARE;
+		standby_count -= 1;
+		removeDomain(cluster[i].failure_domain, &standby_compare);
+	}
+}
```
freeekanayaka, line 349: I don't have time right now to go through this in detail, but I assume it's basically a verbatim translation of the equivalent code in `go-dqlite`? In app.RoleChanges.Adjust().

freeekanayaka (approved):
Looks good overall. I left some other comments.

It'd be nice to try to enable role management in the dqlite server, disable it in the dqlite `app` code and then try to run the `test/roles.sh` suite and see how it behaves.

It'd be also nice to add to the dqlite suite itself a test similar to `go-dqlite`'s `test/roles.sh`, either as shell script or C test. That is going to require some work though and could be done separately.

src/roles.c
```diff
@@ -0,0 +1,525 @@
...
+ *   joined nodes? go-dqlite does this, but I'm not convinced it's important,
+ *   or that it should run on the server if we do decide we want it.
+ */
+
+/* XXX */
+#define NUM_TRACKED_DOMAINS 5
+
+struct change_record
+{
+	uint64_t id;
```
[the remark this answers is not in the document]
freeekanayaka: I think we had this conversation before, the principle is that model/language types are independent from protocol/serialization concerns.

At some point we might change the serialization type of a given model/language type without changing the model/language type (thus without breaking the API/ABI). For example we might change the protocol to allow for greater values, without requiring a language/model change.

One recent example of that was the allowed number of statement parameters. The protocol version X only allowed for 256 parameters. But the associated language type was **not** `uint8_t` (or whatever Go equivalent). By separating the two concerns we were able to bump the protocol version to version X+1 with a different serialization format/type for the number of the statement parameters without needing to change the language. In the implementation there was a missing check for the 256 limit, but that's fine, it was just a bug, and the fix was not to change the language type from `int` to `uint8_t` (or whatever exact Go types), but rather to add the check. 

In this specific case the current situation is that the failure domain must be a value between `0` and `2^64`, but we don't want to enforce that with a `uint64_t` type, we want to enforce that with a check in the code (and add documentation if missing). This way we can evolve the wire protocol in the future, if required. We're most probably not going to need that for the failure domain values specifically, but this the general idea.

> Is our code actually prepared to deal with a hypothetical target where `unsigned long long` and `uint64_t` are not the same type?

That I didn't check. If it's not, it should be fixed with a check on the value, and documented.

Note that boundary checks are just one type of model value checks. In general any expectation we have on a value (e.g. the value should be "odd", or it should be a multiple of 3 or whatever) should be explicitly coded.

> It seems to me that if the IDs, failure domains, etc. are required by the wire protocol to be 64 bits, we should just be using `uint64_t` or a typedef of it everywhere.

As said, the wire protocol is opaque, and can change. What should be more stable is the language/model API.

> Using the "wobbly-width" C integer types in that kind of situation is (IMO) either pointless, if the width of the type is effectively fixed (as for `unsigned long long`), or buggy, if the width can vary across targets.

This is indeed the point, the width of the type is **not** effectively fixed **by design**. The current boundary values are [0..2^64] (and as said they should be documented and checked), but those could change in the future without changing the **type**.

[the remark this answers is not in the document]
freeekanayaka: > (I don't find the argument from optimization persuasive because I can't think of a target where `unsigned long long` is 128 bits, and even on such a target the gains from using a 128-bit type over a 64-bit type seem to me very unlikely to justify the complexity incurred by mixing `unsigned long long` and `uint64_t` when these are not the same width.)

The optimization argument is ancillary and definitely not the main, the point is more to separate concerns conceptually. As said, failure domains don't **need** to be 64 bits, and should not be modeled as such. As said, this doesn't matter much in this specific case, but it might matter in other cases, and for consistency we apply the same approach everywhere.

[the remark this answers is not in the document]
freeekanayaka: In a 32-bit world, it would be good practice to use `int` for your APIs, not `int32_t`. At that point you might have checks in place to ensure that your values were at most 2^32 (because for example you were serializing them using `int32_t`). Then you start supporting new 64-bit architectures, and you don't need to change the API (`int` to `int64_t`), but you might change the serialization and relax your boundary checks to allow for greater values. User code does not need to change.

src/server.c
```diff
@@ -114,7 +110,6 @@ int dqlite__init(struct dqlite_node *d,
 		rv = DQLITE_ERROR;
 		goto err_after_ready_init;
 	}
-#endif
```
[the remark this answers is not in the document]
rabits: Oh, yeah, forgot about this PR. So raft become a big problem for me at a time, and since I don't use dqlite anymore - I can't continue to work on the macos support.

PR #496, opened 2023-04-13
client/protocol.c: Various improvements

cole-miller:
This is in preparation for the C client. One or two bug fixes (like making sure to return early in RESPONSE), some changes to integer types, refactoring the "strdup or OOM" pattern into a function, and adding support for custom connect functions.

### DIFF
--- a/src/client/protocol.h
+++ b/src/client/protocol.h
@@ -33,10 +33,13 @@ enum {
 
 struct client_proto
 {
+	int (*connect)(void *, const char *, int *);
+	void *connect_arg;
 	int fd;		     /* Connected socket */
 	uint32_t db_id;      /* Database ID provided by the server */
 	char *db_name;       /* Database filename (owned) */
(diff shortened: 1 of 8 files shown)


src/client/protocol.h
```diff
@@ -90,8 +93,16 @@ struct client_file
  * in milliseconds. */
 void clientContextMillis(struct client_context *context, long millis);
 
+void *mallocChecked(size_t);
+void *callocChecked(size_t, size_t);
+char *strdupChecked(const char *);
+char *strndupChecked(const char *, size_t);
+
+void makeValueOwned(struct value *);
+void freeOwnedValue(struct value);
```
freeekanayaka, line 102: At the moment these functions don't seem to be used outside of `protocol.c`. Is there a plan to use them elsewhere too?

MathieuBordere:
If you could sort out the CI failures (you're probably on it).

PR #501, opened 2023-04-20
C client implementation

cole-miller:
I'm sending this up as one PR, adding commits incrementally. The first commit is just the new header file `dqlite/client.h`, so that any tweaks to that interface that affect the implementation can get ironed out sooner rather than later (I do have a lot of the implementation written locally).

### DIFF
--- a/src/client/easy.c
+++ b/src/client/easy.c
@@ -716,6 +716,71 @@ err:
 	return 1;
 }
 
+struct dqlite
+{
+	struct dqlite_server *server;
+	struct client_proto proto;
+	uint32_t db_id;
+};
+
+int dqlite_open(dqlite_server *server, const char *name, dqlite **db)
+{
+	struct client_context context;
+	struct client_proto new_proto = {0};
+	struct client_proto old_proto;
+	const struct client_node_info *info;
+	struct dqlite *result;
+	int rv;
+
+	rv = pthread_mutex_lock(&server->mutex);
+	assert(rv == 0);
+	info = findNodeInCache(&server->cache, server->proto.server_id);
+	if (info == NULL) {
+		/* TODO */
+		abort();
+	}
+	rv = clientOpen(&new_proto, info->addr, info->id);
+	if (rv != 0) {
+		return 1;
+	}
+	old_proto = server->proto;
+	server->proto = new_proto;
+	rv = pthread_mutex_unlock(&server->mutex);
+	assert(rv == 0);
+
+	clientContextMillis(&context, 5000);
+	rv = clientSendOpen(&old_proto, name, &context);
+	if (rv != 0) {
+		goto err_after_steal_client;
+	}
+	rv = clientRecvDb(&old_proto, &context);
+	if (rv == DQLITE_CLIENT_PROTO_RECEIVED_FAILURE &&
+	    old_proto.errcode == SQLITE_IOERR_NOT_LEADER) {
+		rv = reconnectToLeader(&old_proto, &context);
+		if (rv != 0) {
+			goto err_after_steal_client;
+		}
+		rv = clientSendOpen(&old_proto, name, &context);
+		if (rv != 0) {
+			goto err_after_steal_client;
+		}
+		rv = clientRecvDb(&old_proto, &context);
+	}
+	if (rv != 0) {
+		goto err_after_steal_client;
+	}
+
+	result = callocChecked(1, sizeof *result);
+	result->server = server;
+	result->proto = old_proto;
+	*db = result;
+	return 0;
+
+err_after_steal_client:
+	clientClose(&old_proto);
+	return 1;
+}
+
 int dqlite_server_handover(dqlite_server *server)
 {
 	int rv = dqlite_node_handover(server->local);


include/dqlite/client.h
```diff
@@ -0,0 +1,361 @@
...
+ * Configure the target number of voting nodes for this cluster.
+ *
+ * This must be an odd number.
+ */
+int dqlite_cluster_set_voters(dqlite_cluster *cluster, unsigned n);
+
+/**
+ * Configure the target number of standby nodes for this cluster.
+ */
+int dqlite_cluster_set_standbys(dqlite_cluster *cluster, unsigned n);
```
freeekanayaka, line 114: What about all other knobs/functionality part of `dqlite_node`? `dqlite_node_enable_disk_mode()`, `dqlite_node_set_snapshot_params()`, `dqlite_node_set_network_latency()`, `dqlite_node_set_failure_domain()`, `dqlite_node_recover()` etc.

Will `dqlite_cluster` need to play catch-up with `dqlite_node` regarding knobs and functionality? (e.g. all the times we add a `dqlite_node_xxx` function we also need to add the equivalent `dqlite_cluster_xxx` function in order for dqlite C client users to "see" it).

At first sight it feels that `dqlite_cluster` is augmenting the API surface of `dqlite_node` instead of complementing it. That might be fine, not sure, but I first wanted to ask if it's intentional.

freeekanayaka:
The core client part (the functions equivalent to sqlite3) looks good to me. I'd like to understand the cluster part better.

freeekanayaka:
I guess in practice we'll decide what parts of `dqlite_node` makes sense to expose also to `dqlite_cluster` and what don't, like we do for the Go `app` helper.

I don't see many ways to avoid this, since `dqlite_cluster` is basically a higher-level construct around the lower level `dqlite_node`.

include/dqlite/client.h
```diff
@@ -0,0 +1,361 @@
...
+int dqlite_cluster_start(dqlite_cluster *cluster);
+
+/**
+ * Open a database connection on the given cluster.
+ *
+ * This is the analogue of sqlite3_open. @name is the name of the database
+ * to open, which will be created if it does not exist. All nodes in the cluster
+ * share a "namespace" for databases.
+ */
+int dqlite_open(dqlite_cluster *cluster, const char *name, dqlite **db);
```
freeekanayaka, line 129: Maybe we could rename this one to `dqlite_cluster_client()`?

The reason is that we might want to leave the `dqlite_open` name for a future lower-level API, an API that does not need a `dqlite_cluster` (and `dqlite_cluster_client()` would be basically a helper around `dqlite_open()`, which also knows about current cluster addresses).

Users should be able to use `dqlite_open()` to create client a "pure" dqlite client, without any underlying `dqlite_node` (think for example implementing a dqlite shell in C). The parameters you would pass to `dqlite_open()` could be something like:

```C
dqlite_open(int node_store_type, void* node_store, const char *name, dqlite **db);
```

Where `node_store_type`/`node_store` would be some generic way to pass cluster addresses. Or any other equivalent naming/API.

`dqlite_cluster_client()` could then internally be implemented using `dqlite_open()`.

freeekanayaka:
After some more thought, I believe I now realize where the the friction between `dqlite_cluster` and `dqlite_node` comes from (the fact that `dqlite_cluster` would need to play "catch-up" with `dqlite_node` features).

The current APIs both in `dqlite` and in `go-dqlite` grew somehow organically after we figured out how dqlite should be consumed. For example, the `app.App` API of `go-dqlite` was born after we integrated dqlite into LXD and gathered more experience about how projects would consume dqlite and what their needs are.

Arguably the `app.App` Go package from `go-dqlite` doesn't have anything which is strictly Go-specific, and all the functionality/logic it implements is actually required by every project consuming `dqlite`, independently from the language (Go, C, Python, whatever). That was probably sort of clear even back then when `app.App` was implemented, but it was still written in Go as there wasn't much time available for develpment (`app.App` was just a by-product of the implementation of the Kubernetes kine/dqlite backend, which had a tight delivery deadline), and it was also a good way to prototype the core abstractions of the API . We're now starting to fix that: role management is being moved to the C side (and later down the road it can probably be dropped from `app.App`) and `dqlite_cluster` will basically move the rest of `app.App`, implementing equivalent logic in C, and making it easy to reuse it in any language via C bindings. After everything will be implemented in C, `app.App` will probably won't have much reasons to exist (at least in principle, we'll keep it around for backward compatibility), and new Go bindings could be created that leverage `dqlite_cluster`. So far so good, and I believe we're all on the same page up to here.

I would go one step forward, and point out that `dqlite_cluster` is implementing the API that `dqlite_node` should ideally have had from the very beginning (I'm talking about the functionality, the naming of structs and functions is just a detail). In other words, if we were writing dqlite from scratch now, we would expose the API that `dqlite_cluster` offers, not the current one of `dqlite_node`. After `dqlite_cluster` is in place, there won't be much reason for any dqlite consumer to use `dqlite_node`, in the same way that there is now little reason for Go consumers to use `dqlite.Node` instead of `dqlite.app.App`.

It'd be nice to take this chance to fix our core abstractions (namely `dqlite_node`).

What I propose is:

- Move all `dqlite_cluster` APIs from `dqlite/client.h` to `dqlite.h`, since they are not quite the client, as explained above they are rather what the server API (`dqlite_node`) should have looked like since the beginning.
- Rename `dqlite_cluster` to something like `dqlite_server`. Note that as explained `dqlite_cluster` is really a more mature version of `dqlite_node`, it's not actually controlling the cluster, but rather the node: `dqlite_cluster_create` does not create a dqlite cluster, it creates a single dqlite node, `dqlite_cluster_start` does not start a dqlite cluster, it starts a dqlite node, `dqlite_cluster_stop` does not stop a dqlite cluster, it stops a dqlite node, `dqlite_cluster_set_server_bind_address` can set only the address of the local node, not any node. The `dqlite_cluster` name can be saved for later use, for example if we implement APIs that truly act on the whole cluster (for example we could add a `dqlite_cluster_set_voters` function that actually sets the number of voters at the cluster level, for all nodes, without requiring consumers to start nodes with the same setting for voters and bookkeep that configuration externally, possibly with the risk of it going out of sync across nodes and having nodes with different configurations for the desired voters count)
- Dump dqlite version to `2.0` and document that `dqlite_server` is the new official API for creating and controlling a dqlite node.
- Add a deprecation note to all `dqlite_node` docstrings
- Explain that version `2.0` is fully backward-compatible with `1.0`, but new code should use the new APIs. At some point in a reasonably distant future we might want to drop `dqlite_node` entirely, or alternatively keep it it indefinitely if preferred.

Below is a reference sketch, basically just a rename of `dqlite_cluster` to `dqlite_server`.

It has also a small adjustment, breaking up `dqlite_cluster_declare_node` into `dqlite_server_set_address` and `dqlite_server_set_peers`. I'm not quite sure here of the best API of this, in terms of naming and semantic, just wanted to stress the fact that this is a local config not a cluster one, this is a bit of a detail and suggestions are welcome (perhaps we can keep the API same as `dqlite_cluster_declare_node` and just change the name, but I could not come up with a good one).

```C
typedef struct dqlite_server dqlite_server;
typedef int (*dqlite_connect_func)(void *, const char *, int *);

/**
 * Create a dqlite server that will be part of a dqlite cluster and begin
 * configuring it.
 *
 * @path is the path to a directory where the dqlite server will store
 * persistent state. The directory must exist before calling this function.
 * On success, a pointer to the created object is stored in @server.
 *
 * No reference to the string pointed to by @path is kept after this
 * function returns.
 */
int dqlite_server_create(const char *path, dqlite_server **server);

/**
 * Set the network address of the given dqlite server.
 *
 * Other dqlite servers must be able to connect to this server using the
 * given address.
 *
 * This function should be called when starting a server for the very first
 * time. After that first start, calling this function is optional, since the
 * address will be stored locally and will be automatically used during
 * subsequent restarts. If called again, this function should be passed the
 * same value as the first time.
 * 
 * @address is the (abstract) address of the server.
 *
 * No reference to the string pointed to by @address is kept after this function
 * returns.
 */
int dqlite_server_set_address(dqlite_server *server,
				const char *address);

/**
 * Inform this server about existing servers already part of the cluster.
 * 
 * This function should be called when starting for the very first time a
 * server that is meant to join an existing cluster of dqlite servers .
 * 
 * If the server is meant to be the first server of a new cluster (i.e. it's
 * the "bootstrap" server), then this function should not be called or
 * it should be passed an empty list of peers.
 *
 * After that first start, calling this function is optional, and it will be a
 * no-op (information about the cluster members is stored locally and
 * refreshed at regular intervals).
 *
 * @peers is an array of (abstract) addresses of existing servers of the
 * cluster that this server should join. The list does not need to be
 * exhaustive and contain all the existing servers of the cluster, even just
 * one existing server is enough, as long as it's currently running.
 * @n_peers is the length of the @peers array.
 *
 * No reference to the strings contained in @peers or to @peers itself is
 * kept after this function returns.
 */
int dqlite_server_set_peers(dqlite_server *server,
				const char **peers,
				unsigned n_peers);

/**
 * Set the address on which the server will listen for requests
 * from clients and messages from other servers.
 *
 * The format of the @address argument must be one of
 *
 * 1. "<HOST>"
 * 2. "<HOST>:<PORT>"
 * 3. "@<PATH>"
 *
 * Where <HOST> is a numeric IPv4/IPv6 address, <PORT> is a port number, and
 * <PATH> is an abstract Unix socket path. The port number defaults to 8080 if
 * not specified. In the second form, if <HOST> is an IPv6 address, it must be
 * enclosed in square brackets "[]". In the third form, if <PATH> is empty, the
 * implementation will automatically select an available abstract Unix socket
 * path, which can then be retrieved with dqlite_server_get_bind_address().
 *
 * If an abstract Unix socket is used the dqlite server will accept only
 * connections originating from the same process.
 *
 * No reference to the memory pointed to by @address is kept, so any memory
 * associated with them can be released after the function returns.
 */
int dqlite_server_set_bind_address(dqlite_server *server,
					   const char *address);

/**
 * Customize the function that the server will use to connect to other servers.
 *
 * The same callback will also be used by the client returned by dqlite_server_client().
 * 
 * The callback may be invoked concurrently on multiple threads, and is responsible for handling
 * any required synchronization internally.
 */
int dqlite_server_set_connect_func(dqlite_server *server,
				    dqlite_connect_func connect,
				    void *connect_arg);

/**
 * Configure the target number of voting nodes for the server's cluster.
 *
 * This must be an odd number and should be set to the same value
 * on all servers of the cluster.
 */
int dqlite_server_set_voters(dqlite_server *servers, unsigned n);

/**
 * Configure the target number of standby nodes for server's cluster.
 *
 * This must be set to the same value on all servers of the cluster.
 */
int dqlite_server_set_standbys(dqlite_server *server, unsigned n);

/**
 * Launch the server.
 * 
 * If it the first time the server is started, and it's not the bootstrap
 * server, it will attempt to join the target cluster configured using
 * dqlite_server_set_peers().
 */
int dqlite_server_start(dqlite_server *server);

/**
 * Create a client for connected to server's cluster.
 *
 * This a helper around dqlite_open(), that automatically passes it the members
 *  of server's cluster.
 */
int dqlite_server_client(dqlite_server *server, const char *name, dqlite **db);
```

include/dqlite/client.h
freeekanayaka: > It's called `dqlite_open` by analogy with `sqlite3_open`, since it was designed to be the direct analogue of that function. But I don't want to spend time debating the name, so I'll change it to dqlite_cluster_cluster if you feel strongly.

I'm saying that we'll need a way to create a client **without** also starting a node, for example if you want to implement a dqlite shell (similar to the current `cmd/dqlite/dqlite.go` program in Go dqlite).

Usually you don't need to run a server in order to create a client, the client can be created independently, and a common naming for creating clients is `open`. We definitely would have `dqlite_open`, analogous to `sqlite3_open`, that's not under debate, but it would be purely a client, without coupling it with the server. I'm feeling strongly about anything, just trying to give my contribution.

freeekanayaka:
> I personally don't think duplicating API surface between dqlite_node and dqlite_cluster (or whatever we end up calling it) is a problem (it's not very much code and exposing a new option on dqlite_cluster is a very small diff) -- and I'd rather not block work on the client interface on making a 2.0 of the server interface. Maybe it's possible to work on the client independently for now, but in a way that keeps the option of updating the server as you describe later on?

It's not about the diff or about duplication. It's about the fact that `dqlite_node` should not be used and keeping both API sets just promotes confusion.

What I'm proposing would not block client work in any way, I'm not sure why you think so. It's a very small name change on top of what you are suggesting, to help users understand what API they should use and not end up with a confusing set of two APIs, one of which is effectively legacy (`dqlite_node`) and the other one with a name that feels to be managing clusters (`dqlite_cluster`) but it's actually managing nodes and is effectively the successor of `dqlite_node`. So, in practice what I'm suggesting is `s/dqlite_cluster/dqlite_server/` moving it to `dqlite.h` + deprecating `dqlite_node`. That's it. YMMV, it's your work..

freeekanayaka:
> > ```
> > int dqlite_server_set_peers(dqlite_server *server,
> >   			const char **peers,
> >   			unsigned n_peers);
> > ```
> 
> I think this interface is not as good as `dqlite_cluster_declare_node` (happy to bikeshed the name):
> 
> * it requires an additional allocation for the array of peers
> * it does not support declaring the node's own abstract address with DQLITE_CLUSTER_ME

That'd done with `dqlite_server_set_address`.

> or the bootstrap node's abstract address with DQLITE_CLUSTER_BOOTSTRAP. The first is necessary so each instance (I've been using this term in my head for the pair of "server + associated client") can add itself to the cluster on startup.

Right, that be:

```C
dqlite_server_set_address(s, "1.2.3.4")
const char *peers[1] = {"3.4.5.6"} 
dqlite_server_set_peers(s, peers, 1);
```

Assuming `3.4.5.6` is the address of an existing server, then the server in the code above will start with address `1.2.3.4` and join the cluster.

> and the second is useful to optimize the first startup and also to avoid this caveat in the new docstring:
>   > If the server is meant to be the first server of a new cluster (i.e. it's the "bootstrap" server), then this function should not be called or it should be passed an empty list of peers.

The thing is that you don't generally need to pass `DQLITE_CLUSTER_BOOTSTRAP`. Any address of an running existing server in the cluster is sufficient. Actually, after some days of weeks the original `DQLITE_CLUSTER_BOOTSTRAP` node might have even be removed from the cluster. What I'm trying to say is that in order for a new node to start and join the cluster all it needs is its own address and the address of at least one other node in the cluster. It does not need all nodes and it does not need to know what of those nodes is the bootstrap node. So requiring or encouraging passing that information seems an extra burden. Again, YMMV.

include/dqlite/client.h
freeekanayaka: It's not much effort. You just use a different name and leave the door open for the future. It might come in 3 years, but if the project grows as it should it will, so planning is always good, especially for public APIs.

To be clear, my ones are observation and hints. You're of course the one in charge and you can take them into account or not.

freeekanayaka:
> Sorry, it's my fault for responding glibly while I was feeling frustrated and not taking the time to understand what you were suggesting. I appreciate your feedback and will try to be more measured.

Ok, no worries!

> 
> > The thing is that you don't generally need to pass DQLITE_CLUSTER_BOOTSTRAP. Any address of an running existing server in the cluster is sufficient. Actually, after some days of weeks the original DQLITE_CLUSTER_BOOTSTRAP node might have even be removed from the cluster. What I'm trying to say is that in order for a new node to start and join the cluster all it needs is its own address and the address of at least one other node in the cluster. It does not need all nodes and it does not need to know what of those nodes is the bootstrap node.
> 
> My thinking -- which I didn't make sufficiently clear, sorry -- was that it's useful to know which node is the bootstrap node, because that node will be the leader when the cluster is starting for the first time, so that's where we should send the initial ADD request for each starting node. But that might not be important.

It's important, but there's a solution, because if you don't know who the leader is but you know the address of any other existing node, than that node can tell you who the leader is, and the ADD can be directed to it. You'll likely going to get that behavior for free, if you use the client machinery to send the ADD request: the client will have code to find the leader anyways.

So the problem of not knowing what the bootstrap node is, has a solution.

However **requiring** to know what the bootstrap node is, has no easy solution I think, for example in the case you're adding a node much later in the lifecycle of the cluster, and the original bootstrap node is not there anymore. You can surely make the proposed version of `dqlite_cluster_declare_node` workaround that, but I'm raising a flag here to point that there are some rough edges that might end up being confusing.
 
> I think your proposed split of dqlite_server_set_address + dqlite_server_set_peers and my dqlite_cluster_declare_node accomplish the same goals, and I'd be fine with either of them.
> 

Indeed they are mostly equivalent, and there are drawbacks in each case. We can also think about other options. I feel that talking about these aspects is important, it's easy to get public APIs out in the wild, but once they are out you can't take them back. So in my mind it's worth spending time designing an aspect that will be around for years, to make sure most of the ramifications have been taken into account. I personally don't feel completely set on either of the two proposals yet, and would be happy to think about them more and looking at them from various angles.

Note that this shouldn't block the implementation, since these type of API-level changes are using not very laborious once the core implementation is there.

> As for the dqlite_node -> dqlite_server proposal, again, I have to apologize for being glib and responsing without taking the time to digest your comment. I designed dqlite/client.h to be completely independent from dqlite.h, with the idea that everything you'd need to e.g. replace SQLite with dqlite in your application would be declared in dqlite/client.h, leaving dqlite.h as the basis for consumers like go-dqlite that want to take the client side into their own hands. But it would also be perfectly fine to do what you propose and have dqlite_server as a type that's shared between those two headers, and I'll defer to you on which is better.

We should take everything into account, and see how everything fits in the broader picture.

I think we're mostly saying the same thing: for what I can see the `dqlite/client.h` you are describing is what **every** dqlite consumer will want to use, including go-dqlite which implements its own client (or any other language binding that implements its own client). The reason is that the `dqlite/client.h` in this PR is not only a client, but also has the `dqlite_cluster` features which are essential for all dqlite consumers. If `dqlite_cluster` had been there before `go-dqlite` existed, `go-dqlite` would have used that instead of re-inventing the wheel with the `app.App` package.

I don't have an opinion about whether `dqlite.h` should be separate from `dqlite/client.h`. We could also put everything in `dqlite.h`, since we're most probably not going to ship 2 different shared libraries (one for the client and one for the server), so having separate headers is somehow arbitrary. Unless I'm missing something.

If we put everything in `dqlite.h`, the goal of having a single file that has everything you need to replace SQLite in your application is preserved. And the goal of exposing the newer and richer `dqlite_cluster`/`dqlite_server` API more properly would be preserved too: it would supersede the deprecated `dqlite_node` API, and non-C consumers like go-dqlite that implement their own native client would use the `dqlite_cluster`/`dqlite_server` C bindings for the server, instead of `dqlite_node` which is weaker and puts more burden on them.

freeekanayaka:
> Okay, putting everything in dqlite.h and making dqlite_cluster (or whatever we're calling it) a full replacement for dqlite_node sounds good to me. Thanks for the feedback!

Sounds good thanks! I'd suggest `dqlite_server` to make it clear that its functions act on a single node and not on the cluster as a whole.

include/dqlite.h
```diff
@@ -17,8 +17,278 @@
...
+ * The abstract address will be passed to the custom connect function configured
+ * with dqlite_server_set_connect_func, or to the default connect function if no
+ * custom function is configured. The default connect function accepts IPv4
+ * addressed, IPv6 addresses, and Unix-domain socket addresses; for a complete
+ * description of the syntax, see dqlite_server_set_bind_address.
+ *
+ * No reference to @path or @addr is kept after this function returns.
+ */
+int dqlite_server_create(const char *path,
+			 const char *addr,
```
freeekanayaka, line 74: Any particular reason for which we are requiring the `addr` parameter?

I would expect that this bit of information is only needed for the very first time the server is started (similarly to what happens to other bits of information such as the peers), and will then be persisted in a file under `path` (along with some other info, such as an auto-generated server ID).

It seems we could just have:

```C
int dqlite_server_create(const char *path, **dqlite_server);
```

instead. And:

```C
int dqlite_server_set_address(dqlite_server* server, const char *address);
```

for the initial configuration, the very first time a server is started (it would be harmless to call `dqlite_server_set_address` unconditionally, like for the other functions).

include/dqlite.h
```diff
@@ -17,8 +17,278 @@
...
+ * @server to act as the bootstrap server itself, pass @server's own abstract
+ * address to this function. At most one bootstrap server may be specified. The
+ * bootstrap server declaration only has an effect when a server is starting up
+ * for the first time (but it is harmless to call this function
+ * unconditionally).
+ *
+ * No reference to @addr is kept after this function returns.
+ */
+int dqlite_server_set_bootstrap_address(dqlite_server *server,
+					const char *addr);
```
freeekanayaka, line 92: Couldn't this just be:

```C
int dqlite_server_set_bootstrap(dqlite_server *server);
```

or possibly:

```C
int dqlite_server_set_bootstrap(dqlite_server *server, bool enable);
```

?

Or whatever equivalent naming/semantics.

I would expect this function to be called only on the server that is going to be the bootstrap server, so there's no need to pass it the address since we already have it.

There's no need to call this function on other servers, since calling `dqlite_server_set_peer_address` is enough (but see also the comment I made to `dqlite_server_set_peer_address`).

include/dqlite.h
```diff
@@ -17,8 +17,278 @@
...
+
+/**
+ * Declare the abstract address of another server that will participate in a
+ * cluster with this server.
```
freeekanayaka, line 96: As far as I can tell the purpose of this function is to let a new server that is starting for the first time to join an existing cluster. So I'd change this docstring to:

```C
/**
 * Set the address of an existing server of the cluster that this server should join.
```

or any equivalent wording.

Note that any address of an existing server in the cluster is fine and that the address of the bootstrap node is not needed for joining (the bootstrap node might not even exist anymore, if it was removed from the cluster).

freeekanayaka: I mean, the address of the bootstrap node is not needed specifically. Of course you could pass it to `dqlite_server_set_peer_address`, but you don't need to know that it s the bootstrap server, and also if you don't pass it at all, any other address of an existing server is fine. Hope it's clear what I mean.

include/dqlite.h
```diff
@@ -17,8 +17,304 @@
...
+				   void *arg);
+
+/**
+ * Configure the target number of voting nodes for this server's cluster.
+ *
+ * When this server is the cluster leader, it will try to promote and demote
+ * other servers to reach the target voter count. You should configure the same
+ * target on all servers that will participate in the cluster.
+ */
+int dqlite_server_set_target_voters(dqlite_server *server, int n);
```
freeekanayaka, line 150: The fact that we require to configure the same value on all servers is a bit unfortunate. I believe we have a few cluster-wide settings (target voters, target standbys and network latency come to my mind) that we currently configure/persist on a per-server basis, and could instead be saved either in the raft log or (better) in a dqlite table or db. In that way we would not require users to book-keep these settings and to make sure they are in sync across all servers.

For example, setting or changing the number of target voters could be done by running a bit of SQL, like:

```
PRAGMA voters=5;
```

or `PRAGMA dqlite.voters=5;` or `UPDATE dqlite.config SET value=5 WHERE key='voters'` or whatever other SQL.

See also https://www.sqlite.org/pragma.html.

This will probably be more work to implement, though.

If it's critical to expose these settings in the very first iteration of this C client support, it'd be also okay to leave `dqlite_server_set_target_voters` and `dqlite_server_set_target_standbys` as they are in this PR. In the future we could implement the more convenient cluster-wide SQL-based setting, and make it take precedence if used.

If it's not so critical to have these settings exposed in the very first iteration of the C client (after the default settings should be good enough for most use cases), maybe we could not implement these functions right now, and expose these settings later via SQL.

include/dqlite.h
```diff
@@ -17,8 +17,304 @@
...
+ * string, as passed to dqlite_server_create or dqlite_server_set_peer. @fd is
+ * an address where a socket representing the connection should be stored. The
+ * callback should return zero if a connection was established successfully or
+ * nonzero if the attempt failed.
+ */
+typedef int (*dqlite_connect_func)(void *arg, const char *addr, int *fd);
+
+/* The following functions (dqlite_server_* and the functions that mimic the
+ * sqlite3 API) use SQLite result codes. dqlite tries to return result codes
+ * that match what SQLite would return in the same situation, but in some cases
```
freeekanayaka, line 44: Returning SQLite result codes surely makes sense for the functions that mimic the sqlite3 API.

However I'm not sure about the `dqlite_server` ones. It feels we might end up shoehorning them, since the failure modes of the `dqlite_server` APIs are a bit different and if we need to **add** ad-hoc codes for `dqlite_server`, then having a different set altogether might be a saner choice.

This is a detail that can be revisited later (either in one way or the other) when things become more clear.

include/dqlite.h
```diff
@@ -17,8 +17,304 @@
...
+/**
+ * Open a database connection on the dqlite cluster.
+ *
+ * This request will be transparently forwarded to the cluster leader as needed.
+ *
+ * This is the analogue of sqlite3_open. @name is the name of the database
+ * to open, which will be created if it does not exist. All servers in the
+ * cluster share a "namespace" for databases.
+ */
+int dqlite_open(dqlite_server *server, const char *name, dqlite **db);
```
freeekanayaka, line 206: One idea to make `dqlite_open` a bit more future-proof (I'm thinking to the goal of supporting creating clients without requiring a locally running server, for example for implementing interactive remote shells) would be to change this signature to something like:

```C
int dqlite_open(void *connector, const char *name, dqlite **db, int flags);
```

where the meaning of `connector` would be defined using a flag.

For example:

```C
dqlite_server *server;
dqlite_server_create(..., &server);
...
dqlite_open(server, "my-db", DQLITE_OPEN_USING_SERVER);
```

or similar names. In the future we could also support passing something that is not a server instead, but rather some other object that just knows about the addresses of some servers in the cluster and that can connect to them.

freeekanayaka:
> I agree that kind of extensibility could be useful in the future, but it seems like it would be simple enough to do what SQLite does by adding a dqlite_open_v2 with the signature you suggest in a future release, while offering dqlite_open with the simple semantics. Does that make sense to you?

I'm personally okay with the `_v2` approach for changing/extending existing APIs. There was some push back on this approach from other people in the past, but if everyone on the current dqlite team is fine with that we could decide that the `_v2`-type approach is what we'll do in general in these cases and so we don't need to future-proof new APIs too hard.

include/dqlite.h
```diff
@@ -17,8 +17,307 @@
...
+int dqlite_server_set_target_voters(dqlite_server *server, int n);
+
+/**
+ * Configure the target number of standby nodes for this server's cluster.
+ *
+ * When this server is the cluster leader, it will try to promote and demote
+ * other servers to reach the target standby count. You should configure the
+ * same target on all servers that will participate in the cluster.
+ */
+int dqlite_server_set_target_standbys(dqlite_server *server, int n);
```
freeekanayaka, line 148: It would probably make sense to change `int n` into `unsigned n`, since this value is supposed to be positive (same for `dqlite_server_set_target_voters`).

freeekanayaka:
> I'm just pushing up implementation commits now, one or two functions at a time. Feel free to review, but I might still rewrite history as I notice mistakes/points of incompleteness

Thanks.

Would it be possible to break this down into several PRs? Ideally each PR would implement a coherent bit of work towards the final goal, with its own set of commits.

For example:

- https://github.com/canonical/dqlite/pull/501/commits/d9c61bf13506158f88c920c14eeeeb0b720b8c4b could already be turned into a standalone PR. It's a self-contained change that is easy to review and can be committed, without carrying it along with a very big PR (it would be just additional noise).

- the changes to `dqlite.h` (https://github.com/canonical/dqlite/pull/501/commits/c23eb91ec52e2894b2bc9a7ff10d822bc5320011 and https://github.com/canonical/dqlite/pull/501/commits/853c8df497b1bc8e3868143b151737c86d06350a) could already be committed, although there are still of pending comments (https://github.com/canonical/dqlite/pull/501/files#r1177036513 and https://github.com/canonical/dqlite/pull/501/files#r1180086935) that would be worth addressing. It's okay that those interfaces are not yet implemented.
- the `struct node_store_cache` sub-object and all the functions used to create it, access it, modify it and destroy it could be part of a standalone PR with its own set of commits
- the `struct dqlite_server` object and all its related functions can be part of a standalone PR with its own set of commits
- the `struct dqlite` object and all its functions (i.e. the actual client) can be part of a standalone PR with its own set of commits. Since there are a lot of functions there, maybe they could even be split into more PRs, since some of them might turn to be relatively of big and complex on their own. If there are helper sub-objects and associated functions (similarly to the `struct node_store_cache` case) they could be committed as separate PRs.

Note that these PRs don't need to be the "final" version of the code they contain, they can be just minimum viable versions that might be subject to further changes and incremental PRs (either to extend them or to change them a bit).

All of this would make reviewing (our side) and addressing review comments (your side) much easier and more manageable.

freeekanayaka:
PS: you can also wait a bit longer before start pushing the first PRs, if the associated code (for example, say, `struct node_store_cache`) is still highly unstable and you need more time to stabilize it and rewrite history, as you mentioned yourself.

But it'd be still be beneficial to have that work broken down into several PRs and sub-commits, which, again, by no means need to be final. Even in case of change it will be easier to review incrementally.

Issue #504, opened 2023-04-24
Raft uv_timer leak when creating and destroying node


MathieuBordere:
Reported by @stgraber, see failures here https://jenkins.linuxcontainers.org/job/lxd-github-commit/4903/ 

It's related to a [TODO](https://github.com/canonical/dqlite/blob/c3ac4d8ad19f7057a5723a0e4b8d17d1d59fb079/src/server.c#L84) . This [timer](https://github.com/canonical/raft/blob/cabe35b941c00e838dd289507b7bff21d8018718/src/uv.c#L119) is not properly closed when only calling `dqlite_node_create` and `dqlite_node_destroy` without starting and stopping the node. This sequence of calls is hit in the node recovery logic, where we explicitly don't want to run the node and made visible by the newly introduced assert in [this](https://github.com/canonical/dqlite/commit/08f8d0748fe2ce4b9ae025bb8eb545f86f14bd86#diff-1abc5651133d108c0c420d9411925373c711133e7748d9e4f4c97d5fb543fdd9) change.

I will remove the newly introduced assert to fix the LXD test suite and track the proper fix in this issue.

PR #505, opened 2023-04-24
server: Remove assert of uv_loop_close rv.

MathieuBordere:
The assert can be reintroduced once the TODO has been fixed.

See #504


freeekanayaka:
Thanks for the detailed explanation. Please could you also add a reference to #504 in the comment added by this PR? It has valuable info to understand the situation.

PR #506, opened 2023-04-28
Polyfill __builtin_assume_aligned for tcc

cole-miller:
I've been messing around with using Fabrice Bellard's tcc for local builds; this is the only change that's really required for libdqlite to build successfully (have not tried the test suite yet).

The only reason we call __builtin_assume_aligned is to silence a warning from GCC/Clang about unaligned accesses; tcc doesn't have that warning.

### DIFF
--- a/src/format.c
+++ b/src/format.c
@@ -8,6 +8,12 @@
 
 #include "format.h"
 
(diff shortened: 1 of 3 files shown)


freeekanayaka:
Nice.

PR #507, opened 2023-05-02
C client (interface only)

cole-miller:
As suggested, I'm breaking the C client work up into smaller PRs. This one just has the changes to dqlite.h, already discussed in #501.


include/dqlite.h
```diff
@@ -17,8 +17,307 @@
...
+				   void *arg);
+
+/**
+ * Configure the target number of voting nodes for this server's cluster.
+ *
+ * When this server is the cluster leader, it will try to promote and demote
+ * other servers to reach the target voter count. You should configure the same
+ * target on all servers that will participate in the cluster.
+ */
+int dqlite_server_set_target_voters(dqlite_server *server, int n);
```
freeekanayaka, line 139: I'll copy here the unaddressed comment from the other PR https://github.com/canonical/dqlite/pull/501#pullrequestreview-1400729689:

The fact that we require to pass to these functions the same value on all servers is a bit unfortunate. I believe we have a few cluster-wide settings (target voters, target standbys and network latency come to my mind) that we currently configure/persist on a per-server basis, and could instead be saved either in the raft log or (better) in a dqlite table or db. In that way we would not require users to book-keep these settings and to make sure they are in sync across all servers.

For example, setting or changing the number of target voters could be done by running a bit of SQL, like:

```
PRAGMA voters=5;
```

or `PRAGMA dqlite.voters=5;` or `UPDATE dqlite.config SET value=5 WHERE key='voters'` or whatever other SQL.

See also https://www.sqlite.org/pragma.html.

This will probably be slightly more work to implement, but perhaps it's worth doing instead of introducing the `dqlite_server_set_target_voters/standbys` functions.

If it's critical to expose these settings in the very first iteration of this C client support, it'd be also okay to leave `dqlite_server_set_target_voters` and `dqlite_server_set_target_standbys` as they are in this PR. In the future we could implement the more convenient cluster-wide SQL-based setting, and make it take precedence if used.

If it's not so critical to have these settings exposed in the very first iteration of the C client (after all the default settings should be good enough for most use cases), maybe we could also not implement these functions right now, and expose these settings later via SQL as described above.

I'm mainly interested in pointing this out for discussion and come up with a plan.

include/dqlite.h
```diff
@@ -17,8 +17,307 @@
...
+
+/**
+ * Configure the target number of standby nodes for this server's cluster.
+ *
+ * When this server is the cluster leader, it will try to promote and demote
+ * other servers to reach the target standby count. You should configure the
+ * same target on all servers that will participate in the cluster.
+ */
+int dqlite_server_set_target_standbys(dqlite_server *server, int n);
```
freeekanayaka, line 148: Copied from https://github.com/canonical/dqlite/pull/501#pullrequestreview-1405464373:

It would probably make sense to change `int n` into `unsigned n`, since this value is supposed to be positive (same for `dqlite_server_set_target_voters`).

include/dqlite.h
freeekanayaka: Ok. Then I would suggest to remove the `dqlite_set_target_voters/standbys` from this PR for the moment and instead open an issue for implementing the cluster-wide approach.

After everything else (`dqlite_server` implementation plus `dqlite` client) has been completed and merged we can come back to that issue and decide exactly what to do. It should be relatively easy to implement.

Worst case scenario, if there is no time left to implement the cluster-wide approach, we can opt for `dqlite_set_target_voters/standbys` as temporary solution (those functions are really trivial to implement), but I really don't think we'll need that. I'm just mentioning this possibility for sake of completeness.

include/dqlite.h
freeekanayaka: Ok, it's a mistake in `go-dqlite` then. We don't need to worry about it right now though, especially if we remove `dqlite_server_set_target_voters/standbys` as suggested above.

include/dqlite.h
```diff
@@ -17,8 +17,289 @@
...
+ * server in each cluster. After the first startup, the bootstrap server is no
+ * longer special and this function is a no-op.
+ */
+int dqlite_server_set_bootstrap(dqlite_server *server);
+
+/**
+ * Declare the address of an existing server in the cluster, which should
+ * already be running.
+ *
+ * This function
```
MathieuBordere, line 86: something went wrong here

freeekanayaka (approved):
LGTM, thanks!

MathieuBordere (approved):

PR #511, opened 2023-05-04
dqlite.h: Introduce DQLITE_API

cole-miller:
By analogy with RAFT_API. As a complement, add `-fvisibility=hidden` to the CFLAGS for libdqlite.a.

### DIFF
--- a/include/dqlite.h
+++ b/include/dqlite.h
@@ -9,6 +9,9 @@
 #define DQLITE_API __attribute__((visibility("default")))
 #endif
 
+/* XXX */
+#define DQLITE_VISIBLE_TO_TESTS DQLITE_API
+
 /**
  * Version.
  */
--- a/src/client/protocol.h
+++ b/src/client/protocol.h
@@ -3,6 +3,8 @@
 #ifndef DQLITE_CLIENT_PROTOCOL_H_
 #define DQLITE_CLIENT_PROTOCOL_H_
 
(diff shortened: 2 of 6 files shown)


freeekanayaka:
> Ah, the tests need to be able to call things from client/protocol.h, let me fix that.

I didn't look at the failure nor the test code, but the idea would that any test that exercises internal parts of dqlite would go under `test/unit` (which statically links the code and has access to all symbols, i.e. white box approach), and tests that exercise the user API would under `test/integration` (which dynamically links libdqlite.so and runs the exactly as a user would).

freeekanayaka:
> > Ah, the tests need to be able to call things from client/protocol.h, let me fix that.
> 
> I didn't look at the failure nor the test code, but the idea would that any test that exercises internal parts of dqlite would go under `test/unit` (which statically links the code and has access to all symbols, i.e. white box approach), and tests that exercise the user API would under `test/integration` (which dynamically links libdqlite.so and runs the exactly as a user would).

We might need to temporarily put in place some hack to allow things to call `client/protocol.h` symbols, but with the goal of cleaning that up at some point.

freeekanayaka:
Looks good to me, and it's indeed better than before.

Please would you file an issue for cleaning this up properly in the future? And reference this PR in it. Thanks

freeekanayaka (approved):

Issue #291, opened 2021-04-21
Implement a C client library


stgraber:
Currently our main client is go-dqlite, but it'd be good to have a standard C client library which would allow for porting of traditional sqlite3 codebases over to dqlite. This should somewhat closely follow what's present in sqlite3 but with added functions to setup and communicate with the cluster.

w-oertl:
I'm interested in using dqlite in a project implemented in C on Linux. Now I have a server process that handles RPC from clients,  executing various SQL commands on a local database. The clients do not send raw SQL statements, only function numbers with arguments (MsgPack based). This ensures that all accesses can be properly authenticated and traced.

I thought that I could simply link the server to dqlite, add an initialization call, open my database with this special vfs, start three instead of one server, and voilá, replication and scalability. Unfortunately, it seems to be far more complicated. I don't quite grok the documentation, and there seems no simple C based example of this obvious use case.

If I see to it that all clients connect to the leader for write transactions, and to one of the others for read transactions, reacting properly to changes in leadership, wouldn't that render redirecting writes to the leader unnecessary? But how to determine what node is the current leader?

This idea of a C client library sounds like what I've been looking for in vain. I've eyed rqlite, but it's even more Go and must be used via an HTTP connection, so it's even further from what I'm looking for.

So please enlighten me - is this use case possible at all? Would I have to completely revamp my project to use rqlite? Thanks!

freeekanayaka:
> I thought that I could simply link the server to dqlite, add an initialization call, open my database with this special vfs, start three instead of one server, and voilá, replication and scalability.

That's not possible, because as you point below the writes need to go to the leader, so you need some sort of connection/protocol from one of your server to the leader and issue SQL statements via that connection protocol.

> Unfortunately, it seems to be far more complicated. I don't quite grok the documentation, and there seems no simple C based example of this obvious use case.

There's such an example in Go (look at the [demo program](https://github.com/canonical/go-dqlite/blob/master/cmd/dqlite-demo/dqlite-demo.go)). There reason there's not an example in C is because there's no C client library (i.e. a client that speaks the protocol mention above, that sends SQL statements to the leader).

> If I see to it that all clients connect to the leader for write transactions, and to one of the others for read transactions,

At the moment they also need to connect to the leader for read transactions. Support for read transactions on non-leader node would be possible but is not yet implemented.

> reacting properly to changes in leadership, wouldn't that render redirecting writes to the leader unnecessary?

I'm not sure what you mean here.

[...]

PR #514, opened 2023-05-19
Implement dqlite_server API

cole-miller:
Finally! Still need to add a test or two, but I have run a sample program through the whole dqlite_server lifecycle successfully on my machine. Progress on #291 -- this contains only the dqlite_server_* functions, with the SQLite-like parts of the API left for a separate PR.

### DIFF
--- a/test/integration/test_server.c
+++ b/test/integration/test_server.c
@@ -1,0 +1,270 @@
+#include "../../include/dqlite.h"
+#include "../../src/server.h"
+#include "../lib/fs.h"
+#include "../lib/munit.h"
+#include "../lib/runner.h"
+
+#include <stdio.h>
+#include <time.h>
+
+SUITE(server);
+
+#define N_SERVERS 3
+
+struct fixture
+{
+	char *dirs[N_SERVERS];
+	dqlite_server *servers[N_SERVERS];
+};
+
+static void *setup(const MunitParameter params[], void *user_data)
+{
+	(void)params;
+	(void)user_data;
+	struct fixture *f = munit_malloc(sizeof *f);
+	unsigned i;
+	int rv;
+
+	for (i = 0; i < N_SERVERS; i += 1) {
+		f->dirs[i] = test_dir_setup();
+		rv = dqlite_server_create(f->dirs[i], &f->servers[i]);
+		munit_assert_int(rv, ==, 0);
+	}
+
+	return f;
+}
+
+static void teardown(void *data)
+{
+	struct fixture *f = data;
+	unsigned i;
+
+	for (i = 0; i < N_SERVERS; i += 1) {
+		dqlite_server_destroy(f->servers[i]);
+		test_dir_tear_down(f->dirs[i]);
+	}
+	free(f);
+}
+
+#define PREPARE_FILE(i, name, ...)                              \
+	do {                                                    \
+		char path[100];                                 \
+		snprintf(path, 100, "%s/%s", f->dirs[i], name); \
+		FILE *fp = fopen(path, "w+");                   \
+		fprintf(fp, __VA_ARGS__);                       \
+		fclose(fp);                                     \
+	} while (0)
+
+#define NODE(x) x
+
+#define NODE0_ID "3297041220608546238"
+
+void start_each_server(struct fixture *f)
+{
+	const char *addrs[] = {"127.0.0.1:8880", "127.0.0.1:8881"};
+	int rv;
+
+	rv = dqlite_server_set_address(f->servers[0], "127.0.0.1:8880");
+	munit_assert_int(rv, ==, 0);
+	rv = dqlite_server_set_auto_bootstrap(f->servers[0]);
+	munit_assert_int(rv, ==, 0);
+	f->servers[0]->refresh_period = 100;
+	rv = dqlite_server_start(f->servers[0]);
+	munit_assert_int(rv, ==, 0);
+
+	rv = dqlite_server_set_address(f->servers[1], "127.0.0.1:8881");
+	munit_assert_int(rv, ==, 0);
+	rv = dqlite_server_set_auto_join(f->servers[1], addrs, 1);
+	munit_assert_int(rv, ==, 0);
+	f->servers[1]->refresh_period = 100;
+	rv = dqlite_server_start(f->servers[1]);
+	munit_assert_int(rv, ==, 0);
+
+	rv = dqlite_server_set_address(f->servers[2], "127.0.0.1:8882");
+	munit_assert_int(rv, ==, 0);
+	rv = dqlite_server_set_auto_join(f->servers[2], addrs, 2);
+	munit_assert_int(rv, ==, 0);
+	f->servers[2]->refresh_period = 100;
+	rv = dqlite_server_start(f->servers[2]);
+	munit_assert_int(rv, ==, 0);
+}
+
+void stop_each_server(struct fixture *f)
+{
+	int rv;
+
+	rv = dqlite_server_stop(f->servers[2]);
+	munit_assert_int(rv, ==, 0);
+	rv = dqlite_server_stop(f->servers[1]);
+	munit_assert_int(rv, ==, 0);
+	rv = dqlite_server_stop(f->servers[0]);
+	munit_assert_int(rv, ==, 0);
+}
+
+TEST(server, restart_follower, setup, teardown, 0, NULL)
+{
+	struct fixture *f = data;
+	struct timespec ts = {0};
+	int rv;
+
+	/* Between operations we sleep for 200 milliseconds, twice
+	 * the configured refresh period, so that the refresh task
+	 * has a chance to be triggered. */
+	ts.tv_nsec = 200 * 1000 * 1000;
+
+	start_each_server(f);
+
+	nanosleep(&ts, NULL);
+
+	rv = dqlite_server_stop(f->servers[1]);
+	munit_assert_int(rv, ==, 0);
+
+	nanosleep(&ts, NULL);
+
+	rv = dqlite_server_start(f->servers[1]);
+	munit_assert_int(rv, ==, 0);
+
+	nanosleep(&ts, NULL);
+
+	stop_each_server(f);
+
+	return MUNIT_OK;
+}
+
+TEST(server, restart_leader, setup, teardown, 0, NULL)
+{
+	struct fixture *f = data;
+	struct timespec ts = {0};
+	int rv;
+
+	/* Between operations we sleep for 200 milliseconds, twice
+	 * the configured refresh period, so that the refresh task
+	 * has a chance to be triggered. */
+	ts.tv_nsec = 200 * 1000 * 1000;
+
+	start_each_server(f);
+
+	nanosleep(&ts, NULL);
+
+	rv = dqlite_server_stop(f->servers[0]);
+	munit_assert_int(rv, ==, 0);
+
+	nanosleep(&ts, NULL);
+
+	rv = dqlite_server_start(f->servers[0]);
+	munit_assert_int(rv, ==, 0);
+
+	nanosleep(&ts, NULL);
+
+	stop_each_server(f);
+
+	return MUNIT_OK;
+}
+
+TEST(server, bad_info_file, setup, teardown, 0, NULL)
+{
+	struct fixture *f = data;
+	int rv;
+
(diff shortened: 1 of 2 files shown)


include/dqlite.h
```diff
@@ -171,7 +174,10 @@ DQLITE_API int dqlite_server_stop(dqlite_server *server);
 /**
  * Free resources owned by the server.
  *
- * This should be called after dqlite_server_stop.
+ * You should always call this function to finalize a server created with
+ * dqlite_server_create, whether or not that function returned successfully.
+ * If the server has been successfully started with dqlite_server_start,
+ * then you must stop it with dqlite_node_stop before calling this function.
```
freeekanayaka, line 180: I think s/dqlite_node_stop/dqlite_server_stop/

freeekanayaka:
Thanks, diff and commits look definitely more manageable review-wise. I'll wait for the tests before digging deeper.

src/client/easy.c
```diff
@@ -0,0 +1,780 @@
...
+	for (i = 0; i < server->cache.len; i += 1) {
+		role_name =
+		    (server->cache.nodes[i].role == DQLITE_SPARE)
+			? "spare"
+			: ((server->cache.nodes[i].role == DQLITE_STANDBY)
+			       ? "standby"
+			       : "voter");
+		k = fprintf(f, "%s%c%" PRIu64 "%c%s%c",
+			    server->cache.nodes[i].addr, '\0',
+			    server->cache.nodes[i].id, '\0', role_name, '\0');
```
freeekanayaka, line 224: The format of the node store file seems a mix of binary format and text format. It's binary because it's a sequence of 0-terminated strings, but it's also text because some values like the ID and the role are converted to strings.

What about having a fully textual format, that one can easily inspect with `cat`?

Probably something like:

```C
k = fprintf(f, "%s%c%" PRIu64 "%c%s%c",
			    server->cache.nodes[i].addr, ' ',
			    server->cache.nodes[i].id, ' ', role_name, '\n');
```

This would require addresses to have no whitespaces in them, which is probably a reasonable constraint.

There should also be some kind of versioning of the file, something as simple as `format 1` line at the beginning of the file would be enough. Or anything equivalent.

src/client/easy.c
```diff
@@ -0,0 +1,780 @@
...
+	if (server->bootstrap && server->cache.len > 0) {
+		goto err;
+	}
+
+	server->is_new = true;
+	server->dir_fd = open(server->dir_path, O_RDONLY | O_DIRECTORY);
+	if (server->dir_fd < 0) {
+		goto err;
+	}
+	info_fd = openat(server->dir_fd, "info-fresh", O_RDWR | O_CREAT, 0664);
```
freeekanayaka, line 625: What about renaming this to `server.info` or `server-info` or `info.txt`? The `-fresh` bit feels a bit like an implementation detail.

src/client/easy.c
```diff
@@ -0,0 +1,780 @@
...
+	if (info_fd < 0) {
+		goto err_after_open_dir;
+	}
+	store_fd =
+	    openat(server->dir_fd, "node-store-fresh", O_RDWR | O_CREAT, 0664);
```
freeekanayaka, line 630: Same here, I'd suggest `node-store.info` or `node-store.txt` or something along those lines.

src/client/easy.c
```diff
@@ -0,0 +1,780 @@
...
+		      "node-store-fresh");
+	(void)rv;
+}
+
+/**
+ * The format for the local info file consists of concatenated null-terminated
+ * strings: address, then ID (decimal). Example:
+ *
+ * @a\017\0
+ */
```
freeekanayaka, line 242: Same as for the node store file, I'd suggest to make it fully textual and add a format version.

freeekanayaka (changes requested):
First round of comments!

The testing side feels a bit light, as basically no error condition is exercised. I'm personally ok with that, but of course the risk or regressions or bugs not caught at review time is higher. That's something that can be improved later on if desired.

test/integration/test_client_easy.c
```diff
@@ -0,0 +1,116 @@
...
+	char *p;
+	int rv;
+
+	for (i = 0; i < N_SERVERS; i += 1) {
+		strncpy(dirpath, "/tmp/dqlitetestXXXXXX", 100);
+		p = mkdtemp(dirpath);
+		if (p == NULL) {
+			munit_error("mkdtemp");
+		}
+		f->dirs[i] = strdup(dirpath);
```
freeekanayaka, line 34: There is a `test_dir_setup()` helper declared in `test/lib/fs.h`, it will reduce boilerplate/duplication, e.g.:

```C
f->dirs[i] = test_dir_setup();
rv = dqlite_server_create(f->dirs[i], &f->servers[i]);
```

test/integration/test_client_easy.c
```diff
@@ -0,0 +1,116 @@
...
+static void teardown(void *data)
+{
+	struct fixture *f = data;
+	unsigned i;
+	DIR *d;
+	struct dirent *de;
+
+	for (i = 0; i < N_SERVERS; i += 1) {
+		dqlite_server_destroy(f->servers[i]);
+		d = opendir(f->dirs[i]);
```
freeekanayaka, line 51: Similarly, this could be:

```C
test_dir_tear_down(f->dirs[i]);
```

it will remove the directory recursively and free the allocated path string.

test/integration/test_server.c
```diff
@@ -0,0 +1,262 @@
...
+
+	rv = dqlite_server_stop(f->servers[1]);
+	munit_assert_int(rv, ==, 0);
+
+	sleep(2);
+
+	rv = dqlite_server_start(f->servers[1]);
+	munit_assert_int(rv, ==, 0);
+
+	sleep(2);
```
freeekanayaka, line 124: Sleeping for 2 seconds seems a relatively long time for a test, also considering that there are several `sleep(2)` calls in the test, and several tests that do the same, adding to the overall execution time of the test suite.

What is the purpose of sleeping in the first place? Exercise the refresh logic?

Ideally the test should monitor for some specific event (e.g. the node store file being refreshed or something like that) and then unblock when that happens, or timeout after a while.

Alternatively, I think lowering the time to something like 100 or 200 milliseconds should be enough to cover whatever this is supposed to test.

Depending on the goal of the test, some other knobs might need to be tweaked, like lowering the refresh period to 50 milliseconds or lowering the network latency with `dqlite_node_set_network_latency_ms()`.

I'm also fine leaving the `sleep(2)` for now if it's useful and the alternatives are complicated, we can improve this later, but wanted to check what the purpose is. and why the specific value of 2 seconds was chosen. Adding a comment would help too.

src/client/easy.c
freeekanayaka: FWIW the current binary format of dqlite was designed to minimize encoding and memory copy costs: it takes advantage of memory alignment and native binary representation (e.g. for numbers), a bit like https://capnproto.org/ and other zero-copy serialization formats. Performance is of course not relevant here, but since we already have a format and tooling for it I think it'd make sense to use it.

The internal C code/tooling that we have to support this format can be improved if desired, but at the end of the day all encoders are a variation of `my_encode(struct my_object *obj, void **buf /* out */)` and decoding is the way round.

The NUL-terminated format is surely handy for C, probably a bit less for other languages. If you don't want to use the format we already, I'd still suggest to turn this into a newline-terminated (`\n` instead of `0`) which at least has the advantage to be fully textual and more friendly to scripting languages.

src/server.c
```diff
@@ -994,3 +997,831 @@ dqlite_node_id dqlite_generate_node_id(const char *address)
...
+	bool is_new;
+	bool bootstrap;
+	char *dir_path; /* owned */
+	dqlite_node *local;
+	uint64_t local_id;
+	char *local_addr; /* owned */
+	char *bind_addr;  /* owned */
+	dqlite_connect_func connect;
+	void *connect_arg;
+	unsigned refresh_period;
```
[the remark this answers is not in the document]
freeekanayaka: Also, I'd suggest to make this field store the value in milliseconds instead of seconds. This should be handy for integration tests, so we can have this logic trigger faster.

test/integration/test_server.c
```diff
@@ -0,0 +1,262 @@
...
+	munit_assert_int(rv, ==, 0);
+	rv = dqlite_server_start(f->servers[1]);
+	munit_assert_int(rv, ==, 0);
+
+	rv = dqlite_server_set_address(f->servers[2], "127.0.0.1:8882");
+	munit_assert_int(rv, ==, 0);
+	rv = dqlite_server_set_peer_address(f->servers[2], "127.0.0.1:8880");
+	munit_assert_int(rv, ==, 0);
+	rv = dqlite_server_set_peer_address(f->servers[2], "127.0.0.1:8881");
+	munit_assert_int(rv, ==, 0);
```
freeekanayaka, line 88: From the point of view of somebody not familiar with dqlite, this sequence of calls to `dqlite_server_set_peer_address()` seems to be a bit ambiguous.

Usually one would expect that a call like `my_object_set_x(object, some_value)` would set the value of `x` to `some_value` on `object`, and a subsequent call like `my_object_set_x(object, other_value)` would replace the previously set `some_value` with `other_value`.

Also, one would expect that calling `my_object_set_x(object, some_value)` two times in a row is idempotent.

However the current semantics of `dqlite_server_set_peer_address()` is different from the "intuitive" one, and one must read the documentation to understand the details. For example calling `dqlite_server_set_peer_address(server, "127.0.0.1:8880")` two times is not idempotent, and calling it with another address does not replace the previous address. The current semantics is also somehow different from the one of the other `dqlite_server_set_xxx()`, which are idempotent and overwrite previous values.

What about replacing `dqlite_server_set_peer_address()` with `dqlite_server_set_peer_addresses()`? Or even `dqlite_server_set_peers()`, for brevity. E.g.

```C
int dqlite_server_set_peers(dqlite_server *s, const char *addresses[], unsigned n);

const char *peers[] = {"127.0.0.1:8880", "127.0.0.1:8881"};
rv = dqlite_server_set_peers(f->servers[2], peers, 2);
munit_assert_int(rv, ==, 0);
```

This would have the intuitive idempotent and overwriting semantics, like the other `set` functions.

freeekanayaka: Maybe also:

```C
int dqlite_server_set_auto_join(dqlite_server *s, const char *peers[], unsigned n);
```

would be even a bit more intuitive. The function turns on the behavior of automatically joining a cluster which has the given peers.

test/integration/test_server.c
freeekanayaka: > > What is the purpose of sleeping in the first place? Exercise the refresh logic?
> 
> Yes, that was the idea. I can shorten the refresh period and the sleep, I think waiting explicitly for the refresh logic to run is tricky without introducing some sort of test-only API to invoke a callback whenever refresh occurs (but maybe that would be a good idea anyway?).

As a general rule, for integration tests it'd be great to keep things mostly as a black-box (no test-only APIs, unless we have too).

For this behavior specifically we could for example watch the file system with `inotify` and check that the node store was refreshed correctly.

Or perhaps later down the road when the client will be in place, it will be easier to put in place a test that depends on the refresh logic to succeed (for instance changing the cluster membership so that the client won't be able to find the leader until the node store is refreshed).

I don't think we don't need to go through the `inotify` complication right now, we can revisit this later on if it will become more practical. Reducing the test time by making the refresh period lower is good enough.

freeekanayaka (approved):
Thanks!

I didn't do a fully deep review (especially for error handling and concurrency/threading), but the external APIs and file formats look good to me, plus the tests cover reasonable ground. Any quirk of the implementation can be refined later if needed.

src/server.h
```diff
@@ -58,6 +59,44 @@ struct dqlite_node
...
+	bool is_new;
+	bool bootstrap;
+	char *dir_path; /* owned */
+	dqlite_node *local;
+	uint64_t local_id;
+	char *local_addr; /* owned */
+	char *bind_addr;  /* owned */
+	dqlite_connect_func connect;
+	void *connect_arg;
+	unsigned refresh_period; /* in milliseconds */
```
MathieuBordere, line 96: I think I'd prefer a 64 bit type, because above you convert to nanoseconds, by multiplying with `1000 * 1000`, so you will quickly overflow a 32bit number.

MathieuBordere (commented):
Looks good, minor comments.

MathieuBordere (approved):

PR #525, opened 2023-06-21
C client API: remaining basic functions

cole-miller:
Not yet ready for final review.


MathieuBordere (commented):
Already looks very good on a first pass imo

src/server.c
```diff
@@ -1786,3 +1789,513 @@ void dqlite_server_destroy(dqlite_server *server)
 	close(server->dir_fd);
 	free(server);
 }
+
+int dqlite_open(dqlite_server *server, const char *name, dqlite **db, int flags)
```
freeekanayaka, line 1793: I'd suggest to move all these new functions to a new `src/client.c` file, since `src/server.c` should be used for server code. Perhaps the `src/client/protocol.{c,h}` files could be moved to `src/`, without the sub-directory.

PR #526, opened 2023-07-19
Expose raft_uv_auto_recovery setting for dqlite nodes

cole-miller:
So that Jepsen can make use of it -- see canonical/jepsen.dqlite#110.

### DIFF
--- a/include/dqlite.h
+++ b/include/dqlite.h
@@ -431,6 +431,20 @@ DQLITE_API int dqlite_node_set_target_voters(dqlite_node *n, int voters);
 DQLITE_API int dqlite_node_set_target_standbys(dqlite_node *n, int standbys);
 
 /**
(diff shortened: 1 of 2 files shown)


include/dqlite.h
```diff
@@ -430,6 +430,18 @@ DQLITE_API int dqlite_node_set_target_voters(dqlite_node *n, int voters);
...
 
+/**
+ * Enable or disable auto-recovery for corrupted disk files.
+ *
+ * When auto-recovery is enabled, files in the data directory that are
+ * determined to be corrupt may be removed by dqlite at startup. This allows
+ * the node to start up successfully in more situations, but comes at the cost
+ * of possible data loss, and may mask bugs.
+ *
+ * Auto-recovery is enabled by default.
```
freeekanayaka, line 441: I'd add that this function needs to be called before starting the node.

Issue #540, opened 2023-11-21
Multi threading bug in tracef()


just-now:
In multi-threading environment, using static quantifier in line 22 near by `static char _msg[1024]; ` may cause data corruption. It's worth removing `static`.
``` 
#define tracef(...)                                                         \                                                                                   
        do {                                                                \                                                                                   
                if (UNLIKELY(_dqliteTracingEnabled)) {                      \                                                                                   
                        static char _msg[1024];                             \                                                                                   
                        snprintf(_msg, sizeof(_msg), __VA_ARGS__);          \                                                                                   
                        struct timespec ts = {0};                           \                                                                                   
                        /* Ignore errors */                                 \                                                                                   
                        clock_gettime(CLOCK_REALTIME, &ts);                 \                                                                                   
                        int64_t ns = ts.tv_sec * 1000000000 + ts.tv_nsec;   \                                                                                   
                        fprintf(stderr, "LIBDQLITE %" PRId64 " %s:%d %s\n", \                                                                                   
                                ns, __func__, __LINE__, _msg);              \                                                                                   
                }                                                           \                                                                                   
        } while (0)                                                                                                                                             
```

MathieuBordere:
fixed in #542

PR #542, opened 2023-11-28
Initial log format rework

just-now:
- Removed static qualifier from tracef() to fix potential multithreading bug;
 - Added log levels.

### DIFF
--- a/src/tracing.c
+++ b/src/tracing.c
@@ -1,14 +1,98 @@
 #include "tracing.h"
-
+#include <stdio.h> /* stderr */
 #include <stdlib.h>
+#include <string.h>      /* strstr, strlen */
+#include <sys/syscall.h> /* syscall */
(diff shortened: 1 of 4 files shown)


MathieuBordere (approved):

src/tracing.c
```diff
@@ -1,14 +1,98 @@
...
+void stderrTracerEmit(const char *file,
+		      unsigned int line,
+		      const char *func,
+		      unsigned int level,
+		      const char *message)
+{
+	assert(tracer__level < TRACE_NR);
+
+	if (level >= tracer__level)
+		tracerEmit(file, line, func, level, message);
```
cole-miller, line 97: We generally use braces even for single-line `if`, `for`, `while`.

cole-miller:
Fixes #540

PR #545, opened 2023-12-04
State cb

### LINKED ISSUE #541: Jepsen: assertion failure in vfs.c
This Jepsen job:

https://github.com/canonical/jepsen.dqlite/actions/runs/6950544166/job/18910920901

Tripped this assertion:

https://github.com/canonical/dqlite/blob/09108b890e26965558f9117aa554513a4956be3b/src/vfs.c#L2343

MathieuBordere:
The state_cb is triggered immediately when the raft state changes,
instead of the monitor_cb, that is only polling for the state change at
the start of a loop iteration. This avoids some bugs where dqlite has to
perform cleanups on leadership loss, but wasn't informed yet of the
leadership loss.

fixes #541 

Note: When we release a new dqlite version, we will also need an 
accompanying new raft version that supports registering the state_cb.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -29,6 +29,26 @@
 
 #define NODE_STORE_INFO_FORMAT_V1 "v1"
 
+/* Called by raft every time the raft state changes. */
+static void state_cb(struct raft *r,
+		     unsigned short old_state,
+		     unsigned short new_state)
+{
+	struct dqlite_node *d = r->data;
+	queue *head;
+	struct conn *conn;
+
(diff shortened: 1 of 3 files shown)


cole-miller (approved):
Looks good to me, could you elaborate on why the test changes are needed?

MathieuBordere:
> Looks good to me, could you elaborate on why the test changes are needed?

The unchanged tests only waited for the leader barrier to be applied, so they didn't test an ex-leader trying to apply frames to it's WAL. The changes explicitly wait for the barrier first and then wait for the `COMMAND_FRAMES` to be applied.

PR #550, opened 2024-01-16
Thread pool

just-now:
Draft


test/unit/ext/test_uv_pool.c
```diff
@@ -0,0 +1,132 @@
...
+	 *
+	 *      - for initialization:
+	 *      uv__handle_unref(&loop->wq_async);
+	 *      loop->wq_async.flags |= UV_HANDLE_INTERNAL;
+	 *
+	 *      - and for finalization:
+	 *      uv_walk() ... if (h->flags & UV_HANDLE_INTERNAL) continue;
+	 */
+	if (count == WORK_ITEMS_NR)
+		xx_loop_async_close(req->loop);
```
cole-miller, line 63: How will we arrange for xx_loop_async_close to be called at the appropriate time in the real dqlite?

cole-miller (commented):
First pass with some nitpicks and higher-level stuff, now going to go through commit-by-commit and make sure I understand the changes that were made from the original code.

src/lib/threadpool.c
```diff
@@ -150,15 +149,25 @@ static void worker(void *arg)
...
+		//XXX printf("worker: idx=%u\n", ta->idx);
+		/* XXX === Assinging work item affinity */
+		if (!xx__queue_empty(&thread_queues[ta->idx])) {
+		    q = xx__queue_head(&thread_queues[ta->idx]);
+		    //XXX printf("worker: q=%p idx=%u\n", q, ta->idx);
+		    xx__queue_remove(q);
+		    xx__queue_init(q);
+		    xx__queue_insert_head(&wq, q);
+		}
+
```
cole-miller: Should the global `wq` be eliminated? It seems like it no longer serves a purpose now that each thread has its own queue

src/lib/threadpool.c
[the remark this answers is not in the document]
just-now: Updated.

src/lib/threadpool.c
```diff
@@ -0,0 +1,454 @@
...
+#define xx__req_unregister(loop, req)              \
+	do {                                       \
+		assert(xx__has_active_reqs(loop)); \
+		(loop)->active_reqs--;             \
+	} while (0)
+
+#define container_of(ptr, type, member) \
+	((type *)((char *)(ptr)-offsetof(type, member)))
+
+#define ARRAY_SIZE(a) (sizeof(a) / sizeof((a)[0]))
```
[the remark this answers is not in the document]
just-now: fixed.

src/lib/threadpool.c
```diff
@@ -0,0 +1,454 @@
...
+static unsigned int idle_threads;
+static unsigned int nthreads;
+static uv_thread_t *threads;
+static struct thread_args *thread_args;
+static uv_key_t thread_key;
+static struct xx__queue *thread_queues;
+static struct xx__queue exit_message;
+static struct xx__queue wq;
+
+static void xx__cancelled(struct xx__work *w __attribute__((unused)))
```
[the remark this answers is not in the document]
just-now: fixed

src/lib/threadpool.c
```diff
@@ -0,0 +1,454 @@
...
+	uv_once_t child_once = UV_ONCE_INIT;
+	memcpy(&once, &child_once, sizeof(child_once));
+}
+
+static void init_once(void)
+{
+	if (pthread_atfork(NULL, NULL, &reset_once))
+		abort();
+	init_threads();
+}
```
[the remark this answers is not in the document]
just-now: fixed.

src/lib/threadpool.h
```diff
@@ -0,0 +1,56 @@
...
+
+	struct xx__queue wq;
+	uv_mutex_t wq_mutex;
+	uv_async_t wq_async;
+	uint64_t active_reqs;
+};
+
+void xx_loop_close(struct xx_loop_s *loop);
+int xx_loop_init(struct xx_loop_s *loop);
+int xx_queue_work(uv_loop_t *loop,
```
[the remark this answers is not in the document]
just-now: from the user's perspective, it's worth leaving `uv_loop_t *` as it is always around unlike `xx_loop_t *`.

test/unit/ext/test_uv_pool.c
just-now: I wanted to call this function from `dqlite/src/server.c:524 // static void stopCb(uv_async_t *stop)`.

just-now (commented):
I'm planning to replace prefix `xx_` elsewhere to something more meaningful.
Any ideas? Maybe `u_` (utils) or `tu_` (thread utils)?

MathieuBordere (commented):
Looks good, minor nitpick. You could try to replcace the current uses of uv_queue_work with this implementation and fire the jepsen tests to see if something is weird.

Makefile.am
```diff
@@ -29,6 +29,7 @@ libdqlite_la_SOURCES = \
   src/lib/buffer.c \
   src/lib/fs.c \
   src/lib/transport.c \
+  src/lib/threadpool.c \
```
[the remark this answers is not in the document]
just-now: not a nit, I need to imbue myself with the style of this project

src/lib/threadpool.c
just-now: I left this queue purposely for the sole reason that we will need to delay serving I/O requests coming to sqlite3 databases in each thread until a set of checkpoint operations is completed. 
In case if we choose this scenario such a queue could be a part of the mechanism.
I propose to leave it for now and remove it (just a matter of 1-2 hours) if needed.

src/lib/threadpool.c
```diff
@@ -0,0 +1,368 @@
...
+			idle_threads -= 1;
+		}
+
+		/* Process work item thread affinity */
+		if (!QUEUE__IS_EMPTY(&thread_queues[ta->idx])) {
+			q = QUEUE__HEAD(&thread_queues[ta->idx]);
+			QUEUE__REMOVE(q);
+			QUEUE__INIT(q);
+			QUEUE__INSERT_HEAD(&wq, q);
+		}
```
cole-miller, line 88: If we switch to having a separate input broadcast channel for checkpoint notifications from the main thread then we'll have to check that before we take an item off the thread-specific work queue, and switch to a different processing loop if a notification is found until we've executed the checkpoint. I'm not sure exactly how to integrate that here though

src/lib/threadpool.c
```diff
@@ -0,0 +1,368 @@
...
+			break;
+		}
+
+		QUEUE__REMOVE(q);
+		QUEUE__INIT(q);
+
+		uv_mutex_unlock(&mutex);
+
+		w = QUEUE__DATA(q, struct xx__work, wq);
+		w->work(w);
```
cole-miller, line 103: I think we will end up having to deviate from the in-order processing of the input queue in at least some cases, and even letting the work callbacks mess with the state of the queue. For example, we have to deal with query transactions that we step part of the way and then suspend because the output queue for the rows is full, but then return to later. But again I'm not quite sure how the write the worker loop in this file to enable this

cole-miller (changes requested):
Thanks for working on this, it's great to see it starting to come together!

cole-miller:
> I'm planning to replace prefix xx_ elsewhere to something more meaningful. Any ideas? Maybe u_ (utils) or tu_ (thread utils)?

No strong feelings, maybe `apool` (for "addressable/affinity pool")?

just-now:
JFYI: The most stable pool without ordering is 207278083cd44e0e5d87ea15a52f48b3bcd266f6

src/lib/threadpool.h
```diff
@@ -0,0 +1,88 @@
...
+	uv_loop_t *loop;       /* The uv loop, item is being associated with */
+	enum pool_work_type type;
+
+	void (*work_cb)(pool_work_t *w);
+	void (*after_work_cb)(pool_work_t *w);
+};
+
+struct pool_s {
+	uint64_t	  magic;
+	uv_loop_t         loop;
```
cole-miller, line 71: I'd rather not embed the loop within the pool like this -- the existing dqlite code already keeps a pool around and stores it in a particular way, having to change that code in order to integrate this new thread pool seems not ideal. It seems that the motivation for embedding the loop like this is to be able to recover a pointer to the pool from a pointer to the loop in callbacks, so could we avoid it by just storing a pointer to the pool in pool_work_s etc.? Also note that uv_async_t has a `void *` userdata field that might be helpful

src/lib/threadpool.h
```diff
@@ -0,0 +1,88 @@
...
+int  pool_init(pool_t *pool);
+void pool_fini(pool_t *pool);
+void pool_close(pool_t *pool);
+void pool_queue_work(pool_t *pool,
+		     pool_work_t *w,
+		     uint32_t cookie,
+		     enum pool_work_type type,
+		     void (*work_cb)(pool_work_t *w),
+		     void (*after_work_cb)(pool_work_t *w));
+pool_t *uv_loop_to_pool(const uv_loop_t *loop);
```
cole-miller, line 84: Let's avoid naming our own functions with the `uv_` prefix so they're not confusable with functions from libuv

src/lib/threadpool.h
```diff
@@ -0,0 +1,88 @@
...
+
+struct pool_s {
+	uint64_t	  magic;
+	uv_loop_t         loop;
+	struct pool_impl *pi;
+};
+
+int  pool_init(pool_t *pool);
+void pool_fini(pool_t *pool);
+void pool_close(pool_t *pool);
```
cole-miller, line 77: I'm wondering about the separation between these two functions, could they be unified?

src/lib/threadpool.c
```diff
@@ -0,0 +1,549 @@
...
+                                           worker and planner threads lock */
+	uint32_t       nthreads;
+	pool_thread_t *threads;
+
+	queue          outq;            /* Output queue used by libuv part */
+	uv_mutex_t     outq_mutex;      /* Output queue lock */
+	uv_async_t     outq_async;      /* Signalled when output queue is not
+                                           empty and libuv loop has to process
+                                           items from it */
+	uint64_t       active_ws;       /* Number of all work items in flight */
```
cole-miller, line 111: Could you add a note that this field is only accessed on the main thread?

src/lib/threadpool.c
```diff
@@ -0,0 +1,549 @@
...
+		.pi = pi,
+	};
+
+	pi->qos = 0;
+	pi->o_prev = WT_BAR;
+	pi->exiting = false;
+	pi->in_flight = 0;
+	pi->nthreads = POOL_THREADPOOL_SIZE;
+
+	val = getenv("POOL_THREADPOOL_SIZE");
```
cole-miller, line 393: Rather than getenv with a fallback in this file I'd prefer that the number of threads be set by an argument to `pool_init` -- the decision of how many threads to use should be made closer to the dqlite entry point

src/lib/threadpool.c
```diff
@@ -0,0 +1,549 @@
...
+                                           items from it */
+	uint64_t       active_ws;       /* Number of all work items in flight */
+
+	queue          ordered;         /* Queue of WT_ORD{N} items */
+	queue          unordered;       /* Queue of WT_UNORD items */
+	struct sm      planner_sm;      /* State machine of the scheduler */
+	uv_cond_t      planner_cond;
+	uv_thread_t    planner_thread;  /* Scheduler's thread */
+
+	uv_key_t       thread_key;      /* TLS: stores thread id */
```
cole-miller, line 119: Is there a use-case for storing the thread ID other than to enable testing? Not a big deal either way but I'm curious

cole-miller (changes requested):
Thanks, this was a pleasure to review! (Sorry it took longer than I thought -- trying to be thorough.)

src/lib/sm.h
```diff
@@ -0,0 +1,43 @@
...
+struct sm_conf {
+	uint32_t    flags;
+	uint64_t    allowed;
+	const char *name;
+};
+
+struct sm {
+	int    rc;
+	int    state;
+	bool (*is_locked)(const struct sm *);
```
MathieuBordere, line 28: so it looks like `is_locked` is optional, but `invariant` is required, maybe make that explicit somewhere in a comment.

src/lib/queue.h
```diff
@@ -47,6 +60,29 @@ typedef void *queue[2];
 		QUEUE__NEXT_PREV(e) = QUEUE__PREV(e); \
 	}
 
+#define QUEUE__SPLIT(h, q, n)                    \
```
MathieuBordere, line 63: A little comment about what the function does would be welcome.

src/lib/threadpool.c
```diff
@@ -0,0 +1,529 @@
...
+	pool_impl_t *pi = pool->pi;
+
+	PRE(threads_nr <= THREADPOOL_SIZE_MAX);
+
+	pi = pool->pi = calloc(1, sizeof(*pool->pi));
+	if (pi == NULL)
+		return UV_ENOMEM;
+
+	*pi = (pool_impl_t) {
+		.qos = 0,
```
MathieuBordere, line 484: How is this QoS functionality exposed to the user?

src/lib/threadpool.c
```diff
@@ -0,0 +1,529 @@
...
+	return q_to_w(q)->thread_id;
+}
+
+static bool planner_invariant(const struct sm *m, int prev_state)
+{
+	pool_impl_t *pi = container_of(m, pool_impl_t, planner_sm);
+	queue *o = &pi->ordered;
+	queue *u = &pi->unordered;
+
+	return ERGO(sm_state(m) == PS_NOTHING, empty(o) && empty(u)) &&
```
MathieuBordere, line 200: I think `planner_invariant` returns `true` when `sm_state(m)` is not one of the enum values, because the first argument to `ERGO` will always be `false`, that's probably unwanted.

MathieuBordere (approved):

src/lib/threadpool.c
just-now: it's impossible to reach such condition due to the following reasons:
sm_init() will move the state to initial state and even if the user will pass incorrect state to sm_move(), PRE(m->conf[sm_state(m)].allowed & BITS(next_state)) will catch this.

src/lib/threadpool.c
just-now: OK, I'll expose it through pool_init().

src/lib/sm.h
just-now: fixed.

just-now:
Closed.
Split into a series of commits and posted as a new PR.

PR #554, opened 2024-01-29
Include more debug info in debug builds

cole-miller:
See [here](https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html#Options-for-Debugging-Your-Program) for the description of what `-g3` does. More debuginfo seems better to me!


MathieuBordere:
Looks good to me.

MathieuBordere (approved):

PR #557, opened 2024-02-02
github: Add downstream check.

MathieuBordere:
Similar as the one in the raft repo.


cole-miller:
Let's see if it works!

PR #569, opened 2024-02-07
Fix many cases of arithmetic on void pointers

cole-miller:
Currently we use `void *` as a cursor type when parsing and serializing, and this results in many cases of performing arithmetic on void pointers -- technically UB. Change these types to `char *` everywhere.

### DIFF
--- a/src/client/protocol.c
+++ b/src/client/protocol.c
@@ -103,7 +103,8 @@ static int peekUint64(struct cursor cursor, uint64_t *val)
 	if (cursor.cap < 8) {
 		return DQLITE_CLIENT_PROTO_ERROR;
 	}
-	*val = ByteFlipLe64(*(uint64_t *)cursor.p);
+	memcpy(val, cursor.p, sizeof(*val));
+	*val = ByteFlipLe64(*val);
 	return 0;
 }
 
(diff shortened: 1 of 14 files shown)


src/client/protocol.c
```diff
@@ -103,7 +103,8 @@ static int peekUint64(struct cursor cursor, uint64_t *val)
 	if (cursor.cap < 8) {
 		return DQLITE_CLIENT_PROTO_ERROR;
 	}
-	*val = ByteFlipLe64(*(uint64_t *)cursor.p);
```
MathieuBordere, line 106: Where is the arithmetic on a void pointer happening in this case (and similar cases) ?

cole-miller: There is no void pointer arithmetic going on, the problem is just that once `cursor.p` is no longer a void pointer the compiler starts to emit a warning about alignment on the cast here. Using memcpy instead gets rid of that.

cole-miller: `char *` has no alignment requirements (and can alias with anything) so changing the type can't create any alignment problems.

cole-miller: But those situations would already have UB with `void *` right?

MathieuBordere: yeah, didn't know void* and char* had same alignment requirements.

PR #596, opened 2024-02-20
Raise aio-max-nr on GHA

cole-miller:


cole-miller:
Ah, it's probeAsyncIO that's failing with EAGAIN.

cole-miller:
Summary: the issue with the flaky test `cluster/restart` is that it somewhat consistently hits the aio-max-nr ceiling in CI during raft_uv startup (probeAsyncIO). This PR addresses that by introducing a retry loop into test_server_start (with exponential backoff). That seems to fix the problem for `cluster/restart`, but `conn/query/run` now runs into a similar issue (somewhat reproducibly), with the error originating here:

https://github.com/canonical/dqlite/blob/c9eda92e7b8d7b5f9c21835cacd346b19c25a03d/src/raft/uv_writer.c#L44-L63

This is trickier to fix because the error is propagated all the way out to the client. We don't want to put a retry loop inside dqlite or raft library code; that should live inside the test suite, wherever we call `clientSendExec`, etc. But this is a large refactoring that I'm not going to attempt right now.

I think aio-max-nr wasn't a problem before in raft or dqlite CI because we didn't run as many tests in parallel. Now that we have the raft tests and the dqlite tests running in the same CI job and at the same time, there's more contention for this scarce resource.

cole-miller:
Raising aio-max-nr to 1000000 seems to reduce the rate at which these tests fail.

PR #597, opened 2024-02-21
Add a state machine library

just-now:


src/utils.h
```diff
@@ -13,4 +14,13 @@
 
 #define DBG() fprintf(stderr, "%s:%d\n", __func__, __LINE__)
 
+#define UNUSED __attribute__((unused))
+
+#define container_of(e, type, field) \
+	((type *)(uintptr_t)((char *)(e)-offsetof(type, field)))
```
cole-miller, line 20: I slightly prefer CONTAINER_OF for function-like macros

src/utils.h
```diff
@@ -13,4 +14,13 @@
 
 #define DBG() fprintf(stderr, "%s:%d\n", __func__, __LINE__)
 
+#define UNUSED __attribute__((unused))
+
```
cole-miller, line 18: Let's use casting to `void` for this purpose instead

cole-miller (changes requested):
Only stylistic comments

src/utils.h
just-now: I have fixed this.

Still i'd notice that it's well known macro people used to write in lowercase. No one would treat it as a function.

src/utils.h
just-now: fixed.

PR #603, opened 2024-02-21
Test no-lz4 configuration in CI

cole-miller:
Right now the build is broken when liblz4-dev is not installed.


cole-miller:
This is needed to fix PPA builds

MathieuBordere (approved):

PR #610, opened 2024-02-28
raft: Close fd used for fallocate probing

cole-miller:
The file descriptor leak here was noted by the LXD folks: https://github.com/canonical/lxd/issues/12808#issuecomment-1969219953

cc @tomponline @simondeziel


tomponline (approved):
Thanks!

PR #630, opened 2024-03-11
Add CI job for latest SQLite, libuv, liblz4

### LINKED ISSUE #565: Run dqlite tests against most recent SQLite release.
In addition to the versions installed by the package managers, add runs to the Github actions workflows that test against the latest released SQLite.

Would have caught #564 earlier.

cole-miller:
Closes #565

### DIFF
--- a/.github/workflows/latest-deps.yml
+++ b/.github/workflows/latest-deps.yml
@@ -16,14 +16,6 @@ jobs:
           sysctl fs.aio-max-nr
           sudo sysctl -w fs.aio-max-nr=1000000
 
-    - name: Purge distro libuv
-      run: |
-          sudo apt remove libuv1
-
-    - name: Install liblz4
-      run: |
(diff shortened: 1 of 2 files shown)


cole-miller:
I took the opportunity to also fix `--enable-build-sqlite` to not require spamming `#pragma GCC diagnostic ignored` :) now it should really be as simple as dropping the amalgamation into the dqlite source root.

The new job gets the latest libuv release by parsing https://dist.libuv.org (which is clearly designed to be machine-readable), and the latest SQLite release by following the helpful instructions [here](https://sqlite.org/download.html#encoding). lz4 just does GitHub releases so I fell back to good old `git tag | sort -V -r`.

cole-miller:
I don't know what's going on with the segfaults but marking as draft for now just to be safe

cole-miller:
Ready to go!

PR #632, opened 2024-04-03
sm: Trace states by name

cole-miller:
This makes the traces easier to read.


just-now (approved):
+1.
Thanks for adding state names to transition logs..

PR #639, opened 2024-04-23
Add local_buf and is_local to raft_entry

cole-miller:
The local_buf field stores data that is persisted locally (and stored in the in-memory log) but not send in AppendEntries messages. This will be used to store the vfs2_wal_slice when we introduce vfs2.

Each local_buf always has its own heap allocation---the `batch` is not used.

The is_local field is not persisted nor sent in AppendEntries. It exists only in the in-memory log and is true when the entry in question originated on the current node when it was leader. This will be used during vfs2 integration to determine whether to use vfs2_unhide or vfs2_commit when a COMMAND entry is applied.


cole-miller:
This change breaks compatibility with all extant libraft releases, so as part of the same PR we need to disable the "external libraft" CI job and remove the option from the build system.

cole-miller:
Marking as draft while I resolve the reports from ASan (which I wasn't seeing locally...)

cole-miller:
Failing tests on Jammy are due to https://github.com/microsoft/linux-package-repositories/issues/130.

just-now (approved):
+1

PR #640, opened 2024-04-23
Three portability fixes around integer width

cole-miller:
This unbreaks the dqlite build on alpha and x32. Supersedes #634; thanks @gibmat for the patch.


gibmat:
Looks good, thanks!

PR #642, opened 2024-04-24
Prepare to merge dqlite-next into master

cole-miller:
Let's do all our work on one branch, with experimental stuff (like thread pool and vfs2 integration) guarded behind `#ifdef DQLITE_NEXT`. This will avoid a big merge headache down the line.


cole-miller:
@just-now I think this is good to go, would appreciate you confirming that I didn't miss anything important when adding ifdefs.

Makefile.am
```diff
@@ -3,6 +3,10 @@ AM_CFLAGS += $(CODE_COVERAGE_CFLAGS)
 AM_CFLAGS += $(SQLITE_CFLAGS) $(UV_CFLAGS) $(PTHREAD_CFLAGS)
 AM_LDFLAGS = $(UV_LIBS) $(PTHREAD_LIBS)
 
+if DQLITE_NEXT_ENABLED
+AM_CFLAGS += -DDQLITE_NEXT
```
letFunny, line 7: [Nit]: Do we normally indent inside conditionals, I can see examples of either way below.

src/server.c
```diff
@@ -186,9 +188,11 @@ int dqlite__init(struct dqlite_node *d,
 err_after_raft_transport_init:
 	raftProxyClose(&d->raft_transport);
 err_after_pool_init:
+#ifdef DQLITE_NEXT
 	pool_close(&d->pool);
 	pool_fini(&d->pool);
 err_after_loop_init:
```
letFunny, line 194: I may be wrong here but I think the label that exists in `master` at the moment is `err_after_loop_init` so the if def should be:
```c
....
err_after_raft_transport_init:
	raftProxyClose(&d->raft_transport);
#ifdef DQLITE_NEXT
err_after_pool_init:
	pool_close(&d->pool);
	pool_fini(&d->pool);
#endif
err_after_loop_init:
	uv_loop_close(&d->loop);
....
```

letFunny (commented):
It is a little bit tricky to review but I tried to have `master` and this PR side by side to see if they had the same behavior when `DQLITE_NEXT` was not defined. Looks good to me apart from one comment I included, thanks!

Makefile.am
cole-miller: It seems that `if` needs to be at the start of a line, so we cannot have proper indendation for conditionals within conditionals. Given that I'd prefer to have no indentation at all; I'll harmonize everything on that style in a new commit. (Did I mention I don't like Autotools?)

src/server.c
cole-miller: That way of doing it doesn't quite work as-is because there is a `goto err_after_pool_init` in the error branch after `raftProxyInit` above. We would have to change that to

```
        rv = raftProxyInit(&d->raft_transport, &d->loop);
        if (rv != 0) {
#ifdef DQLITE_NEXT
                goto err_after_pool_init;
#else
                goto err_after_loop_init;
#endif
        }
```

The way I've written it looks kind of weird but gets away with only one `ifdef`, which I think is nice. Thoughts?

cole-miller:
Pushed https://github.com/canonical/dqlite/pull/642/commits/24a4879de5c3dbddc860818232bc77d1e4514c9f to codify that dqlite-next relies (will rely) on using the bundled raft sources and headers, since we will be making incompatible changes like #639.

src/leader.c
```diff
@@ -364,10 +364,15 @@ static void leaderExecV2(struct exec *req, enum pool_half half)
 	unsigned i;
 	int rv;
 
+#ifdef DQLITE_NEXT
```
just-now, line 367: Not insisting as it is a temporary solution, just wanted to say that there is a way to orgranise it in a bit more different way:
```
#ifdef DQLITE_NEXT
leaderExecV2(req, POOL_TOP_HALF);
leaderExecV2(req, POOL_TOP_BOTTOM);
#else
queue_pool_work(...)
#endif 
```

just-now (approved):
Looks good to me.

PR #645, opened 2024-04-25
Use system raft headers when not building raft

cole-miller:
This is a bug that I introduced during the raft import: the bundled raft.h (which concatenates raft's raft.h, raft/uv.h, and raft/fixture.h) is included everywhere even if we don't use the bundled raft sources. Fix this so we always use the headers that correspond to the raft code that will be called at runtime. This is necessary to support breaking changes to raft data structures and functions under DQLITE_NEXT.


just-now (approved):
+1

PR #641, opened 2024-04-23
Revert tolerance for EAGAIN when probing async I/O

cole-miller:
This reverts canonical/raft#485, hopefully (not yet confirmed) fixing the regression canonical/lxd#13326.

Signed-off-by: Cole Miller <cole.miller@canonical.com>

PR #647, opened 2024-04-26
github: Restore raft testing on other filesystems

cole-miller:
I neglected to set this up when doing the raft import, and was reminded because this additional testing should pick up the original issue that [this patch](https://github.com/canonical/raft/pull/485)---since reverted by #641---addressed.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -51,7 +51,7 @@ jobs:
       run: |
            ./test/raft/lib/fs.sh setup
            export $(test/raft/lib/fs.sh detect)
-           ./unit-test
+           ./unit-test --show-stderr
            ./integration-test
            ./raft-core-fuzzy-test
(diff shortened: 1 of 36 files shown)


letFunny:
Does this mean that the CI is no longer going to be green? If that is the case, do we plan to fix this tests?

cole-miller:
I don't plan to merge this with the tests failing, we need to fix them one way or another. Currently trying to figure out whether there's a proper fix that doesn't cause the same issue as canonical/raft#485.

marco6 (approved):
I will soon merge this PR. To make sure that these changes preserve the author, I will manually solve all the conflicts, considering the most of those already took place (like the dumb `#undef` at the end of the file.

I will also limit the test to all file systems except `XFS` as that has a problem with async io that will be solved in a separate PR (which will also re-enable testing it).

All in all, thanks for this @cole-miller .

PR #649, opened 2024-04-27
gateway: Clean up error codepaths

cole-miller:


cole-miller:
Closing---I don't think this is the right place to spend our effort on improved error reporting.

PR #651, opened 2024-05-02
configure: Disable default CFLAGS from AC_PROG_CC

cole-miller:
AC_PROG_CC adds some default compiler flags unless you tell it not to:

> If using the GNU C compiler, set shell variable GCC to ‘yes’. If output variable CFLAGS was not already set, set it to -g -O2 for the GNU C compiler (-O2 on systems where GCC does not accept -g), or -g for other compilers.

([Source](https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/C-Compiler.html).) 

Also tweak DEBUG_ENABLED handling in Makefile.am: we always pass `-g3` in the basic CFLAGS set from configure.ac, so no point adding it again there. Instead, pass `-O0`, which should be the default but here serves as a kind of documentation.

Credit to @letFunny for noticing that `-O2` was getting passed to gcc even for builds configured with `--enable-debug`.


cole-miller:
FYI @freeekanayaka

cole-miller:
I've confirmed locally that this results in the intended CFLAGS being placed into the generated Makefile.

PR #652, opened 2024-05-03
uv_writer: Don't skip later KAIO events when one is EAGAIN

cole-miller:
@marco6 pointed out this `return` that should be a `continue`, which causes us to skip processing some KAIO write results if we get several of them in a batch and one of them has status code EAGAIN.

In practice, when running our test suite, it seems that io_getevents never returns more than one event in this context, which together with the dependence on getting EAGAIN explains why we never noticed this while developing dqlite or raft. But in production we could certainly encounter `n_events > 1`.

### DIFF
--- a/src/raft/uv_writer.c
+++ b/src/raft/uv_writer.c
@@ -207,7 +207,7 @@ static void uvWriterPollCb(uv_poll_t *poller, int status, int events)
 				req->status = RAFT_IOERR;
 				goto finish;
 			}
-			return;
+			continue;
 		}
 
 		uvWriterReqSetStatus(req, (int)event->res);


src/raft/uv_writer.c
```diff
@@ -207,7 +207,7 @@ static void uvWriterPollCb(uv_poll_t *poller, int status, int events)
 				req->status = RAFT_IOERR;
 				goto finish;
 			}
-			return;
+			continue;
 		}
 
 		uvWriterReqSetStatus(req, (int)event->res);
```
letFunny, line 213: Why is it that we don't call `uvWriterReqSetStatus` in `uvWriterAfterWorkCb`? Maybe it is not needed but I am not sure.

letFunny (approved):
Thanks, looks good to me! I just have one question.

src/raft/uv_writer.c
cole-miller: It's done at the bottom of `uvWriterWorkCb`.

PR #654, opened 2024-05-21
github: Test on Ubuntu Noble

cole-miller:

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -53,7 +53,7 @@ jobs:
     - name: Coverage
       env:
         CC: ${{ matrix.compiler }}
-      if: ${{ matrix.compiler == 'gcc' }}
+      if: ${{ matrix.os == 'ubuntu-22.04' && matrix.compiler == 'gcc' }}
       run: |
         make code-coverage-capture
 
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -12,6 +12,7 @@ jobs:
         os:
           - ubuntu-20.04
           - ubuntu-22.04
+          - ubuntu-24.04
         compiler:
           - gcc
           - clang


Makefile.am
```diff
@@ -352,6 +352,10 @@ CODE_COVERAGE_OUTPUT_FILE=coverage.info
 CODE_COVERAGE_IGNORE_PATTERN="/usr/include/*"
 CODE_COVERAGE_BRANCH_COVERAGE=1
 CODE_COVERAGE_LCOV_OPTIONS=$(CODE_COVERAGE_LCOV_OPTIONS_DEFAULT) --rc lcov_excl_br_line="assert\("
+# FIXME this works around an incompatibility between lcov v2 and our copy of ax_code_coverage.m4.
+# Once GNU publishes an updated version of ax_code_coverage.m4 that works out-of-the box with
+# lcov v2, this line should be removed.
+CODE_COVERAGE_LCOV_OPTIONS += --ignore-errors unused
```
letFunny, line 358: I assume these are just warnings and the coverage is still reported correctly, right?

From what I see on the man pages,
```
unused: the include/exclude/erase/omit/substitute pattern did not match any file pathnames.
```
so I just want to check we are sure we are not missing any paths.

letFunny (commented):
Thanks!

Makefile.am
cole-miller: Yeah, this was a bad fix. I ended up (https://github.com/canonical/dqlite/pull/654/commits/2b2982d2a4b07f663de746ad0569188a8f121d2f) just changing CI to only collect coverage on Jammy, so we don't have to deal with the lcov v2 changes at all until GNU publishes an updated ax_code_coverage.m4. (I don't really want to carry our own modified copy, since that creates more work whenever we want to refresh it using changes from GNU.)

I do not think the error reported here is a problem for coverage correctness; it just means that our build doesn't use any files from `/tmp`, which makes sense. I guess that ax_code_coverage.m4 passes `--exclude /tmp/*` unconditionally, though I couldn't say confidently why.

There are also some warnings about using an option `--lcov-branch-coverage` that has been renamed in `v2` to just `--branch-coverage`; again, that's coming straight from our vendored ax_code_coverage.m4.

cole-miller:
Ready to go I think.

letFunny (approved):
Thanks!

PR #655, opened 2024-05-22
Snapshot installation DLD added, WIP

just-now:


src/raft/recv_install_snapshot.c
```diff
@@ -8,6 +8,348 @@
...
+ *
+ * 7. When the iteration has finished the Leader sends
+ * InstallShapshot(..., done=true) message to the Follower. It moves
+ * the Follower back to NORMAL state and the state machine
+ * corresponding to the Follower on the Leader is moved to
+ * SNAPSHOT_DONE_SENT state.
+ *
+ * 8. The Leader sends AppendEntries() RPC to the Follower and
+ * restarts the algorithm from (1). The Leader's state machine is
+ * being moved to FOLLOWER_ONLINE state.
```
letFunny, line 155: Would it be more accurate to say that the Leader's state machine is moved to FOLLOWER_ONLINE once the response `InstallSnapshotResult` is received. From the text written in 8. it looks like it waits in FOLLOWER_ONLINE and sends AppendEntries() there.

src/raft/recv_install_snapshot.c
```diff
@@ -8,6 +8,348 @@
...
+ *
+ * @TODO: We'd need to review other options here.
+ *
+ * The Follower maintains persistent bootcounter variable incremented
+ * every time during process start and sends it back to the Leader as
+ * a part of heart bit in AppendEntriesResult() message.
+ *
+ * Bootcounters are being used to restart snapshot installation
+ * procedures crashed in the middle of their execution on the Leaders'
+ * and Follower's sides.  When Leader notices that bootcounter changed
```
letFunny, line 192: How is the bootcounter communicated from follower to leader?

src/raft/recv_install_snapshot.c
```diff
@@ -8,6 +8,348 @@
...
+ *
+ * - If a party notices that last known chunk index sent back to it
+ *   doesn't match it's own, the communication get's restarted from
+ *   the lowest known index.
+ *
+ * ==Crashes of the Leader and Follower.
+ *
+ * @TODO: We'd need to review other options here.
+ *
+ * The Follower maintains persistent bootcounter variable incremented
```
letFunny, line 186: Instead of persisting it, would it be enough to create a random number on start-up and communicate it to the leader? (assuming we need to communicate the bootcounter to the leader through some mechanism). When the random number changes that means a reboot has happened.

src/raft/recv_install_snapshot.c
```diff
@@ -8,6 +8,348 @@
...
+
+struct page_from_to {
+	pageno_t from;
+	pageno_t to;
+};
+
+struct raft_install_snapshot_mv {
+	int version;
+
+	struct page_from_to *mv;
```
letFunny, line 69: Q: Why have a nested struct instead of inlining the fields? Asking out of pure curiosity.

letFunny (commented):
Just skimmed through it quickly, the comments and the explanation look great. It is very easy to follow, thanks for that!

src/raft/recv_install_snapshot.c
```diff
@@ -8,6 +8,348 @@
...
+ * To handle use-cases when any party of the communication becomes
+ * unavailable for a while without crash the following assumtions are
+ * made:
+ *
+ * - Signature() or InstallSnapshot(MV/CP) messages are idempotent and
+ *   can be applied to the persistent state many times resulting the
+ *   same transition.
+ *
+ * - Each message with data chuncks has an information about the
+ *   "chunk index". Chunk indexes come in monotonically increasing
```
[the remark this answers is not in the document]
just-now: I'll change "chunk" to "page". Still the passage you mentioned is a bit esoteric. 

Signature() and InstallSnapshot(CP()) RPCs send a subset of elements of a set of all "checksums" and all "pages" respectively. Subsets can be denoted as a range (from, to] + a data base name the range belongs to. So the chunk itself corresponds to a range + db_name + an array of elements of corresponding sets (array of checksums or pages).

Practical AIs:
- Extract better abstractions to send parts (chunks, pieces, subsets) of data representing pages or checksums.
- update Signature() and InstallSnapshot(CP()) RPCs w.r.t. this.

src/raft/recv_install_snapshot.c
```diff
@@ -8,6 +8,348 @@
...
+typedef unsigned long int pageno_t;
+
+struct page_checksum_t {
+	pageno_t   page_no;
+	checksum_t checksum;
+};
+
+struct raft_signature {
+	int version;
+
```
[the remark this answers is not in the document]
just-now: Will take care of it.

src/raft/recv_install_snapshot.c
letFunny: I meant having to separate fields in `raft_install_snapshot_mv` called `to` and `from`.

src/raft/recv_install_snapshot.c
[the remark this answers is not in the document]
letFunny: No, you are right, I was reading the comment wrong. We are dealing with "chucks" that, as you say, are identified using a range and the db name, so I think the terminology is correct.

just-now:
absorbed by https://github.com/canonical/dqlite/pull/657

just-now:
absorbed by https://github.com/canonical/dqlite/pull/657

PR #656, opened 2024-05-23
Add file locking to prevent two dqlite instances from using the same directory concurrently

cole-miller:
It was suggested that some of the corruption issues reported by LXD users might be due to two LXD daemon processes running concurrently, causing two dqlite instances to concurrently modify the common data directory. LXD already has some mutual exclusion logic to prevent two daemons from running at the same time:

https://github.com/canonical/lxd/blob/1514a400f11a82b90a5294b5c1f31cd2c6dd9311/lxd/endpoints/socket.go#L39

But even so it seems worth it to do some file locking to prevent this definitively on the dqlite side.

Note that this obviously won't work unless both the contending processes are running versions of dqlite that include the locking.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -3,6 +3,7 @@
 #include <errno.h>
 #include <sched.h>
 #include <stdlib.h>
+#include <sys/file.h>
 #include <sys/un.h>
 #include <time.h>
 #include <uv.h>
@@ -63,6 +64,7 @@ int dqlite__init(struct dqlite_node *d,
 	ssize_t count;
 
 	d->initialized = false;
+	d->lock_fd = -1;
 	memset(d->errmsg, 0, sizeof d->errmsg);
 
 	rv = snprintf(db_dir_path, sizeof db_dir_path, DATABASE_DIR_FMT, dir);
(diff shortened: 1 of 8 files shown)


cole-miller:
cc @tomponline

tomponline:
Thanks for this @cole-miller 

Once this is merged it'll be auto integrated on next build into LXD's latest/edge channel, and then tested through our CI as well as our daily snap CI (https://github.com/canonical/lxd-ci/actions/workflows/tests.yml?query=event%3Aschedule)

So we'll see if there are any breakages then.

Also thanks for checking if LXD checks for existing processes. I suppose it is possible potentially for LXD's unix socket to have been closed before the DB is closed and thus allow a new LXD process to start and open the DB concurrently  - ill double check this our side.

tomponline:
@cole-miller seems like LXD does correctly shutdown its local unix listener after closing the DB:

https://github.com/canonical/lxd/blob/main/lxd/daemon.go#L1899-L1917

src/server.c
```diff
@@ -829,6 +831,35 @@ static bool taskReady(struct dqlite_node *d)
...
+ * while dqlite is running (between dqlite_node_start and dqlite_node_stop),
+ * and also during recovery (dqlite_node_recover_ext). This prevents corruption
+ * if two instances of dqlite try to use the same directory concurrently. */
+#define LOCK_FILENAME "dqlite-lock"
+
+static int acquire_dir(const char *dir, int *fd_out)
+{
+	char path[PATH_MAX];
+	snprintf(path, sizeof(path), "%s/%s", dir, LOCK_FILENAME);
+	int fd = open(path, O_RDONLY|O_CREAT, S_IRUSR|S_IWUSR|S_IRGRP|S_IWGRP|S_IROTH);
```
letFunny, line 844: As discussed in our meeting I am not sure whether this approach vs open & unlink is better. Opening should work as well because only one process has to hold the lock, but I do not have a strong opinion.

letFunny (approved):
Thanks!

src/server.c
cole-miller: Discussed elsewhere---opening with `O_CREAT|O_EXCL` and then unlinking doesn't provide mutual exclusion (the second call to `open` creates a new inode, living alongside the old one which is no longer linked into the filesystem). And using `O_EXCL` without unlinking has the problem that if dqlite crashes the "lock" won't be released. So it seems like flock is the way to go here.

PR #657, opened 2024-06-04
snapshot installation DLD, latest version

letFunny:


src/raft/recv_install_snapshot.c
```diff
@@ -8,6 +8,350 @@
...
+ *
+ * ==Crashes of the Leader and Follower.
+ *
+ * Crashes of the Leader are handled by Raft when a new leader is elected
+ * and the snapshot process is restarted.
+ *
+ * If the Follower crashes, it will restart its state machine into the NORMAL
+ * state. If the Leader then sends a message which assumes the Follower is at
+ * the state prior to the crash, the Follower will reply using the message's
+ * result RPC using the failed=true flag. Upong receiving the message the
```
just-now, line 198: I feel like I'm being picky to words, still it's better not to feed the reader of this DLD by unfulfilled hopes that the algorithm can detect a failure of the counter part of the communication. 

I'd propose to rename (sorry, it's not about words) `failed` flag with `unexpected` having a semantics that the Leader re-started the communication because the Follower by any reason treats incoming messages come "in unexpected order" w.r.t. defined states and expectations of the Follower (today it's just restarted, but tomorrow the same mechanism is good to use elsewhere, let's not limit ourselves).

> If the Leader then sends a message which assumes the Follower is at the state prior to the crash

Also, I'd propose to phrase quoted part of the sentence in terms of "unexpected sequence of events".

just-now: Additional notes: 

Historically, my initial intent was to put a flag similar to unexpected into `AppendEntriesReply()` message so that it could be the only mechanism delivering such results. I'm not proposing to change anything in the currently provided mechanism (because I don't have a clear proposal by this time), still I'll be happy beyond any beliefs if we'll be able to generalize it.

src/raft/recv_install_snapshot.c
```diff
@@ -8,6 +8,350 @@
...
+ * |        |                        V
+ * |        +--------------- SNAPSHOT_DONE_SENT
+ * |                                 | InstallSnapshotResult() received
+ * +---------------------------------+
+ *
+ * Follower's state machine:
+ *
+ *                         +---------+
+ *                         |         | Signature/CP_Result/MV_Result received
+ *                         |         | *Result(failed=true) sent
```
just-now, line 245: I'd expect that the trigger you mentioned shall be somewhere around `@` (`@ || %`). Also, could you think about rephrasing it? Maybe the following passage will give you an idea:
 (%) - Signature/CP_Result/MV_Result received and in the current state receiving of such message type is unexpected. Result(failed=true) sent.

BTW, `(@)` message has semantics of unexpected event (Leader re-elected) also.

just-now (approved):
I'm good to approve this DLD. It has enough information to go forward to the CODE phase. For sure it also has a few blanks, for example, it's not yet defined the mechanism to resent messages when the counter part becomes unavailable, but IMHO this could be defined "on the go". Hopefully the last one will find it's reflection in the DLD.

src/raft/recv_install_snapshot.c
letFunny: I was assuming the `failed` (now `unexpected`) flag would only be used to detect crashes on the follower side, hence why it was only present in the `NORMAL` state but you are right the mechanism has to be more flexible.

I have changed the diagram in the latest commit.

src/raft/recv_install_snapshot.c
letFunny: I don't think you are being picky with words, the DLD is very important and will later be read in order to understand the code so it is better to spend more time and do it properly. In https://github.com/canonical/dqlite/pull/657/commits/6633bd83a2cf2c99805d1acbcbf9224d6485dd5d I have changed the `failed` flag to `unexpected` and tried to reword the text with your suggestions. I am not sure I addressed all of your concerns so if there is anything missing please tell me.

PR #659, opened 2024-06-10
initial snapshot state machine implementation

letFunny:


Makefile.am
```diff
@@ -219,27 +219,24 @@ libraft_la_CFLAGS = $(AM_CFLAGS)
...
 raft_core_unit_test_SOURCES = \
-  src/tracing.c \
-  src/raft/byte.c \
-  src/raft/compress.c \
-  src/raft/configuration.c \
-  src/raft/err.c \
-  src/raft/flags.c \
-  src/raft/heap.c \
-  src/raft/log.c \
+  $(basic_dqlite_sources) \
```
cole-miller, line 222: I'm a bit concerned about making the raft test artifacts link in arbitrary dqlite object code or even libdqlite itself---feels like we should maintain a separation here even if it requires compiling a subset of source files (like tracing.c) more than once.

Makefile.am
```diff
@@ -339,6 +337,9 @@ libsqlite3_la_SOURCES = sqlite3.c
 libsqlite3_la_CFLAGS = -g3
 
 unit_test_LDADD += libsqlite3.la
+if BUILD_RAFT_ENABLED
+raft_core_unit_test_LDADD += libsqlite3.la
```
cole-miller, line 341: I think tests that use libsqlite3 should be in the dqlite part of the tree.

src/raft.h
```diff
@@ -328,6 +329,25 @@ struct raft_append_entries_result
 };
 #define RAFT_APPEND_ENTRIES_RESULT_VERSION 1
 
+typedef unsigned int checksum_t;
+typedef unsigned long int pageno_t;
```
cole-miller, line 333: I would prefer fixed-size integers here, also `pageno_t` should be only 32 bits (that this is enough is fixed by the SQLite file formats).

src/raft.h
```diff
@@ -340,9 +360,73 @@ struct raft_install_snapshot
...
+	enum raft_result result;
+};
+#define RAFT_INSTALL_SNAPSHOT_RESULT_VERSION 0
+
+struct raft_signature {
+	int version;
+
+	const char *db;
+	struct page_from_to page_from_to;
+	unsigned int cs_page_no;
```
cole-miller, line 379: pageno_t?

src/raft/recv_install_snapshot.c
```diff
@@ -9,6 +11,9 @@
 #include "replication.h"
 
 #include "../lib/sm.h"
+#include "src/raft.h"
+#include "src/raft/recv_install_snapshot.h"
+#include "src/raft/uv_os.h"
```
cole-miller, line 16: I think the relative paths here are wrong, they should be `../raft.h` etc. (Arguably we should switch to just passing `-Isrc` into the build so we can dispense with all the `..` stuff but for now that's where we're at.)

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+static int insert_checksums(struct raft_io_async_work *req)
 {
-	(void) leader;
-	(void) msg;
-	(void) leader_states;
-	//switch (sm_state(leader)) {
-	//}
+	int rv;
+
+	struct insert_checksum_data *data = (struct insert_checksum_data *)req->data;
```
cole-miller, line 330: You should be able to drop the explicit cast here and thus avoid stuttering

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+	sqlite3_stmt *stmt = data->state->ht_stmt;
+	for (unsigned int i = 0; i < data->cs_nr; i++) {
```
cole-miller, line 332: `unsigned` not `unsigned int`

src/raft/recv_install_snapshot.c
cole-miller, line 330: You should be able to drop the explicit cast here.

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+static int create_ht_and_stmt(struct raft_io_async_work *req)
+{
+	char db_filename[30];
+	// TODO tracef with err_msg and tracef in general.
+	char *err_msg;
+	int rv;
+
+	struct create_ht_data *data = (struct create_ht_data *)req->data;
+	struct snapshot_leader_state *state = data->state;
+	sprintf(db_filename, "ht-%lld", state->follower_id);
```
cole-miller, line 429: `snprintf`

src/raft.h
```diff
@@ -328,6 +329,25 @@ struct raft_append_entries_result
...
+
+struct page_from_to {
+	pageno_t from;
+	pageno_t to;
+};
+
+enum raft_result {
+	OK = 0,
+	UNEXPECTED = 1,
+	DONE = 2,
```
cole-miller, line 348: I think these names should be prefixed (`RAFT_RESULT_OK` etc.)

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+	};
+	state->io->async_work(work, calculate_local_checksums);
+}
+
+void leader_tick(struct sm *leader, const struct raft_message *msg)
+{
+	(void)leader_states;
+
+	assert(leader != NULL);
+	assert(msg != NULL);
```
cole-miller, line 654: `PRE`?

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+			return;
+		}
+		async_send_mv_or_cp(state);
+	}
+}
+
+__attribute__((unused)) static bool leader_invariant(const struct sm *sm,
+						     int prev_state)
+{
+	bool rv;
```
cole-miller, line 723: I would prefer that `rv` always be an `int` (that is, to use a different name here).

src/raft/recv_install_snapshot.h
```diff
@@ -3,8 +3,47 @@
...
+struct snapshot_leader_io {
+	bool (*log_index_found)(raft_index index);
+	int (*async_work)(struct raft_io_async_work *req,
+			  raft_io_async_work_cb cb);
+	void (*async_send_message)(struct raft_io_send *req,
+			struct raft_message *msg,
+			raft_io_send_cb cb);
+	int (*read_page)(uint64_t offset,
+			int64_t page_size,
+			char *page);
```
cole-miller, line 22: This will have to be async

src/raft/recv_install_snapshot.h
```diff
@@ -3,8 +3,47 @@
...
+};
+
+struct snapshot_leader_state {
+	struct sm sm;
+	raft_id follower_id;
+	sqlite3 *ht;
+	sqlite3_stmt *ht_stmt;
+	struct snapshot_leader_io *io;
+	unsigned long last_page;
+	unsigned long last_page_acked;
```
cole-miller, line 33: pageno_t?

src/raft/recv_install_snapshot.h
```diff
@@ -3,8 +3,47 @@
...
+			int64_t page_size,
+			char *page);
+	// read_chunk
+};
+
+struct snapshot_leader_state {
+	struct sm sm;
+	raft_id follower_id;
+	sqlite3 *ht;
+	sqlite3_stmt *ht_stmt;
```
cole-miller, line 30: Generally I think we want to have all mentions of sqlite3 types and functions on the dqlite side of the boundary.

src/raft/recv_install_snapshot.h
```diff
@@ -3,8 +3,47 @@
...
 #define RECV_INSTALL_SNAPSHOT_H_
 
+#include <sqlite3.h>
+
 #include "../raft.h"
 
+// Forward declaration.
+struct snapshot_leader_state;
+
+struct snapshot_leader_io {
```
cole-miller, line 13: Probably we will end up putting these in the fsm vtable or the io vtable or some mix of those.

Makefile.am
letFunny: The problem is that we are going to be using code from dqlite in Raft soon. We need to read pages, check the wal, etc. Right now I needed for the state machines but in the future I thought that we were going to combine both Raft and dqlite.

Makefile.am
letFunny: We are using sqlite to create the hash-table, it does not have anything to do with dqlite per se. The raft code itself (code for creating the ht) has a dependency on SQLite, not only the tests. We could abstract it away to maintain Raft more agnostic but I am not sure there is a point any more in the extra complexity.

src/raft/recv_install_snapshot.h
letFunny: Agree, it is convenient for now to have them here for the tests.

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+bool is_main_thread(void)
+{
+	// TODO: thread local storage.
 	return true;
 }
 
+void send_snapshot_cb(struct raft_io_send *req, int status) {
+	(void)req;
+	(void)status;
+	struct snapshot_leader_state *state = (struct snapshot_leader_state *) req->data;
```
just-now: not sure if it is a good moment for this (but it's always not late to step on the path of samurai).

It looks more maintainable and better from perspective of code analysis (there's no guarantee that req->data lives the same live as the structure containing it) just to embed `struct raft_io_send` into `struct snapshot_leader_state` and use container_of() to obtain `struct snapshot_leader_state` back in the function.

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+	struct snapshot_leader_state *state =
+		CONTAINER_OF(leader, struct snapshot_leader_state, sm);
+	struct snapshot_leader_io *io = state->io;
+	assert(state != NULL && io != NULL);
+
+	PRE(is_main_thread());
+	PRE(msg->server_id == state->follower_id);
+	// TODO: timeouts.
+	int leader_state = sm_state(leader);
+	if (leader_state == LS_FOLLOWER_ONLINE) {
```
just-now, line 665: Suggestion: Case construction looks more idiomatic and natural here.

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+{
+	bool rv;
+	(void)prev_state;
+	// TODO: if we need msg pointer it is better to store it in the
+	// state thanto pass it.
+
+	struct snapshot_leader_state *state =
+		CONTAINER_OF(sm, struct snapshot_leader_state, sm);
+
+	if (sm_state(sm) == LS_SIGNATURES_CALC_STARTED ||
```
just-now, line 731: if it's more useful, you can use `>=` for states here.

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+	sm_move(&state->sm, LS_SNAPSHOT_CHUNCK_SENT);
+	raft_free(req);
+}
+
+void async_send_mv_or_cp(struct snapshot_leader_state *state) {
+	(void)state;
+	struct raft_message *reply;
+	struct raft_io_send *reply_req;
+
+	reply = raft_malloc(sizeof(*reply));
```
just-now, line 606: Question:
What if we use macro declarations like these for allocations?
```
#define ALLOC_ARR(arr, nr)  calloc((nr), sizeof ((arr)[0]))
#define ALLOC_PTR(ptr)      ALLOC_ARR(ptr, 1)
```
cc @cole-miller

just-now:
```
ALLOC_PTR(reply);
```

src/raft/recv_install_snapshot.c
letFunny: Wouldn't that be a problem in the context of scheduling async callbacks? We could have several `raft_io_send` when processing several messages. Or should we control that and not process more than 1 message?

src/raft/recv_install_snapshot.c
letFunny: I was using a switch statement and had some problems when omitting one `break;` and also when creating variables as `case` does not create a scope.

src/raft/recv_install_snapshot.c
cole-miller: One issue with is is that we have a few different allocators that are used in different parts of the code, so until we clean that up (which I want to do) we'd need to duplicate these macros for each allocator.

src/raft/recv_install_snapshot.c
```diff
@@ -328,34 +279,482 @@ static const struct sm_conf leader_states[LS_NR] = {
...
+	reply_req = raft_malloc(sizeof(*reply_req));
+	assert(reply_req != NULL);
+	reply_req->data = state;
+
+	// TODO: construct proper message.
+	reply->type = RAFT_IO_INSTALL_SNAPSHOT_CP;
+	state->io->async_send_message(reply_req, reply, send_mv_or_cp_cb);
+}
+
+void calculate_local_checksums(struct raft_io_async_work *req, int status) {
```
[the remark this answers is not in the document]
cole-miller: And a matching precondition for callbacks that shouldn't run on the main thread?

just-now (approved):
Good work!

src/raft/recv_install_snapshot.h
```diff
@@ -3,8 +3,308 @@
...
+	work_op work_cb;
+	work_op after_cb;
+	struct sm sm;
+};
+
+struct sender {
+	sender_cb_op cb;
+};
+
+struct timeout {
```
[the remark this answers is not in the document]
letFunny: Agree, I was planning to introduce that gradually in the next patches.

src/raft/recv_install_snapshot.h
```diff
@@ -3,8 +3,308 @@
...
+
+	FS_HT_CREATE,
+	FS_HT_WAIT,
+
+	FS_SIGS_CALC_STARTED,
+	FS_SIGS_CALC_LOOP,
+	FS_SIGS_CALC_MSG_RECEIVED,
+	FS_SIGS_CALC_DONE,
+
+	FS_SIG_RECEIVING,
```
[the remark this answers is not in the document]
letFunny: In the beginning the names were uniform but I wanted to convey somehow that the leader has a loop that involves sending messages and waiting for replies or timeouts, but the follower waits indefinitely for messages, the loop so to speak has different semantics.

I agree that the names should be more uniform in next patches.

cole-miller (approved):
Thanks, this is great!

PR #660, opened 2024-06-17
Added a very simple install stanpshot sm protocol impl for the leader

just-now:


letFunny (approved):
Thanks, it is much clearer now! Let's merge it and continue iterating to integrate it.

just-now:
Let's use the prototype as a reference. From the logistics it's easier just to copy it by pieces to existing code.

PR #662, opened 2024-06-25
DLD of in-place checkpoints

just-now:


just-now:
I forgot to post this DLD which has been already discussed.

src/fsm.c
```diff
@@ -82,6 +82,102 @@ static int databaseReadUnlock(struct db *db)
...
+ *
+ * Effects of cp_create and cp_write are serialized because of the
+ * strucure of thread pool and per database thread affinity.
+ *
+ * Definition: Checkpoint itself is a state of the sqlite3 databases on
+ * the disk synced to a certain raft index.
+ *
+ * To create a checkpoint it's required to finish all tx_read and
+ * cp_read operations, initiate a set cp_create operations and wait
+ * untill it's completion without accepting new tx_reads and cp_reads.
```
cole-miller, line 162: We don't have to wait until *all* the `cp_create` operations have finished to accept new `tx_read` and `cp_read` on some specific database, right? Once the checkpoint completes (synchronously) on one database we can resume working with that database normally even if other databases are still being checkpointed.

I was also going to write here that it's okay to accept *new* transactions after setting a barrier but before the "draining" is complete, because these transactions will set their read marks at the end of the WAL so they can't interfere with our checkpoint. But that's not actually correct because new transactions can piggyback on an existing read mark.

cole-miller (changes requested):
Thank you!

src/fsm.c
```diff
@@ -82,6 +82,102 @@ static int databaseReadUnlock(struct db *db)
...
+ * can't intersect with cp_read and tx_read. This means that there
+ * will not be readmarks in WAL touched by the "readers".
+ *
+ * Still cp_create operation can intersect with tx_write opeations.
+ * tx_write operations can copy new pages and move mxFrame "pointer"
+ * in WAL.  cp_create operation can set readmarks and move nBackfill
+ * "pointer" to mark the area of the WAL which will be copied to the
+ * database state during sqlite3_wal_checkpoint() call.
+ *
+ * Effects of cp_create and cp_write are serialized because of the
```
letFunny, line 154: I think here it should be `tx_write` instead of `cp_write`?

src/fsm.c
```diff
@@ -82,6 +82,102 @@ static int databaseReadUnlock(struct db *db)
...
+ * tx_write:   w--------w-------f---w---f-------f
+ *             |-- db read period --|  |-- db write period --|
+ *             RRRRRRRRRRRQRQRQRRQQQQ   CCCCCC
+ * tx_read(R): s-s-s-f-s-f-f-f
+ * cp_create(C):                        c-c-...
+ * cp_read(Q):            r-r-r-f-f-f  /
+ *                                    /
+ * barrier:                   bbbbbbbbb
+ *
+ * (f) - correspoding operation completed.
```
letFunny, line 138: I am having a bit of trouble figuring out what the letters in the diagram represent (`s`/`f`/`r`/...), can we add them to the legend below or add a very brief explanation of how I should be reading them?

letFunny (commented):
Thanks for this DLD! It is almost ready, I agree with Cole's suggestion that makes the text a little bit more readable and we also need to follow on his other comment on multi-step vs single-step transactions. On my side, I only added two comments where I thought the text could be a little bit more clear.

PR #670, opened 2024-07-09
feat: proper timeout handling for snapshots

letFunny:
Added both unit tests and tests that use libuv timers directly

## Test with uv_timer_t
uv_timer_t

## Test with mock implementation of timers
ut

### DIFF
--- a/src/raft/recv_install_snapshot.c
+++ b/src/raft/recv_install_snapshot.c
@@ -346,21 +346,21 @@ enum to_state {
 static const struct sm_conf to_sm_conf[TO_NR] = {
 	[TO_INIT] = {
 		.flags   = SM_INITIAL | SM_FINAL,
-		.name    = "to_init",
+		.name    = "init",
 		.allowed = BITS(TO_STARTED),
 	},
 	[TO_STARTED] = {
 		.flags   = SM_FINAL,
-		.name    = "to_started",
+		.name    = "started",
 		.allowed = BITS(TO_EXPIRED) | BITS(TO_CANCELED),
 	},
 	[TO_EXPIRED] = {
 		.flags   = SM_FINAL,
-		.name    = "to_expired",
+		.name    = "expired",
 	},
 	[TO_CANCELED] = {
 		.flags   = SM_FINAL,
-		.name    = "to_canceled",
+		.name    = "canceled",
 	},
 };
 /* clang-format on */
@@ -454,7 +454,7 @@ static void leader_to_start(struct leader *leader,
 	sm_move(&to->sm, TO_STARTED);
 }
 
-static void leader_to_fini(struct leader *leader, struct timeout *to)
+static void leader_to_cancel(struct leader *leader, struct timeout *to)
 {
 	leader->ops->to_stop(to);
 	sm_move(&to->sm, TO_CANCELED);
@@ -740,7 +740,7 @@ again:
 			leader_to_start(leader, &leader->rpc.timeout, 10000, rpc_to_cb);
 			return;
 		case RPC_REPLIED:
-			leader_to_fini(leader, &leader->rpc.timeout);
+			leader_to_cancel(leader, &leader->rpc.timeout);
 			rpc_fini(&leader->rpc);
 			sm_move(sm, leader_next_state(sm));
 			goto again;
--- a/src/raft/recv_install_snapshot.c
+++ b/src/raft/recv_install_snapshot.c
@@ -279,10 +279,9 @@ static const struct sm_conf rpc_sm_conf[RPC_NR] = {
 		         | BITS(RPC_ERROR),
 	},
 	[RPC_FILLED] = {
-		.flags   = SM_INITIAL | SM_FINAL,
 		.name    = "filled",
 		.allowed = BITS(RPC_SENT)
-		        | BITS(RPC_ERROR),
+		         | BITS(RPC_ERROR),
 	},
 	[RPC_SENT] = {
 		.name    = "sent",
@@ -795,10 +794,10 @@ again:
 		}
 		rpc_fill_follower(follower);
 		rc = rpc_send(&follower->rpc, ops->sender_send, follower_sent_cb);
-		if (rc == 0) {
-			break;
+		if (rc != 0) {
+			goto again;
 		}
-		goto again;
+		break;
 	case FS_SIG_PROCESSED:
 	case FS_CHUNCK_PROCESSED:
 	case FS_CHUNCK_REPLIED:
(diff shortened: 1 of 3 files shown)


src/raft/recv_install_snapshot.c
```diff
@@ -262,6 +261,7 @@
 
 enum rpc_state {
 	RPC_INIT,
+	RPC_FILLED,
```
letFunny, line 264: [Question for reviewer]: I introduced this new state because I wanted to differentiate the state where the rpc exists and the state where it is ready to be sent, meaning it is filled (we have called `rpc_fill`). I tried to make it explicit to avoid handling it as a special case. Also it is needed in order to be able to react to `M_MSG_SENT` (see my other comment).

src/raft/recv_install_snapshot.c
```diff
@@ -462,16 +484,14 @@ static void follower_sent_cb(struct sender *s, int rc)
...
-	sm_move(&rpc->sm, RPC_SENT);
 	follower_tick(follower, M_MSG_SENT);
 }
 
 static bool is_a_trigger_leader(const struct leader *leader, const struct raft_message *incoming)
 {
 	(void)leader;
-	/* Leader has no triggers on sending a message, it reacts when follower
-	 * replies */
-	return incoming != M_MSG_SENT;
```
letFunny, line 474: [Question for reviewer]: This was the other change that I am not sure of. I tried to make the way we handle messages more uniform. We react to `M_TIMEOUT` and `M_WORK_DONE` but we didn't use `M_MSG_SENT` we used to update the state of the rpc in the callback only in that case (from SENT to REPLIED we did it in the main loop). With this change we react to `M_MSG_SENT` and update the state in `rcp_tick`.

This also has the side-effect of being able to create the timeout transparently in the main loop.

test/raft/unit/test_snapshot.c
```diff
@@ -284,3 +297,189 @@ TEST(snapshot, leader, set_up, tear_down, 0, NULL) {
...
+
+static void pool_rpc_to_expired(struct rpc *rpc)
+{
+	uv_timer_start(&rpc->timeout.handle, rpc->timeout.cb, 0, 0);
+	progress();
+}
+
+/* TODO(alberto): combine them with tests above once the rest is in place.
+ * Dispatch to one of two implementations ut or pool in general functions. */
+TEST(snapshot_leader, pool_timeouts, set_up, tear_down, 0, NULL) {
```
letFunny, line 412: [Comment for reviewer]: The idea is to keep updating this test with the pool and the rest of the implementation. I plan to keep the unit test above around as it is a very simple way of debugging the logic of the state machine.

letFunny (commented):
Some questions about bits I am unsure of.

src/raft/recv_install_snapshot.c
```diff
@@ -270,18 +270,26 @@ enum rpc_state {
...
 static const struct sm_conf rpc_sm_conf[RPC_NR] = {
 	[RPC_INIT] = {
 		.flags   = SM_INITIAL | SM_FINAL,
 		.name    = "init",
-		.allowed = BITS(RPC_SENT) | BITS(RPC_ERROR),
+		.allowed = BITS(RPC_FILLED)
+		         | BITS(RPC_ERROR),
+	},
+	[RPC_FILLED] = {
+		.flags   = SM_INITIAL | SM_FINAL,
```
cole-miller, line 282: Why these flags?

cole-miller (changes requested):
Thanks, only minor comments!

src/raft/recv_install_snapshot.c
cole-miller: This change makes sense to me but do we still need `is_a_trigger_leader` afterward, other than for symmetry with `is_a_trigger_follower`?

src/raft/recv_install_snapshot.c
cole-miller: I like the additional state.

src/raft/recv_install_snapshot.c
letFunny: Not in this patch but the idea is that we will filter the messages to only get the ones we care about in that stage. That will come in a later patch when we do deduplication, react to unexpected messages, etc. So for now it is pending implementation.

src/raft/recv_install_snapshot.c
letFunny: Missed that

src/raft/recv_install_snapshot.c
```diff
@@ -270,18 +270,26 @@ enum rpc_state {
 	RPC_NR,
 };
 
+/* clang-format off */
 static const struct sm_conf rpc_sm_conf[RPC_NR] = {
 	[RPC_INIT] = {
 		.flags   = SM_INITIAL | SM_FINAL,
 		.name    = "init",
```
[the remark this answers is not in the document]
letFunny: We do have a notion of name for a state machine so I think prefixing the states with the name is redundant. Thoughts @just-now ?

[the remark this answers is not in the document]
just-now: Just a matter of style. I'm OK with both.

src/raft/recv_install_snapshot.c
```diff
@@ -413,30 +424,42 @@ static void follower_work_done(struct work *w)
...
-	to_init(to);
-	leader->ops->to_start(&leader->timeout, 10000, to_cb);
+	leader->ops->to_init(to);
+	sm_init(&to->sm, to_sm_invariant, NULL, to_sm_conf, "to", TO_INIT);
+	leader->ops->to_start(to, delay, to_cb);
 	sm_relate(&leader->sm, &to->sm);
 	sm_move(&to->sm, TO_STARTED);
 }
 
+static void leader_to_fini(struct leader *leader, struct timeout *to)
```
just-now, line 457: Q: Is it better to put a pointer to `v->ops()` into `timeout`-struct itself?

just-now: Q: is it better to rename `leader_to_fini()` to `*_cancel()`?

just-now (commented):
Good patch. Easy to read.

src/raft/recv_install_snapshot.c
letFunny: Agree with the second one. For the first, I am not sure I know what you mean:
* If what you mean is to include `leader->ops` inside the timeout struct, then I hesitant to do it because then the timeout will have primitives like `send_message`, `work_queue`, etc.
* If what you mean is to create another struct of operations for timeouts (`start`, `init`, `stop`) and "pass" the functions from `leader->ops` on creation then that will make timeout more general and allow us to reuse it somewhere else.

just-now (approved):

src/raft/recv_install_snapshot.c
letFunny: As discussed offline, the comment was about option 1 above. It had some challenges that required more code making the benefits less apparent. We agreed to do the changes in future iterations when we need them.

cole-miller (approved):

PR #671, opened 2024-07-11
Remove DQLITE_NEXT guard and unify with disk mode

cole-miller:
This PR removes the DQLITE_NEXT build configuration. Code for the new disk mode that was previously guarded by `#ifdef DQLITE_NEXT` can instead be selected at run time by calling the new `dqlite_node_create_v2` function. At the raft level, this is supported by making the "format version" number a dynamic value instead of a compile-time constant and plumbing it around.

Standalone version of the first part of #661, with the relevant review comments addressed (thanks @letFunny).


include/dqlite.h
```diff
@@ -399,16 +413,6 @@ DQLITE_API int dqlite_node_set_snapshot_params(dqlite_node *n,
...
 
-/**
- * WARNING: This is an experimental API.
- *
- * By default dqlite holds the SQLite database file and WAL in memory. By
- * enabling disk-mode, dqlite will hold the SQLite database file on-disk while
- * keeping the WAL in memory. Has to be called after `dqlite_node_create` and
- * before `dqlite_node_start`.
- */
-DQLITE_API int dqlite_node_enable_disk_mode(dqlite_node *n);
```
letFunny, line 410: Why are we changing the public API? Not that we should not, I am just wondering what are the reasons for defining `dqlite_node_create_v2` now when the feature is not yet stable.

src/raft.h
```diff
@@ -1398,6 +1398,28 @@ RAFT_API void raft_uv_set_tracer(struct raft_io *io,
...
+
+/**
+ * Select which disk format version should be used.
+ *
+ * If this is called after the `init` callback from the raft_io vtable,
+ * and some metadata files were found on disk, the @version is compared
+ * to the format version found in the metadata files, and an error is
+ * returned if they don't match.
+ *
+ * RAFT_UV_FORMAT_V1 is used by default.
```
letFunny, line 1419: Amazing documentation :+1:

src/raft/uv.c
```diff
@@ -112,6 +113,12 @@ static int uvInit(struct raft_io *io, raft_id id, const char *address)
 	if (rv != 0) {
 		return rv;
 	}
+	md_format = metadata.format_version;
```
letFunny, line 116: [Very Nit]: If the variable is only going to live for ~5 lines then I would prefer to use `metadata.format_version` directly. If we move code in the future we do not have to worry about two variables.

src/raft/uv.h
```diff
@@ -47,6 +47,7 @@ typedef unsigned long long uvCounter;
 /* Information persisted in a single metadata file. */
 struct uvMetadata
 {
+	uint64_t format_version;
```
letFunny, line 50: What about naming the enum and using the type here instead of the generic uint64_t?

src/raft/uv_segment.c
```diff
@@ -729,13 +729,14 @@ int uvSegmentBufferFormat(struct uvSegmentBuffer *b)
 	}
 	b->n = n;
 	cursor = b->arena.base;
-	bytePut64(&cursor, UV__DISK_FORMAT);
+	bytePut64(&cursor, format_version);
```
letFunny, line 732: Should we carry the format version in the segment itself?

src/raft/uv_encoding.c
```diff
@@ -143,7 +142,9 @@ static void encodeAppendEntries(const struct raft_append_entries *p, void *buf)
 	bytePut64(&cursor, p->prev_log_term);  /* Previous term. */
 	bytePut64(&cursor, p->leader_commit);  /* Commit index. */
 
-	uvEncodeBatchHeader(p->entries, p->n_entries, cursor, false /* no local data */);
+	/* Note: use RAFT_UV_FORMAT_V1 unconditionally because local data doesn't
+	 * appear in the AppendEntries message. */
```
letFunny, line 146: Nice :+1:

src/raft/uv_encoding.c
```diff
@@ -579,10 +580,13 @@ int uvDecodeEntriesBatch(uint8_t *batch,
 			 size_t offset,
 			 struct raft_entry *entries,
 			 unsigned n,
-			 uint64_t local_data_size)
+			 uint64_t local_data_size,
+			 uint64_t format_version)
 {
 	uint8_t *cursor;
 
+	PRE(ERGO(format_version == 1, local_data_size == 0));
```
letFunny, line 588: You can use `RAFT_UV_FORMAT_V1` instead of 1 here.

src/raft.h
```diff
@@ -1398,6 +1398,28 @@ RAFT_API void raft_uv_set_tracer(struct raft_io *io,
...
 RAFT_API void raft_uv_set_auto_recovery(struct raft_io *io, bool flag);
 
+/**
+ * Values for the format version number that describes the segment and snapshot
+ * files.
+ */
+enum {
+	RAFT_UV_FORMAT_V1 = 1, /* Original */
+	RAFT_UV_FORMAT_V2,     /* Added local_data fields */
+	UV_FORMAT_NR
```
letFunny, line 1408: Maybe `RAFT_UV_FORMAT_NR`? for consistency. Or is there are reason to have a distinct name?

src/raft/uv_metadata.c
```diff
@@ -26,12 +26,7 @@ static int uvMetadataDecode(const void *buf,
 			    char *errmsg)
 {
 	const void *cursor = buf;
-	uint64_t format;
-	format = byteGet64(&cursor);
-	if (format != UV__DISK_FORMAT) {
-		ErrMsgPrintf(errmsg, "bad format version %ju", format);
-		return RAFT_MALFORMED;
-	}
+	metadata->format_version = byteGet64(&cursor);
```
letFunny, line 29: Do we need to validate here that `0 < format_version < UV_FORMAT_NR`?

src/raft/uv_metadata.c
```diff
@@ -146,6 +143,7 @@ int uvMetadataLoad(const char *dir, struct uvMetadata *metadata, char *errmsg)
 	/* Check the versions. */
 	if (metadata1.version == 0 && metadata2.version == 0) {
 		/* Neither metadata file exists: have a brand new server. */
+		metadata->format_version = 0;
```
letFunny, line 146: [Nit]: If we want to set all fields to zero we could do `metadata = {0}` and not have to change this code when we add more fields.

src/raft/uv_metadata.c
```diff
@@ -182,7 +180,7 @@ int uvMetadataStore(struct uv *uv, const struct uvMetadata *metadata)
 	unsigned short n;
 	int rv;
 
-	assert(metadata->version > 0);
+	PRE(1 <= metadata->format_version && metadata->format_version < 3);
```
letFunny, line 183: I think it is better if we use the constants here instead of 1 and 3 :P

src/raft/uv_segment.c
```diff
@@ -748,12 +749,12 @@ int uvSegmentBufferAppend(struct uvSegmentBuffer *b,
...
 	size = sizeof(uint32_t) * 2;            /* CRC checksums */
-	size += uvSizeofBatchHeader(n_entries, true); /* Batch header */
+	size += uvSizeofBatchHeader(n_entries, format_version); /* Batch header */
 	for (i = 0; i < n_entries; i++) {       /* Entries data */
 		size += bytePad64(entries[i].buf.len);
-#ifdef DQLITE_NEXT
-		size += sizeof(struct raft_entry_local_data);
-#endif
+		if (format_version == RAFT_UV_FORMAT_V2) {
+			size += sizeof(struct raft_entry_local_data);
```
letFunny, line 756: Why is it that we use `sizeof(struct raft_entry_local_data)` here and in `src/raft/uv_encoding.c` we hardcode 8?

src/server.c
```diff
@@ -80,7 +81,11 @@ int dqlite__init(struct dqlite_node *d,
 			 "config__init(rv:%d)", rv);
 		goto err;
 	}
+	d->config.disk = disk_mode;
 	rv = VfsInit(&d->vfs, d->config.name);
```
letFunny, line 85: This line is not used if disk_mode is set so we might guard it in the conditional below.

src/server.c
```diff
@@ -80,7 +81,11 @@ int dqlite__init(struct dqlite_node *d,
...
+	if (disk_mode) {
+		dqlite_vfs_enable_disk(&d->vfs);
```
letFunny, line 87: We are not checking the return value of the function here.

src/server.c
```diff
@@ -80,7 +81,11 @@ int dqlite__init(struct dqlite_node *d,
...
+	}
 	sqlite3_vfs_register(&d->vfs, 0);
```
letFunny, line 89: We are also ignoring the return code here.

src/server.c
```diff
@@ -443,30 +470,6 @@ int dqlite_node_set_block_size(dqlite_node *n, size_t size)
 	raft_uv_set_block_size(&n->raft_io, size);
 	return 0;
 }
-int dqlite_node_enable_disk_mode(dqlite_node *n)
```
letFunny, line 446: Nice to see this go!

test/lib/server.c
```diff
@@ -62,7 +62,13 @@ void test_server_start(struct test_server *s, const MunitParameter params[])
 {
 	int rv;
 
-	rv = dqlite_node_create(s->id, s->address, s->dir, &s->dqlite);
+	bool disk_mode = false;
+	const char *disk_mode_param = munit_parameters_get(params, "disk_mode");
+	if (disk_mode_param != NULL) {
+		disk_mode = (bool)atoi(disk_mode_param);
+	}
```
letFunny, line 69: [Nit]: What about?
```c
const char *disk_mode_param = munit_parameters_get(params, "disk_mode");
int disk_mode = disk_mode_param != NULL ? (bool)atoi(disk_mode_param) : false;
```

test/raft/integration/test_uv_load.c
```diff
@@ -575,14 +585,15 @@ TEST(load, openSegmentWithIncompleteBatch, setUp, tearDown, 0, NULL)
...
 {
     struct fixture *f = data;
+
     uint8_t buf[5 * WORD_SIZE] = {
-        UV__DISK_FORMAT, 0, 0, 0, 0, 0, 0, 0, /* Format version */
+        (uint8_t)f->format_version, 0, 0, 0, 0, 0, 0, 0, /* Format version */
         0, 0, 0, 0, 0, 0, 0, 0, /* CRC32 checksums */
         0, 0, 0, 0, 0, 0, 0, 0, /* Number of entries */
-	0, 0, 0, 0, 0, 0, 0, 0, /* Local data size */
+        0, 0, 0, 0, 0, 0, 0, 0, /* Local data size */
```
letFunny, line 596: I think I am misunderstanding something, how can we have local data here where one of the `format_params` says no local data?

test/raft/integration/test_uv_load.c
```diff
@@ -1632,21 +1643,25 @@ TEST(load, openSegmentWithIncompleteBatchHeader, setUp, tearDown, 0, NULL)
...
-	    "load open segment open-1: entries batch 1 starting at byte 8: "
-	    "read header: short read: 8 bytes instead of 24";
-#else
-    const char *msg =
-	    "load open segment open-1: entries batch 1 starting at byte 8: "
-	    "read header: short read: 8 bytes instead of 16";
-#endif
+    const char *msg;
+    switch (f->format_version) {
+    case 1:
```
letFunny, line 1648: Constants :P

test/raft/integration/test_uv_load.c
```diff
@@ -1656,23 +1671,27 @@ TEST(load, openSegmentWithIncompleteBatchData, setUp, tearDown, 0, NULL)
...
-	    "load open segment open-1: entries batch 1 starting at byte 8: "
-	    "read data: short read: 4 bytes instead of 24";
-#else
-    const char *msg =
-	    "load open segment open-1: entries batch 1 starting at byte 8: "
-	    "read data: short read: 4 bytes instead of 8";
-#endif
+    const char *msg;
+    switch (f->format_version) {
+    case 1:
```
letFunny, line 1684: Constants :P

letFunny (commented):
Looking very good Cole! Thanks for creating a different PR for these changes, it was much much easier to review. I left a lot of comments but 90% of them are about using constants :P and the rest are questions that I had.

include/dqlite.h
cole-miller: Basically I decided you were right in this comment: https://github.com/canonical/dqlite/pull/661#discussion_r1663785674; it seems lower-cost and lower-risk to introduce a new constructor that allows asking for disk mode up front than to make the changes to raft initialization that were required to keep dqlite_enable_disk_mode working.

src/raft/uv.c
cole-miller: Yeah, I introduced this to avoid having to wrap the following lines, but it's probably better to just accept the wrapping :)

src/raft/uv.h
cole-miller: Mainly avoiding casts. The value has to be represented as `uint64_t` at the lowest level since that's what the encoding/decoding functions work with, and life is simpler if we keep it the same type everywhere. I also have a slight aversion to named enum types because I think using a normal integer type reminds you to check for out-of-range values. (On the other hand, I think cc has some special diagnostics that only kick in if you `switch` on a value of enum type.)

src/raft/uv_segment.c
cole-miller: Not sure what you mean---keep it as a field in `struct uvSegmentBuffer`?

src/raft.h
cole-miller: Oops, missed this with `sed`

src/raft/uv_metadata.c
cole-miller: Good catch, I'd been thinking that that this was covered by the check in raft_uv_set_format_version plus the comparison of `uv->format_version` and `metadata.format_version`---but if you don't call raft_uv_set_format version then the first one will never happen. So indeed I should restore the previous check here.

src/raft/uv_metadata.c
cole-miller: Yep!

test/lib/server.c
cole-miller: Better indeed!

src/raft/uv_segment.c
cole-miller: I think the patch uses the symbolic form of `sizeof(struct raft_entry_local_data)` (= 16) consistently---the 8s for `sizeof(uint64_t)` are mostly pre-existing and I was trying to keep the patch focused.

src/server.c
cole-miller: dqlite_vfs_enable_disk relies on VfsInit being called first (to allocate `struct vfs`).

src/server.c
cole-miller: Yeah, good point.

src/raft/uv_segment.c
letFunny: Precisely that

src/raft/uv_segment.c
letFunny: Ah I see. I was not suggesting to refactor, just asking why one is different from the other 16 != 8.

src/server.c
letFunny: You are right

test/raft/integration/test_uv_load.c
cole-miller: Good catch---I've fixed this to use a different buf size depending on whether the test runs with format version 1 or 2. That way we get the same semantics in both cases.

test/lib/server.c
cole-miller: I ended up factoring this out into a function since it's used frequently.

src/raft/uv_segment.c
cole-miller: Yeah, good point, this is more consistent with how we pass in `uv->block_size`.

src/raft/uv.h
letFunny: I am not sure about this. I see your point about having to remember to check that the int is on the range of the enum though. Couldn't we make it so untrusted sources return an integer but when we define functions / structs that expect a valid value we use the enum? That way, when passing the enum you get reminded to check for bounds before casting and we also get the benefits of having more type information.

include/dqlite.h
letFunny: Are we going to support both types of disk_mode in the future? The only thing I am worried about is having to introduce `_v3` version of the function soon.

src/raft/uv_segment.c
letFunny: I see that there is a new field now in `uvSegmentBuffer` but it is not consistently used, e.g. `uvSegmentBufferAppend` takes the format_version as an argument an uses that without accessing the one in the struct.

src/raft/uv_metadata.c
letFunny: Are you sure that doing `metadata = {}` is equivalent to initializing all the fields to 0? I thought that you needed the explicit `= {0}` for that, and without it, it means to take the uninitialized memory.

test/lib/util.h
```diff
@@ -24,4 +24,10 @@
 		}                                                           \
 	} while (0)
 
+static inline bool param_bool(const MunitParameter *params, const char *name)
```
letFunny, line 27: [Nit]: Maybe `param_to_bool`?

test/raft/integration/test_uv_load.c
letFunny: Thanks!

test/lib/util.h
cole-miller: I've changed it to `param_get_bool` which for me describes the intention better, does that work?

include/dqlite.h
cole-miller: No, the old disk mode is going away.

letFunny (approved):
Thanks for the changes to the public API, lgtm!

cole-miller:
As discussed on MM I've pushed commits to remove the `flags` argument to `dqlite_node_create_v2`, which now uses disk mode exclusively. `dqlite_node_enable_disk_mode` remains a no-op.

PR #673, opened 2024-07-17
Snapshot pool integration, first version

letFunny:
This PR integrates the pool with the snapshot installation. This is a first version without the real implementation for the callbacks and I/O. It solves a few issues found along the way, namely it includes:
* pool integration for background work,
* emulating the async nature of rpc sending by scheduling a callback using libuv,
* first versions of rpc_fill and is_a_trigger, that means that the code can now generate primitive messages and expect the correct ones,
* tests now assert that the messages sent are the correct ones,
* added missing state which I noticed when writing new tests.

See the commit list for more information as the functionality is split into self-contained units.

### DIFF
--- a/src/lib/threadpool.c
+++ b/src/lib/threadpool.c
@@ -1,13 +1,15 @@
 #include "threadpool.h"
 #include <assert.h>
 #include <stddef.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 #include <uv.h>
+#include <uv/unix.h>
 #include "../../src/lib/queue.h"
 #include "../../src/lib/sm.h"
 #include "../../src/utils.h"
-#include "../tracing.h"
 
 /**
  *  Planner thread state machine.
@@ -74,6 +76,9 @@ static const struct sm_conf planner_states[PS_NR] = {
 	},
 };
 
+static const uint64_t pool_thread_magic = 0xf344e2;
+static uv_key_t thread_identifier_key;
+
 enum {
 	THREADPOOL_SIZE_MAX = 1024,
 };
@@ -124,6 +129,14 @@ struct pool_impl {
 	uint32_t       qos_prio;       /* QoS prio */
 };
 /* clang-format on */
+
+/* Callback does not allow passing data, we use a static variable to report
+ * errors back. */
+static int thread_key_create_err = 0;
+static void thread_key_create(void) {
+	PRE(thread_key_create_err == 0);
+	thread_key_create_err = uv_key_create(&thread_identifier_key);
+}
 
 static inline bool pool_is_inited(const pool_t *pool)
 {
@@ -325,6 +338,7 @@ static void worker(void *arg)
 	pool_work_t *w;
 	queue *q;
 
+	uv_key_set(&thread_identifier_key, (void*)pool_thread_magic);
 	uv_sem_post(ta->sem);
 	uv_mutex_lock(mutex);
 	for (;;) {
@@ -552,7 +566,17 @@ int pool_init(pool_t *pool,
 		return rc;
 	}
 
+	static uv_once_t once = UV_ONCE_INIT;
+	uv_once(&once, thread_key_create);
+	if (thread_key_create_err != 0) {
+		uv_close((uv_handle_t *)&pi->outq_async, NULL);
+		uv_mutex_destroy(&pi->outq_mutex);
+		free(pi);
+		return thread_key_create_err;
+	}
+
 	pool_threads_init(pool);
+
 	return 0;
 }
 
@@ -579,6 +603,10 @@ void pool_close(pool_t *pool)
 	uv_mutex_lock(&pi->mutex);
 	pi->exiting = true;
 	uv_mutex_unlock(&pi->mutex);
+}
+
+bool pool_is_pool_thread(void) {
+	return uv_key_get(&thread_identifier_key) == (void*)pool_thread_magic;
 }
 
 pool_t *pool_ut_fallback(void)
--- a/src/raft/recv_install_snapshot.c
+++ b/src/raft/recv_install_snapshot.c
@@ -426,7 +426,7 @@ static void leader_work_done(pool_work_t *w)
 	struct work *work = CONTAINER_OF(w, struct work, pool_work);
 	struct leader *leader = CONTAINER_OF(work, struct leader, work);
 
-	PRE(leader->ops->is_main_thread());
+	PRE(!leader->ops->is_pool_thread());
 	sm_move(&work->sm, WORK_DONE);
 	leader_tick(leader, M_WORK_DONE);
 }
@@ -436,7 +436,7 @@ static void follower_work_done(pool_work_t *w)
 	struct work *work = CONTAINER_OF(w, struct work, pool_work);
 	struct follower *follower = CONTAINER_OF(work, struct follower, work);
 
-	PRE(follower->ops->is_main_thread());
+	PRE(!follower->ops->is_pool_thread());
 	sm_move(&work->sm, WORK_DONE);
 	follower_tick(follower, M_WORK_DONE);
 }
@@ -447,7 +447,7 @@ static void rpc_to_cb(uv_timer_t *handle)
 	struct rpc *rpc = CONTAINER_OF(to, struct rpc, timeout);
 	struct leader *leader = CONTAINER_OF(rpc, struct leader, rpc);
 
-	PRE(leader->ops->is_main_thread());
+	PRE(!leader->ops->is_pool_thread());
 	if (sm_state(&to->sm) == TO_CANCELED) {
 		return;
 	}
@@ -461,7 +461,7 @@ static void leader_to_cb(uv_timer_t *handle)
 	struct timeout *to = CONTAINER_OF(handle, struct timeout, handle);
 	struct leader *leader = CONTAINER_OF(to, struct leader, timeout);
 
-	PRE(leader->ops->is_main_thread());
+	PRE(!leader->ops->is_pool_thread());
 	if (sm_state(&to->sm) == TO_CANCELED) {
 		return;
 	}
@@ -492,7 +492,7 @@ static void leader_sent_cb(struct sender *s, int rc)
 	struct rpc *rpc = CONTAINER_OF(s, struct rpc, sender);
 	struct leader *leader = CONTAINER_OF(rpc, struct leader, rpc);
 
-	PRE(leader->ops->is_main_thread());
+	PRE(!leader->ops->is_pool_thread());
 	if (UNLIKELY(rc != 0)) {
 		sm_move(&rpc->sm, RPC_ERROR);
 		return;
@@ -505,7 +505,7 @@ static void follower_sent_cb(struct sender *s, int rc)
 	struct rpc *rpc = CONTAINER_OF(s, struct rpc, sender);
 	struct follower *follower = CONTAINER_OF(rpc, struct follower, rpc);
 
-	PRE(follower->ops->is_main_thread());
+	PRE(!follower->ops->is_pool_thread());
 	if (UNLIKELY(rc != 0)) {
 		sm_move(&rpc->sm, RPC_ERROR);
 		return;
@@ -864,7 +864,7 @@ __attribute__((unused)) void leader_tick(struct leader *leader, const struct raf
 	struct sm *sm = &leader->sm;
 	const struct leader_ops *ops = leader->ops;
 
-	PRE(ops->is_main_thread());
+	PRE(!ops->is_pool_thread());
 
 	if (!is_a_trigger_leader(leader, incoming) ||
 	    is_a_duplicate(leader, incoming))
@@ -941,7 +941,7 @@ __attribute__((unused)) void follower_tick(struct follower *follower, const stru
 	    is_a_duplicate(follower, incoming))
 		return;
 
-	PRE(ops->is_main_thread());
+	PRE(!ops->is_pool_thread());
 
 again:
 	switch (sm_state(&follower->sm)) {
--- a/src/raft/recv_install_snapshot.c
+++ b/src/raft/recv_install_snapshot.c
@@ -16,6 +16,20 @@
 #include "../utils.h"
 #include "../lib/threadpool.h"
 
+#define IN_1(E, X) E == X
+#define IN_2(E, X, ...) E == X || IN_1(E,__VA_ARGS__)
+#define IN_3(E, X, ...) E == X || IN_2(E,__VA_ARGS__)
+#define IN_4(E, X, ...) E == X || IN_3(E,__VA_ARGS__)
+#define IN_5(E, X, ...) E == X || IN_4(E,__VA_ARGS__)
+#define IN_6(E, X, ...) E == X || IN_5(E,__VA_ARGS__)
+#define IN_7(E, X, ...) E == X || IN_6(E,__VA_ARGS__)
+#define IN_8(E, X, ...) E == X || IN_7(E,__VA_ARGS__)
+#define IN_9(E, X, ...) E == X || IN_8(E,__VA_ARGS__)
+
+#define GET_IN_MACRO(_1,_2,_3,_4,_5,_6,_7,_8,_9,NAME,...) NAME
+#define IN(E, ...) \
+  (GET_IN_MACRO(__VA_ARGS__,IN_9,IN_8,IN_7,IN_6,IN_5,IN_4,IN_3,IN_2,IN_1)(E,__VA_ARGS__))
+
 /**
  * =Overview
  *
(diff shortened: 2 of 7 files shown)


test/raft/unit/test_snapshot.c
```diff
@@ -359,26 +399,95 @@ TEST(snapshot_leader, timeouts, set_up, tear_down, 0, NULL) {
...
+static void wait_msg_sent(void) {
+	while (!global_fixture.msg_valid) {
+		uv_run(uv_default_loop(), UV_RUN_NOWAIT);
+	}
+}
+
+/* Decorates the callback used when the pool work is done to set the test
+ * fixture flag to true, then calls the original callback.*/
+static void test_fixture_work_cb(pool_work_t *w) {
+	global_fixture.work_done = true;
```
just-now, line 487: Seems a good time to invent an implementation of `is_main_thread()` macro or a function is much closer than we thought... When I looked an `wait_work()` it wasn't obvious that the spin-lock runs in the same thread with `test_fixture_work_cb()`.

Please put a `// PRE(is_main_thread())` to both functions.

test/raft/unit/test_snapshot.c
```diff
@@ -8,6 +8,7 @@
 #include "../../../src/raft.h"
```
just-now, line 8: Curious, why this file is not a part of dqlite-tests?

test/raft/unit/test_snapshot.c
```diff
@@ -407,79 +524,242 @@ static void pool_rpc_to_expired(struct rpc *rpc)
...
+	ut_leader_message_received(leader, ut_sign_result());
+
+	PRE(sm_state(&leader->sm) == LS_REQ_SIG_LOOP);
+	wait_msg_sent();
+	munit_assert_int(uv_get_msg_sent().type, ==, RAFT_IO_SIGNATURE);
+	PRE(sm_state(&leader->sm) == LS_REQ_SIG_LOOP);
+	ut_leader_message_received(leader, ut_sign_result());
+
+	wait_work();
+	wait_work();
```
just-now, line 674: Why do we need two `wait_work()` calls?

test/raft/unit/test_snapshot.c
```diff
@@ -407,79 +524,242 @@ static void pool_rpc_to_expired(struct rpc *rpc)
...
+	sender_cb_op cb;
+};
+
+void uv_sender_send_cb(uv_work_t *req) {
+	(void)req;
+}
+
+void uv_sender_send_after_cb(uv_work_t *req, int status) {
+	(void)req;
+	(void)status;
```
just-now, line 576: This can be removed.

test/raft/unit/test_snapshot.c
```diff
@@ -407,79 +524,242 @@ static void pool_rpc_to_expired(struct rpc *rpc)
...
+	global_fixture.msg_valid = true;
+	struct uv_sender_send_data *data = req->data;
+	data->cb(data->s, 0);
```
just-now, line 579: put `status` instead of 0.

test/raft/unit/test_snapshot.c
```diff
@@ -407,79 +524,242 @@ static void pool_rpc_to_expired(struct rpc *rpc)
...
+}
+
+int uv_sender_send_op(struct sender *s,
```
just-now, line 582: If I'm not mistaken, `uv_sender_*()` functions can be `static`.

test/raft/unit/test_snapshot.c
```diff
@@ -407,79 +524,242 @@ static void pool_rpc_to_expired(struct rpc *rpc)
...
 		.to_init = pool_to_init_op,
 		.to_stop = pool_to_stop_op,
 		.to_start = pool_to_start_op,
-		.ht_create = ut_ht_create_op,
-		.work_queue = ut_work_queue_op,
-		.sender_send = ut_sender_send_op,
+		.ht_create = pool_ht_create_op,
+		.work_queue = pool_work_queue_op,
+		.sender_send = uv_sender_send_op,
+		.is_main_thread = pool_is_main_thread_op,
```
just-now, line 619: It's better not to override `is_main_thread()`.
You have an implementation of `pool_is_main_thread_op()`
Maybe, it's worth thinking about an implementation designed around `uv_key_create()`. We need to check if it's a good idea to mix `__tls` and `uv_`

test/raft/unit/test_snapshot.c
```diff
@@ -359,26 +399,95 @@ TEST(snapshot_leader, timeouts, set_up, tear_down, 0, NULL) {
...
+
+#define MAGIC_MAIN_THREAD 0xdef1
+
+static __thread int thread_identifier;
+
+static void *pool_set_up(MUNIT_UNUSED const MunitParameter params[],
+                   MUNIT_UNUSED void *user_data)
+{
+	/* Prevent hangs. */
+	alarm(2);
```
just-now, line 451: Why?

test/raft/unit/test_snapshot.c
```diff
@@ -359,26 +399,95 @@ TEST(snapshot_leader, timeouts, set_up, tear_down, 0, NULL) {
...
+	global_fixture.pool.flags |= POOL_FOR_UT;
+	thread_identifier = MAGIC_MAIN_THREAD;
+
+	struct fixture *f = munit_malloc(sizeof *f);
+	return f;
+}
+
+static void pool_tear_down(void *data)
+{
+    free(data);
```
just-now, line 463: `pool_close()` and `pool_fini()` are required.

Question: Do we use "ASAN" builds to test the protocol?

test/raft/unit/test_snapshot.c
```diff
@@ -287,10 +318,12 @@ TEST(snapshot_leader, basic, set_up, tear_down, 0, NULL) {
 	ut_disk_io_done(&leader.work);
```
just-now, line 318: Seems strange to run `_io()` and `io_done()` callbacks twice on the same work item, without `work_init()` and `work_queue()` in the middle.

just-now (commented):
The first series of comments to test_snapshot.c. review is in progress.

test/raft/unit/test_snapshot.c
letFunny: This is so if you do some changes and the tests breaks in a way that makes `wait_work` or `wait_msg_sent` stall, then the execution is terminated after 2s and the test fails but you can still see the results of the rest.

test/raft/unit/test_snapshot.c
letFunny: Yes, we are compiling the file with `-fsanitize=address` but for some reason it is not picking up on leaked resources allocated in the tests. I will look into that.

src/raft/recv_install_snapshot.c
```diff
@@ -478,32 +491,116 @@ static void follower_sent_cb(struct sender *s, int rc)
...
+	if (incoming == M_WORK_DONE) {
+		return state == LS_HT_WAIT
+		    || state == LS_PAGE_SENT
+		    || state == LS_PERSISTED_SIG_PART
+			|| state == LS_PAGE_READ;
+	} else if (incoming == M_MSG_SENT) {
+		if (sm_state(&leader->rpc.sm) != RPC_FILLED) {
+			return false;
+		}
+		return state == LS_PAGE_READ
```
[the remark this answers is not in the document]
just-now: Let's discuss this with @cole-miller

src/raft/recv_install_snapshot.c
```diff
@@ -478,32 +491,116 @@ static void follower_sent_cb(struct sender *s, int rc)
...
+		    || state == LS_SNAP_DONE
+		    || state == LS_F_NEEDS_SNAP
+		    || state == LS_REQ_SIG_LOOP
+		    || state == LS_CHECK_F_HAS_SIGS
+		    || state == LS_WAIT_SIGS;
+	}
+
+	/* From now on, the message pointer is a valid pointer. */
+	PRE(incoming != M_MSG_SENT && incoming != M_TIMEOUT &&
+			incoming != M_WORK_DONE);
```
just-now, line 532: Do we really need it here? 
It's quite clear from the lines above that this assumption is obvious.

src/raft/recv_install_snapshot.c
```diff
@@ -478,32 +491,116 @@ static void follower_sent_cb(struct sender *s, int rc)
...
+	PRE(follower->ops->is_main_thread());
 	if (UNLIKELY(rc != 0)) {
 		sm_move(&rpc->sm, RPC_ERROR);
 		return;
 	}
-
 	follower_tick(follower, M_MSG_SENT);
 }
 
 static bool is_a_trigger_leader(const struct leader *leader, const struct raft_message *incoming)
```
just-now, line 502: Does it seem less noisy? What do you think about such a style?

```
static bool is_a_trigger_leader(const struct leader *leader, const struct raft_message *incoming)
{
        int state = sm_state(&leader->sm);

	return
		ergo(incoming == M_WORK_DONE,
		     IN(state, (LS_HT_WAIT, LS_PAGE_SENT,
				LS_PERSISTED_SIG_PART, LS_PAGE_READ))) &&
		ergo(incoming == M_MSG_SENT,
		     sm_state(&leader->rpc.sm) != RPC_FILLED &&
		     IN(state, (LS_PAGE_READ, LS_SNAP_DONE, LS_F_NEEDS_SNAP,
				LS_REQ_SIG_LOOP, LS_CHECK_F_HAS_SIGS))) &&
		ergo(incoming == M_TIMEOUT,
		     IN(state, (LS_PAGE_READ, LS_SNAP_DONE, LS_F_NEEDS_SNAP,
				LS_REQ_SIG_LOOP, LS_CHECK_F_HAS_SIGS,
				LS_WAIT_SIGS))) &&
		ergo(state == LS_F_ONLINE,
		     incoming->type == RAFT_IO_APPEND_ENTRIES_RESULT) &&
		ergo(sm_state(&leader->rpc.sm) != RPC_SENT, false) &&
		ergo(state == LS_PAGE_READ,
		     IN(incoming->type, (RAFT_IO_INSTALL_SNAPSHOT_CP_RESULT
					 RAFT_IO_INSTALL_SNAPSHOT_MV_RESULT))) &&
		ergo(IN(state, (LS_SNAP_DONE, LS_F_NEEDS_SNAP)),
		     incoming->type == RAFT_IO_INSTALL_SNAPSHOT_RESULT) &&
		ergo(IN(state, (LS_REQ_SIG_LOOP, LS_CHECK_F_HAS_SIGS)),
		     incoming->type == RAFT_IO_SIGNATURE_RESULT);
}
```
cc @cole-miller

just-now (commented):
I like this patch.

cole-miller (commented):
A few comments after reading through. This PR is really well-organized, nice work!

test/raft/unit/test_snapshot.c
letFunny: We discussed this offline, it happens because the leader needs to:
1. read a page from its local db file and get its checksum,
2. lookup the checksum in the hashtable that the follower sent.

Because these are sequential operations it is simpler to schedule the second when the first one is finished. It happens in the following way:
1. disk_io calls the work_cb itself,
2. disk_io_done calls the after_work_cb which internally calls leader_tick(M_WORK_DONE),
3. inside leader tick we move to the next state which schedules another background work,
4. disk_io ...
5. disk_io_done ...
6. leader_tick goes into the next state that sends the message.

test/raft/unit/test_snapshot.c
letFunny: I always forget `static` and `const`, always.

test/raft/unit/test_snapshot.c
letFunny: Okay, now I now why. The tests are in fact compiled with ASAN but `pool_init` does not reserve any memory so it is not noticing that.

src/raft/recv_install_snapshot.c
letFunny: The only reason is that from this line on this is a valid pointer. If you were to do `msg->type` on `M_MSG_SENT`, or any of the others, the program would crash.

test/raft/unit/test_snapshot.c
```diff
@@ -395,6 +436,14 @@ static void pool_to_init_op(struct timeout *to)
 	uv_timer_init(uv_default_loop(), &to->handle);
 }
 
+static void pool_work_queue_op(struct work *w, work_op work_cb, work_op after_cb)
+{
+	w->pool_work = (pool_work_t) { 0 };
+	global_fixture.orig_cb = after_cb;
+	global_fixture.work_done = false;
+	pool_queue_work(&global_fixture.pool, &w->pool_work, 0/* */, WT_UNORD, work_cb, test_fixture_work_cb);
```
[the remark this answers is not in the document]
letFunny: Yep it was meant to be a TODO.

src/raft/recv_install_snapshot.h
```diff
@@ -128,7 +132,7 @@ static const struct sm_conf leader_sm_conf[LS_NR] = {
 		.flags   = SM_INITIAL | SM_FINAL,
 		.name    = "online",
 		.allowed = BITS(LS_HT_WAIT)
-			 | BITS(LS_F_ONLINE),
+		         | BITS(LS_F_ONLINE),
```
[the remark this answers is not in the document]
letFunny: I think now it is correct, there was one extra tab there which meant that changing the tab size made = and | not aligned.

src/raft/recv_install_snapshot.h
```diff
@@ -1,34 +1,36 @@
...
 };
 
 struct sender {
 	sender_cb_op cb;
 };
 
 struct timeout {
-	to_cb_op cb;
 	struct sm sm;
+	to_cb_op cb;
```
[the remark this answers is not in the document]
letFunny: As I understand it, there are two reasons:
* we are not always using `uv_timer_t`, for example the unit tests without libuv,
* there is the extra flexibility of being able to schedule another timer later on with the stored callback (which might be unnecessary in libuv where you can reuse the same one).

test/raft/unit/test_snapshot.c
```diff
@@ -373,12 +381,45 @@ TEST(snapshot_leader, timeouts, set_up, tear_down, 0, NULL) {
...
+	/* true when union contains leader, false when it contains follower */
+	bool is_leader;
+	pool_t pool;
+
+	work_op orig_cb;
+	bool work_done;
+};
+
+/* Not problematic because each test runs in a different process. */
+static struct test_fixture global_fixture;
```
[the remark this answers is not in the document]
letFunny: It should still work with `--no-fork` because the tests are going to be run sequentially. Also, if you are doing that in GDB to not follow the forks, you can instead add a local `.gdbinit` in the project folder with:
```
set follow-fork-mode child
```

src/raft/recv_install_snapshot.c
letFunny: If we are going for this style I would prefer to have two distinct  blocks because in the first three `ergos` the pointer is not valid `M_WORK_DONE`, `M_MSG_SENT` and `M_TIMEOUT` are equal to `1`, `2` and `3`, which means that doing `incoming->type` will crash the program.

My fear is that when we come back and do changes in the future we miss that bit about the ordering because it is not explicit.

test/raft/unit/test_snapshot.c
letFunny: I will also add some comments to the `global_fixture` and these functions.

test/raft/unit/test_snapshot.c
just-now: Please ignore this comment. The initial question was whether the `test/raft/unit/test_snapshot.c` shall be a part of dqlite or raft.

src/raft/recv_install_snapshot.h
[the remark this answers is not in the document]
cole-miller: Makes sense!

src/lib/threadpool.c
```diff
@@ -553,6 +556,15 @@ int pool_init(pool_t *pool,
...
 
 	pool_threads_init(pool);
+
+	rc = uv_key_create(&thread_identifier_key);
+	if (rc != 0) {
+		uv_mutex_destroy(&pi->outq_mutex);
+		free(pi);
+		return rc;
+	}
+	uv_key_set(&thread_identifier_key, (void*)main_thread_magic);
```
cole-miller, line 566: Maybe a silly question but should the key be set to some other value for the child threads?

cole-miller: Never mind, I see that `pthread_key_create` guarantees the associated value is initialized to `NULL`.

src/lib/threadpool.c
```diff
@@ -553,6 +556,15 @@ int pool_init(pool_t *pool,
 	}
 
 	pool_threads_init(pool);
+
+	rc = uv_key_create(&thread_identifier_key);
```
cole-miller, line 560: I think this needs to be wrapped in `uv_once`; the man page for `pthread_key_create` says it's the programmer's responsibilty to not call it more than once on a given address, and that can happen with this code if you have multiple nodes running in a process, or even if you create/run multiple nodes serially.

src/lib/threadpool.c
```diff
@@ -553,6 +556,15 @@ int pool_init(pool_t *pool,
...
+	if (rc != 0) {
+		uv_mutex_destroy(&pi->outq_mutex);
+		free(pi);
+		return rc;
```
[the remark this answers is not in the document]
letFunny: My reasoning was that it is part of the invariants so not having the identification can lead to nastier bugs. If we wanted to go that route we would need to ignore the return of `is_main_thread` maybe three values, sort of yes, no, undefined.

src/lib/threadpool.c
letFunny: You are right! I don't know why but my local man pages didn't have information about that but when I look at https://linux.die.net/man/3/pthread_key_create there is a whole section about it

cole-miller: I don't know about multiple threads but creation of multiple nodes (hence pools) within a process happens in the test suite. Maybe it's harmless to call `uv_key_create` more than once on the same key with the same value as long as there's no race, but the language of the man page seems stricter than that:

cole-miller: I do think that ideally it should be safe to call `dqlite_node_create` from multiple threads without synchronization.

src/lib/threadpool.c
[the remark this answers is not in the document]
cole-miller: That's a good point, if the invariants are always going to be turned on then it becomes hard to stub out is_main_threaad. Let's keep it as-is.

src/lib/threadpool.c
letFunny: I also don't know enough about tls to know if it is safe to create the same key more than once. I agree that ideally we should be able to call `dqlite_node_create` from multiple threads without doing anything special. But then we have to change the approach because there is no single "main thread". What we can do instead is to record whether a thread belongs to the threadpool or not, which is conceptually easier when dealing with multiple pools.

I don't think we need to record the provenance in regards to which pool spawned which thread, but what do you think?

letFunny: Take a look at https://github.com/canonical/dqlite/pull/673/commits/6d2669609db19c249dfb0af27d2e2295be1f4e75, I have changed the code to use `uv_once` and also to identify the pool threads.

cole-miller: Yeah, that is unfortunate.

src/lib/threadpool.c
```diff
@@ -128,6 +130,13 @@ struct pool_impl {
 };
 /* clang-format on */
 
+/* Callback does not allow data passing, we use a static variable to report
+ * errors back. */
+static int thread_key_create_err = 0;
+static void thread_key_create(void) {
+	thread_key_create_err = uv_key_create(&thread_identifier_key);
+}
```
cole-miller, line 138: `PRE(thread_key_create_err == 0)`?

src/lib/threadpool.c
letFunny: I thought about that as well but that would be the complement of the threads that are part of the pool, right? You can always negate the assertion.

src/lib/threadpool.c
letFunny: Should be always that way because of static value and `uv_once` only calling this block once per execution.

cole-miller: Yeah, more for documentation/sanity checking. Thanks!

src/lib/threadpool.c
cole-miller: Well, you could set one magic number for loop threads and another for pool threads, right?

just-now (approved):

src/lib/threadpool.c
letFunny: Discussed offline, we are going to postpone discussing this until a need arrives to identify specific threads, e.g. ones running a specific loop.

cole-miller (approved):
Thanks, this is coming along nicely!

PR #675, opened 2024-07-25
Cleanup prepared statement during node close

marco6:
After preparing more queries on k8s-dqlite, I've been hitting quite frequently some segfaults. They all seemed to be happening when a query is running and the node is shutting down.

Apparently, they were. There was some missing cleanup logic around prepared statements in `gateway__leader_close`. This PR adds that bit.

I see that codecov report is showing a small regression in coverage. If someone from the dqlite team could guide me in testing this, I would be up for it.

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -76,6 +76,15 @@ void gateway__leader_close(struct gateway *g, int reason)
 			 */
 			sqlite3_finalize(g->req->stmt);
 			g->req = NULL;
+		} else if (g->req->type == DQLITE_REQUEST_QUERY) {
+			/* In case the statement is a prepared one, it
+			 * will be finalized by the stmt__registry_close
+			 * call below. Nevertheless, we must signal that
+			 * the request is not in place anymore so that any
+			 * callback which is already in the queue will not
+			 * attempt to execute a finalized statement.
+			 */
+			g->req = NULL;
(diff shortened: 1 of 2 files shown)


cole-miller:
> I see that codecov report is showing a small regression in coverage. If someone from the dqlite team could guide me in testing this, I would be up for it.

I believe this is an artifact of codecov comparing the coverage numbers between your branch's HEAD and master's HEAD, instead of using the merge commit. If you look at the [report](https://app.codecov.io/gh/canonical/dqlite/pull/675), the patch lines that are listed as uncovered are all from the diff between `merge_base(marco6/fix-prepared-cleanup, canonical/master)`, not from your commit. (I've noticed this happening before with other PRs.) So not a blocker for merging.

cole-miller:
For testing, I would try to add something in `test/unit/test_gateway.c`, which supports triggering gateway closure (grep for `CLUSTER_DEPOSE`). The tricky part will be getting that to happen while the gateway is in the middle of sending rows. Maybe this could be done relatively easily by just having a query that returns a ton of rows, even something like `SELECT * FROM generate_series(1)`? That might end up being slow or terribly CPU-intensive, though.

cole-miller (approved):
Test looks great, thank you!

PR #678, opened 2024-08-05
Start instrumenting existing code with state machines

cole-miller:
This PR introduces several state machines to track and instrument the behavior of existing dqlite code. Specifically, we instrument:

- Write transactions on the leader (`leader__exec`)
- Requests to append a `COMMAND` entry to the raft log (`raft_apply`)
- AppendEntries processing on the follower (`replicationAppend`)
- Individual in-memory log entries (`log.c`)
- Disk write requests for entries (`UvAppend`)
- On-disk log truncation (`UvTruncate`)

There are lots more things we could instrument but I think this is a good starting point: it gives us enough visibility to follow a write transaction over its whole lifecycle and down to the disk I/O level, on both the follower and the leader. (A big missing piece is linking the histories across nodes; that's nontrivial because our raft messages don't include any kind of ID or room for extensibility, although we could fake something in the I/O fixture.)

The tracking of in-memory log entries is the trickiest part of this PR. I was initially uncertain whether to attach SMs to individual log entries at all, but I found this served as a convenient "hub" to connect other state machines together (e.g. bridging the `raft_apply` and `UvAppend` state machines), plus it gives a foothold for tracking how long it takes to apply each entry.

I also added state machines for the append and truncate requests in the raft I/O fixture (`raft/fixture.c`). This was necessary because the code in `raft/replication.c` now assumes that the I/O backend makes an SM available to call `sm_relate` with. The SMs in the fixture are not copies of the ones in the uv I/O backend, but simpler ones.

Finally, I made a few fixes and additions to the sm code.


cole-miller:
Pictures!

Leader's view of executing a write transaction in a one-node cluster (`./integration-test cluster/restart`):

leader-exec

From top to bottom, we have the `leader__exec` request, the `raft_apply` request, the new log entry, and the `UvAppend` request. You can see how long it took to executing the transaction initially, how long it took to write it to disk, and how long it took to apply it once committed.

Follower's view of handling an AppendEntries (`./integration-test cluster/dataOnNewNode`):

append-follower

At the top is the main `appendFollower` request. Below that are 10 log entries, and at the bottom is the disk write for the entries.

Finally, a different view of AppendEntries handling on the follower's side. This one uses the fixture (`./raft-core-integration-test replication/recvRollbackConfigurationToInitial`), so the I/O related state machines are simpler.

follower-truncate

Here, the follower truncated an entry from its log because of the AppendEntries. You see the `appendFollower` request at the top, then the truncated entry, then the two entries that were included in the AppendEntries, then the disk append. You can see that the truncate and append operations are concurrent---`replicationApply` doesn't wait for the former to finish before kicking off the latter (something I hadn't noticed before I put together this PR).

src/lib/sm.c
```diff
@@ -49,6 +60,7 @@ void sm_init(struct sm *m,
 	m->is_locked = is_locked;
 	m->id = ++id;
 	m->pid = getpid();
+	m->rc = 0;
```
just-now, line 63: Note: using initializers can help not to think about setting to zero unset attributes:
```
*m = {.conf = conf, .state=state ..};
```

src/raft/fixture.c
```diff
@@ -247,13 +297,57 @@ static void ioFlushAppend(struct io *s, struct append *append)
...
+		memcpy(entries, io->entries, n * sizeof *io->entries);
+
+		/* Release any truncated entry */
+		if (io->entries != NULL) {
+			size_t i;
+			for (i = n; i < io->n; i++) {
+				raft_free(io->entries[i].buf.base);
+			}
+			raft_free(io->entries);
+		}
```
just-now, line 330: Marked code and code starting from line 334 look like copy-paste. 
I see a potential to make the code around it better.

src/raft/replication.c
```diff
@@ -880,15 +887,60 @@ static void sendAppendEntriesResult(
...
+	struct sm sm;
 };
 
+static void append_follower_done(struct appendFollower *req, int status)
+{
+	if (status == 0) {
+		sm_move(&req->sm, AF_DONE);
+	} else {
+		sm_fail(&req->sm, AF_FAILED, status);
+	}
```
just-now, line 939: Such a construction can be seen elsewhere in the patch. I'd expect to look at it this following way:
```
sm_move_rc(&req->sm, status == 0 ? DONE : FAILED, status);
```
So maybe it's worth defining a new function in sm.c/h

src/raft/fixture.c
```diff
@@ -597,48 +698,23 @@ static int ioMethodAppend(struct raft_io *raft_io,
...
-		}
-	}
-
-	io->n = n;
+	struct truncate *r;
 
+	sm_init(&trunc->sm, fio_invariant, NULL, fio_states, "fio-trunc", FIO_START);
+	r = raft_malloc(sizeof(*r));
+	assert(r != NULL);
+	r->type = TRUNCATE;
```
just-now, line 711: It's not always clear how the structure with macro inside is initialized.
```
struct truncate {
	REQUEST;
	struct raft_io_truncate *req;
	raft_index index;
};
```
having initializer `*r = {type=TRUNCATE ... }` makes the reader not to dig into details assuming that non mentioned struct fields will be filled with zeros.

src/raft/log.c
```diff
@@ -243,62 +298,70 @@ static int refsInit(struct raft_log *l,
...
-		if (rc != 0) {
-			return RAFT_NOMEM;
-		}
-
-		if (!collision) {
-			return 0;
+		ref = refsTryInsert(l->refs, l->refs_size, term, index, 1, buf,
+				    batch, &collision);
+		if (ref == NULL && !collision) {
+			return NULL;
```
just-now, line 307: Initial logic looks a bit different. `!collision` looks to be a new element in this case.

src/raft/replication.c
```diff
@@ -1140,25 +1208,38 @@ int replicationAppend(struct raft *r,
...
+	rv = deleteConflictingEntries(r, args, &request->sm, &i);
 	if (rv != 0) {
-		return rv;
+		goto err_after_request_alloc;
 	}
 
 	*rejected = 0;
 
 	n = args->n_entries - i; /* Number of new entries */
+	sm_attr(&request->sm, "n", "%zu", n);
```
just-now, line 1242: nit: it's supposed the viewer of attribute graphs will not just into the code right away so it seems it's better to have more wordy and noisy names for attributes. Example:
```
sm_attr(&request->sm, "new_entires_nr", "%zu", n);
```

just-now (commented):
The patch looks good to me. Just minor changes are needed.

src/raft/replication.c
cole-miller: Good idea.

src/raft/fixture.c
cole-miller: Indeed this can be written much more compactly, thanks!

src/raft/log.c
cole-miller: Fixed, the two return values are unnecessarily hard to reason about so I switched to using a non-NULL sentinel for the collision case, which allows hewing more closely to the original code structure.

src/leader.c
```diff
@@ -10,18 +10,73 @@
...
+	EXEC_START,
+	EXEC_BARRIER,
+	EXEC_STEPPED,
+	EXEC_POLLED,
+	EXEC_DONE,
+	EXEC_FAILED,
+	EXEC_NR,
+};
+
+static const struct sm_conf exec_states[EXEC_NR] = {
```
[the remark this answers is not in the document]
cole-miller: I'm not sure. I experimented with switching the state machine definitions over to this style, and it works well for some of them, but once you have more than a few allowed transitions out of some state the lines get long enough to need breaking, and at that point you're more or less back to the beginning. I've pushed a commit that rewrites the definitions in this style, let me know what you think.

src/lib/sm.c
```diff
@@ -43,12 +54,14 @@ void sm_init(struct sm *m,
 
 	PRE(conf[state].flags & SM_INITIAL);
 
-	m->conf = conf;
-	m->state = state;
-	m->invariant = invariant;
-	m->is_locked = is_locked;
-	m->id = ++id;
-	m->pid = getpid();
+	*m = (struct sm){
```
just-now, line 57: This looks cool as it's easier to grep out initialization with `grep "(struct sm){"`.

src/raft/client.c
```diff
@@ -35,30 +37,41 @@ int raft_apply(struct raft *r,
...
+	sm_init(&req->sm, request_invariant, NULL, request_states, "apply-request",
+		REQUEST_START);
+	queue_insert_tail(&r->leader_state.requests, &req->queue);
+
 	/* Append the new entries to the log. */
-	rv = logAppendCommands(r->log, r->current_term, bufs, local_data, n);
-	if (rv != 0) {
-		goto err;
+	index = start;
+	for (unsigned i = 0; i < n; i++) {
```
just-now, line 52: Is the reason of removing `logAppendCommands()` helper is just cleaner code or I'm missing something?

just-now (approved):
Looks good to me.

src/raft/client.c
cole-miller: It's just a cleanup, yes.

Issue #679, opened 2024-08-07
segfault on arm64, statically compiled with musl


SimonRichardson:
Attempting to bootstrap Juju controller on arm64 with the latest version of dqlite causes a segfault. This using musl to statically compile the jujud-controller. The snapcraft.yaml contains all the dependencies for building the binary.

The CI run in question: https://github.com/juju/juju/actions/runs/10212413197/job/28255682924?pr=17836

Repro steps...

This has to be on aarch64 (arm64). I used multipass on a mac M1, but it could easily be done on an aws graviton machine.

Prerequisites:
 1. git
 2. build-essentials (make, etc).
 3. lxd (you might need to correctly run `sudo lxd init --auto`)
 4. snapcraft (grab from snap)

```sh
$ git checkout https://github.com/juju/juju/pull/17836
$ snapcraft --use-lxd
$ sudo snap install *.snap --dangerous
$ sudo snap connect juju:lxd lxd
$ sudo snap connect juju:config-lxd
$ sudo snap connect juju:dot-local-share-juju
$ sudo snap connect juju:ssh-keys
$ juju bootstrap lxd test --keep-broken
```

The segfault should happen.

```sh
$ lxc exec <juju container name> -- bash
$ apt install gdb
$ LIBDQLITE_TRACE=1 gdb /var/lib/juju/tools/3.6-beta2.1-ubuntu-arm64/jujud bootstrap-state --timeout 20m0s --data-dir '/var/lib/juju' --debug '/var/lib/juju/bootstrap-params'
```

The backtrace for the segfault: https://paste.ubuntu.com/p/FcY8WZ9GT3/

hpidcock:
A follow-up to this can be found [here](https://github.com/juju/juju/pull/17836#issuecomment-2275017290).

TLDR: dqlite needs more than 128k of stack memory, musl's default.

SimonRichardson:
I'd be happy to close this, with an advisory that if using musl you need to increase the default stack size in some docs or README.md.

cole-miller:
@SimonRichardson I've updated the README in #700. I was planning to keep this issue to open as a reminder to myself to investigate why dqlite needs a large stack---there might be another issue like the one with EXEC_SQL lurking.

marco6:
Can we re-evaluate this? I think that we've come a long way since this was open and I think that it is not true anymore that we require a large stack.

PR #681, opened 2024-08-10
gateway: Break recursion in handle_exec_sql

cole-miller:
#679 turned up the fact that we can enter bounded recursion in `handle_exec_sql`, with the maximum number of stack frames being proportional to the number of `;`-separated statements in the `EXEC_SQL` request. This PR breaks that recursion by yielding to the event loop at one link in the chain.

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -16,6 +16,7 @@
 
 void gateway__init(struct gateway *g,
 		   struct config *config,
+		   uv_loop_t *loop,
 		   struct registry *registry,
 		   struct raft *raft,
 		   struct id_state seed)
@@ -34,6 +35,10 @@ void gateway__init(struct gateway *g,
 	g->protocol = DQLITE_PROTOCOL_VERSION;
 	g->client_id = 0;
 	g->random_state = seed;
+	g->sched = (uv_timer_t){};
+	if (loop != NULL) {
+		uv_timer_init(loop, &g->sched);
+	}
 }
 
 void gateway__leader_close(struct gateway *g, int reason)
@@ -96,6 +101,11 @@ void gateway__leader_close(struct gateway *g, int reason)
 void gateway__close(struct gateway *g)
 {
 	tracef("gateway close");
+
+	if (g->sched.loop != NULL) {
+		uv_close((uv_handle_t *)&g->sched, NULL);
+	}
+
 	if (g->leader == NULL) {
 		stmt__registry_close(&g->stmts);
 		return;
@@ -718,6 +728,25 @@ static void handle_exec_sql_next(struct gateway *g,
 				 struct handle *req,
 				 bool done);
 
+static void schedule_exec_sql_next_cb(uv_timer_t *t)
+{
+	struct gateway *g = t->data;
+	PRE(g->req != NULL);
+	PRE(g->req->type == DQLITE_REQUEST_EXEC_SQL);
+	handle_exec_sql_next(g, g->req, true);
+}
+
+static void schedule_exec_sql_next(struct gateway *g)
+{
+	PRE(g->req->type == DQLITE_REQUEST_EXEC_SQL);
+	g->sched.data = g;
+	if (g->sched.loop != NULL) {
+		uv_timer_start(&g->sched, schedule_exec_sql_next_cb, 0, 0);
+	} else {
+		schedule_exec_sql_next_cb(&g->sched);
+	}
+}
+
 static void handle_exec_sql_cb(struct exec *exec, int status)
 {
 	tracef("handle exec sql cb status %d", status);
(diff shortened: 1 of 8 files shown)


test/integration/test_client.c
```diff
@@ -145,3 +145,29 @@ TEST(client, querySql, setUp, tearDown, 0, client_params)
...
+
+/* Stress test of an EXEC_SQL with many ';'-separated statements. */
+TEST(client, semicolons, setUp, tearDown, 0, NULL)
+{
+	struct fixture *f = data;
+	(void)params;
+
+	static const char trivial_stmt[] = "CREATE TABLE IF NOT EXISTS foo (n INT);";
+
+	size_t n = 1000;
```
cole-miller, line 157: Note to reviewer: I confirmed on my machine that this value is large enough to cause a stack overflow with the old implementation.

src/gateway.c
```diff
@@ -728,7 +757,15 @@ static void handle_exec_sql_cb(struct exec *exec, int status)
 	sqlite3_finalize(exec->stmt);
 
 	if (status == SQLITE_DONE) {
-		handle_exec_sql_next(g, req, true);
+		/* The next line simply for the event loop to call
```
louiseschmidtgen, line 760:
```suggestion
		/* The next line simply allows for the event loop to call
```

hpidcock:
I've tested this in https://github.com/juju/juju/pull/17911 but it doesn't fix the issue. The stack size on arm64 still needs to be larger than 128k.

But this fix is still important, so we look forward to seeing it land.

src/gateway.c
```diff
@@ -718,6 +728,25 @@ static void handle_exec_sql_next(struct gateway *g,
 				 struct handle *req,
 				 bool done);
 
+static void schedule_exec_sql_next_cb(uv_timer_t *t)
+{
+	struct gateway *g = t->data;
+	PRE(g->req != NULL);
```
letFunny, line 734: Nit: `PRE(g != NULL && g->req != NULL)`?

test/integration/test_client.c
letFunny: Is there a way to test that this crashes if the loop is NULL in `gateway__init` to see that the test is meaningful?

src/gateway.h
```diff
@@ -35,10 +35,20 @@ struct gateway {
...
+	/* handle_exec_sql uses this to defer some work to the next loop
+	 * iteration. */
+	uv_timer_t sched;
 };
 
+/**
+ * Initialize a gateway.
+ *
+ * Passing NULL for the `loop` is permitted. Currently the loop is only used
+ * optionally to break potential recursion in handle_exec_sql.
```
letFunny, line 47: I think it is better if the comment does not refer to the internal function name `handle_exec_sql` but instead talks about the higher level idea.
```suggestion
 * Passing NULL for the `loop` is permitted. Currently the loop is only used
 * optionally to break potential recursion when handling a query that bundles
 *  several sql statements delimited by semicolons.
```

src/gateway.c
```diff
@@ -718,6 +728,25 @@ static void handle_exec_sql_next(struct gateway *g,
...
+	PRE(g->req->type == DQLITE_REQUEST_EXEC_SQL);
+	handle_exec_sql_next(g, g->req, true);
+}
+
+static void schedule_exec_sql_next(struct gateway *g)
+{
+	PRE(g->req->type == DQLITE_REQUEST_EXEC_SQL);
+	g->sched.data = g;
```
letFunny, line 742: Nit: maybe move this to inside the if statement.

src/gateway.h
```diff
@@ -35,10 +35,20 @@ struct gateway {
 	uint64_t protocol;           /* Protocol format version */
 	uint64_t client_id;
 	struct id_state random_state; /* For generating IDs */
+	/* handle_exec_sql uses this to defer some work to the next loop
```
letFunny, line 38: Same comment about `handle_exec_sql` being internal and this being part of the external documentation.

letFunny (approved):
Thanks Cole for fixing this, looking very good, I just left a couple of nitpicks as usual :)

just-now (approved):
Look good to me.

test/integration/test_client.c
cole-miller: I can't see a non-brittle way to do it. Unfortunately we don't have scaffolding for tests of the gateway that also have an event loop available, otherwise that would make it very natural.

cole-miller:
I've addressed comments from the review (thanks @louiseschmidtgen and @letFunny) and pushed another commit that makes this change more precise, so we don't defer the call to `handle_exec_sql_next` unless this function is actually being called recursively.

src/gateway.c
```diff
@@ -727,12 +745,27 @@ static void handle_exec_sql_cb(struct exec *exec, int status)
...
+	 * here. But that can lead to bounded recursion when we have several
+	 * `;`-separated statements in a row that do not generate rows. To make
+	 * sure the stack depth stays under control, we defer have the event
+	 * loop invoke handle_exec_sql_next itself on the next iteration, but
+	 * only if there is a prior call to handle_exec_sql_next above us on
+	 * the stack. We also invoke handle_exec_sql_next directly if the
+	 * gateway doesn't have access to an event loop (this is only the case
+	 * in the unit tests). */
+	if (exec->async || g->defer.loop == NULL) {
+		handle_exec_sql_next(g, req, true);
```
just-now, line 765: Question: Is it worth having an explicit recursion counter as a parameter of a function or an associated structure?
Just curious.

just-now (approved):
Looks good.

src/gateway.c
cole-miller: Good question---I think the best version of this would be to have a depth counter field on `struct exec` and `struct barrier`. But then with the approach in this PR, that field would only ever be 0 or 1, so we'd just be treating it as a boolean anyway. (Unless we wanted to allow some positive recursion depth, but I'm not sure how to pick that constant.)

PR #682, opened 2024-08-11
github: Test with musl

cole-miller:
MicroK8s and Juju consume dqlite as a static library build against musl libc, so let's test this in our CI to avoid unexpected breaks like #679.


cole-miller:
Running into this issue it seems: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=764335.

contrib/build-static.sh
```diff
@@ -0,0 +1,112 @@
...
+
+REPO_MUSL_COMPAT="https://github.com/juju/musl-compat"
+REPO_LIBTIRPC="https://salsa.debian.org/debian/libtirpc"
+REPO_LIBNSL="https://github.com/thkukuk/libnsl"
+REPO_LIBUV="https://github.com/libuv/libuv"
+REPO_LIBLZ4="https://github.com/lz4/lz4"
+REPO_SQLITE="https://github.com/sqlite/sqlite"
+REPO_DQLITE="https://github.com/cole-miller/dqlite"
+
+TAG_MUSL_COMPAT="main"
```
louiseschmidtgen, line 11: Could you please include a test run against the musl tag used in our k8s distributions: https://github.com/canonical/k8s-dqlite/blob/master/hack/env.sh#L13? We might not always be on the same musl version as the juju team.

SimonRichardson: Juju will be happy to move to the later one, now we know the root cause. We intended to move that originally but ended up locating the segfault.

cole-miller: Sounds good.

cole-miller:
I like how using a Dockerfile for this cleans things up. Looking into the segfaults.

configure.ac
```diff
@@ -99,7 +99,7 @@ CC_CHECK_FLAGS_APPEND([AM_CFLAGS],[CFLAGS],[ \
   -fexceptions \
   -fstack-clash-protection \
   -fstack-protector-strong \
-  -fasynchronous-unwind-tables \
+  -fno-exceptions \
```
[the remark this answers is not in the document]
letFunny: Maybe this is used for libbacktrace?

contrib/build-static.sh
```diff
@@ -0,0 +1,149 @@
...
+REPO_SQLITE="https://github.com/sqlite/sqlite.git"
+REPO_DQLITE="https://github.com/cole-miller/dqlite.git"
+
+TAG_MUSL="v1.2.4"
+TAG_LIBTIRPC="upstream/1.3.3"
+TAG_LIBNSL="v2.0.1"
+TAG_LIBUV="v1.48.0"
+TAG_LIBLZ4="v1.9.4"
+TAG_SQLITE="version-3.45.1"
+TAG_DQLITE="musl-ci"
```
letFunny, line 19: Do we need to update the tags manually when new versions come out? Does it make more sense to use the latest version instead?

test/lib/heap.c
```diff
@@ -177,34 +177,7 @@ void test_heap_setup(const MunitParameter params[], void *user_data)
 /* Ensure we're starting leaving a clean memory behind. */
 void test_heap_tear_down(void *data)
 {
-	sqlite3_mem_methods mem;
```
letFunny, line 180: Why are we removing the tear_down completely?

test/raft/integration/test_uv_tcp_connect_addrinfo.c
```diff
@@ -0,0 +1,231 @@
+#include "../../../src/raft.h"
```
letFunny, line 1: I wonder if there is a way to define the test only for glibc without having to duplicate the rest of the file. The other problem is keeping the CI and the make recipes in sync when adding new tests/binaries.

I see that musl does not define a macro to detect it, but maybe there is a way:
* Either define a macro that gets passed when we want to enable the tests, `gcc -D...`.
* Or identify glibc using `__USE_GNU` and use that to guard the tests.

We just have to #ifdef and skip the test early.

letFunny (commented):
Thanks Cole! Just a couple of comments and questions.

test/lib/heap.c
cole-miller: The only point of the teardown was to complain if there were outstanding allocations, but that's already commented out because it was buggy/flaky. Without the teardown we leave the faulty allocator in place, but that's okay because it behaves exactly the same as the default allocator unless you ask for a fault explicitly. It's also just impossible to soundly restore the allocator because SQLite bans calling `sqlite3_config(SQLITE_CONFIG_MALLOC, ...)` after any other sqlite3 interface has been used (https://sqlite.org/c3ref/config.html).

contrib/build-static.sh
cole-miller: Hmm, it probably makes sense to keep in sync with Juju and MicroK8s on the tested version of musl, but latest versions of everything else sounds good to me. We could merge this with the `latest-deps` CI job.

configure.ac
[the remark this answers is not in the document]
cole-miller: I added the libbacktrace integration and this was present in configure.ac much earlier, but good call that it might affect the integration. Will test

test/raft/integration/test_uv_tcp_connect_addrinfo.c
cole-miller: > The other problem is keeping the CI and the make recipes in sync when adding new tests/binaries.

I'm on this, running subsets of tests seems to be poorly supported by Automake but we can have an environment variable that causes the addrinfo binary to return without running tests

> We just have to #ifdef and skip the test early.

The problem is that any tests we want to run cannot be linked into the same binary as `test_addrinfo.c` because that file overrides `getaddrinfo` and it's impossible to avoid calling it. The only way to do it would be to effectively ifdef out all of `test_addrinfo.c` along with the tests that use it. If we went this route, rather than detecting glibc or musl I think it would be better to detect how libc is being linked since that's what makes the difference; we'd need to do some compiler flag sniffing or add a new option in configure.ac. I could go either way on this

cole-miller: I've reverted the commit that introduced the new test binary and switched to controlling this with a `configure.ac` option (that also causes `-all-static` to be passed) and conditional compilation. We can get away without duplicating or complicating the test source files, which is nice.

contrib/build-static.sh
cole-miller: Updated to use latest versions of dependencies except musl, which is pinned at v1.2.4.

configure.ac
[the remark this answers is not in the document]
cole-miller: Confirmed that libbacktrace still works when `-fno-exceptions` is passed.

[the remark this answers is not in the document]
cole-miller: So the DWARF unwind tables are used by libbacktrace and also by GDB for debugging. However, it looks like the build errors disappear if we don't pass `-fexceptions` nor `-fasynchronous-unwind-tables` nor the complements of these. GDB and libbacktrace definitely work with that setup (assuming you pass `-g`, which we do for debug builds), so we should be good.

test/lib/heap.c
cole-miller: I sanity-checked myself on this with GDB and I'm wrong---we call `sqlite3_shutdown` in the sqlite fixture, which should make this code okay. Reverted the commit. I think what I was seeing when working on this PR is a pre-existing issue where we don't use the sqlite fixture everywhere we should; I've added a separate commit to fix this in #694 since it was affecting that PR and the fix is quick. Thanks for poking at this!

letFunny: Thanks for looking into it Cole.

Makefile.am
```diff
@@ -351,6 +363,11 @@ endif
 
 TESTS = $(check_PROGRAMS)
 
+check-norun: $(check_PROGRAMS)
```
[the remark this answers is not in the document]
cole-miller: Yeah, makes sense.

Makefile.am
```diff
@@ -215,8 +223,12 @@ libtest_la_SOURCES += \
   test/raft/lib/tcp.c \
   test/raft/lib/loop.c
 
+if !WITH_STATIC_DEPS
+libtest_la_SOURCES += test/raft/lib/addrinfo.c
```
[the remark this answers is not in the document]
cole-miller: Yes, good call.

PR #683, opened 2024-08-14
Add an API for querying the last raft log entry

cole-miller:
This enables better automation of dqlite_node_recover_ext by exposing enough information to determine which of several nodes has the most up-to-date log. See https://github.com/canonical/go-dqlite/issues/299. Exposing this from libdqlite is the first step, then we can add a wrapper API in go-dqlite.

The implementation in this PR is as simple as calling the `load` method of the uv raft_io object. This is wasteful---it loads all snapshots and entries into memory---but that seems acceptable to me, since it's no more expensive than restarting the node normally and the use-case is an exceptional failure condition (loss of quorum).

### DIFF
--- a/src/raft/raft.c
+++ b/src/raft/raft.c
@@ -323,9 +323,10 @@ int raft_io_describe_last_entry(struct raft_io *io,
 	if (rv != 0) {
 		return rv;
 	}
-	POST(ERGO(n_entries == 0, snapshot != NULL));
 	*index = start_index + n_entries - 1;
-	*term = n_entries > 0 ? entries[n_entries - 1].term : snapshot->term;
+	*term = n_entries > 0 ? entries[n_entries - 1].term :
+		snapshot != NULL ? snapshot->term :
+		0;
 	if (snapshot != NULL) {
 		snapshotDestroy(snapshot);
 	}
--- a/src/raft/raft.c
+++ b/src/raft/raft.c
@@ -10,11 +10,13 @@
 #include "configuration.h"
 #include "convert.h"
 #include "election.h"
+#include "entry.h"
 #include "err.h"
 #include "flags.h"
 #include "heap.h"
 #include "log.h"
 #include "membership.h"
+#include "snapshot.h"
 
 #define DEFAULT_ELECTION_TIMEOUT 1000          /* One second */
 #define DEFAULT_HEARTBEAT_TIMEOUT 100          /* One tenth of a second */
@@ -301,5 +303,32 @@ static int ioFsmVersionCheck(struct raft *r,
 		return -1;
 	}
 
+	return 0;
+}
+
+int raft_io_describe_last_entry(struct raft_io *io,
+				raft_index *index,
+				raft_term *term)
+{
+	raft_term current_term;
+	raft_id voted_for;
+	struct raft_snapshot *snapshot;
+	raft_index start_index;
+	struct raft_entry *entries;
+	size_t n_entries;
+	int rv;
+
+	rv = io->load(io, &current_term, &voted_for, &snapshot,
+		      &start_index, &entries, &n_entries);
+	if (rv != 0) {
+		return rv;
+	}
+	POST(ERGO(n_entries == 0, snapshot != NULL));
+	*index = start_index + n_entries - 1;
+	*term = n_entries > 0 ? entries[n_entries - 1].term : snapshot->term;
+	if (snapshot != NULL) {
+		snapshotDestroy(snapshot);
+	}
+	entryBatchesDestroy(entries, n_entries);
 	return 0;
 }
--- a/src/server.c
+++ b/src/server.c
@@ -1065,6 +1065,25 @@ out:
 	return rv;
 }
 
+int dqlite_node_describe_last_entry(dqlite_node *n,
+				    uint64_t *index,
+				    uint64_t *term)
+{
+	static_assert(sizeof(*index) == sizeof(raft_index),
+		      "unexpected index type size");
+	raft_index *i = (raft_index *)index;
+	static_assert(sizeof(*term) == sizeof(raft_term),
+		      "unexpected term type size");
+	raft_term *t = (raft_term *)term;
+	int rv;
+
+	rv = raft_io_describe_last_entry(&n->raft_io, i, t);
+	if (rv != 0) {
+		return DQLITE_ERROR;
+	}
+	return 0;
+}
+
 dqlite_node_id dqlite_generate_node_id(const char *address)
 {
 	tracef("generate node id");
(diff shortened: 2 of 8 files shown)


cole-miller:
Rather than fix the external libraft CI job, I am going to block this on a separate PR to make `--enable-build-raft` the default and only option.

Other test failures seem to be nondeterministic and due to these numbers varying from test to test, which I need to dig into:

https://github.com/cole-miller/dqlite/blob/6546d543d339691c1ee1f185cbcd02aa4a6be1ba/test/integration/test_cluster.c#L124-L134

test/integration/test_cluster.c
```diff
@@ -97,22 +97,48 @@ TEST(cluster, restart, setUp, tearDown, 0, cluster_params)
...
+	 *
+	 * At 993 records, the leader has generated one legacy checkpoint
+	 * command entry.
+	 *
+	 * At 2200 records, the leader has generated a second checkpoint
+	 * command, plus two barriers. (The number of barriers is nondeterministic;
+	 * but we seem to get two of them pretty consistently in this test.) */
+	size_t fudge = n_records >= 2200 ? 6 :
+			n_records >= 993 ? 3 :
+			2;
```
letFunny, line 138: This seems a bit brittle especially the part about the barriers which are non-deterministic which could lead to CI failures down the line. What about changing the assertion so that if n_records < 993 then,  `last_entry_index <= n_records + 3 && last_entry_index >= n_records`? (or something similar, the idea is to use an interval and not a single value)

src/raft/raft.c
```diff
@@ -303,3 +305,30 @@ static int ioFsmVersionCheck(struct raft *r,
...
+	struct raft_entry *entries;
+	size_t n_entries;
+	int rv;
+
+	rv = io->load(io, &current_term, &voted_for, &snapshot,
+		      &start_index, &entries, &n_entries);
+	if (rv != 0) {
+		return rv;
+	}
+	POST(ERGO(n_entries == 0, snapshot != NULL));
```
letFunny, line 326: This is going to crash when `raft_io_describe_last_entry` is called and there are no entries in the raft log + snapshots, correct? If that is the case, should this be a crash or an empty return?

letFunny (commented):
Thanks for this Cole, just a couple of questions.

letFunny (commented):
Forgot to publish one comment :P

src/raft/raft.c
```diff
@@ -303,3 +305,30 @@ static int ioFsmVersionCheck(struct raft *r,
...
+				raft_term *term)
+{
+	raft_term current_term;
+	raft_id voted_for;
+	struct raft_snapshot *snapshot;
+	raft_index start_index;
+	struct raft_entry *entries;
+	size_t n_entries;
+	int rv;
+
```
just-now, line 320: Is it possible to express and assert the following statement as a precondition?
```
 * This (raft_io_describe_last_entry) should be called after dqlite_node_init, but 
 * the node must not be running.
```

src/server.c
```diff
@@ -1065,6 +1065,25 @@ int dqlite_node_recover_ext(dqlite_node *n,
...
+	static_assert(sizeof(*index) == sizeof(raft_index),
+		      "unexpected index type size");
+	raft_index *i = (raft_index *)index;
+	static_assert(sizeof(*term) == sizeof(raft_term),
+		      "unexpected term type size");
+	raft_term *t = (raft_term *)term;
+	int rv;
+
+	rv = raft_io_describe_last_entry(&n->raft_io, i, t);
+	if (rv != 0) {
```
just-now, line 1081: Nit: can be expressed simpler: ```return rv == 0 ? rv : DQLITE_ERROR```

just-now (approved):
Looks good.

src/raft/raft.c
cole-miller: Good point, thanks, this could potentially be used with the data directory of a node that never got the chance to write a log entry to disk (when bootstrapping we always write the entry with the initial configuration, but joiner nodes can start with nothing at all). In this case we should probably return `(0, 0)` since that will compare less than the tuple for any node that does have entries on disk.

test/integration/test_cluster.c
cole-miller: I've modified this to use the calculated number just as a lower bound. Feels to me like writing something like `last_entry_index <= n_records + CONSTANT` is not much better than picking an exact value because we don't have an explanation for the specific `CONSTANT`; I would rather just have the assertion reflect our belief about the code i.e. "some number of barriers, we don't know how many".

src/raft/raft.c
cole-miller: Fixed in the latest commit.

cole-miller:
Updated to reflect all review comments, plus I have addded a commit to fix #689 to avoid the friction of a separate PR. If we decide that the behavior for one-node elections should stay the same I can update the assertions in this PR accordingly.

test/integration/test_cluster.c
```diff
@@ -97,17 +97,16 @@ TEST(cluster, restart, setUp, tearDown, 0, cluster_params)
...
 	PREPARE("CREATE TABLE test (n INT)", &stmt_id);
 	EXEC(stmt_id, &last_insert_id, &rows_affected);
 
+	PREPARE("INSERT INTO TEST(n) VALUES(?)", &stmt_id);
 	for (int i = 0; i < n_records; ++i) {
-		sprintf(sql, "INSERT INTO test(n) VALUES(%d)", i + 1);
-		PREPARE(sql, &stmt_id);
-		EXEC(stmt_id, &last_insert_id, &rows_affected);
+		EXEC_PARAMS(stmt_id, &last_insert_id, &rows_affected,
+			    {.type = SQLITE_INTEGER, .integer = i});
```
cole-miller, line 109: A fun fact is that this macro invocation sees *five* arguments, the last two of which are `{.type = SQLITE_INTEGER` and `.integer = i}`.

test/integration/test_cluster.c
```diff
@@ -131,12 +131,13 @@ TEST(cluster, restart, setUp, tearDown, 0, cluster_params)
...
-	 * but we seem to get two of them pretty consistently in this test.) */
-	size_t fudge = n_records >= 2200 ? 6 :
+	 * command. */
+	size_t extra = n_records >= 2200 ? 4 :
 			n_records >= 993 ? 3 :
 			2;
-	munit_assert_ullong(last_entry_index, ==, n_records + fudge);
+	/* This assertion isn't exact because we expect to see some barrier
+	 * log entries as well, and the number of these is not deterministic. */
+	munit_assert_ullong(last_entry_index, >=, n_records + extra);
```
letFunny: You could also assert that it is within a range, that would catch errors where this grows too much but won't make the test flaky.

letFunny (approved):
Thanks Cole, I reviewed the last three commits and everything is there. I only added one comment to set an upper bound for the assertions with the extra records. Also, I remember we talked about the term number not increasing, is that what is happening here? (only one node, term is not increased)

test/integration/test_cluster.c
cole-miller: Done

PR #686, opened 2024-08-16
Fix use-after-free in unit tests

cole-miller:
Free the pool's `pool_impl_t` containing the `outq_async` handle only after the callback passed to `uv_close` has fired, in line with [libuv rules](https://docs.libuv.org/en/v1.x/handle.html#c.uv_close), and in contrast to the original code, which could free this allocation while libuv was still holding a pointer into it. This UAF showed up in CI runs for #682 but is hard to reproduce otherwise.

Note that dqlite as used in production is not affected by this bug since the thread pool is not even created in that case.

### DIFF
--- a/src/lib/threadpool.h
+++ b/src/lib/threadpool.h
@@ -104,8 +104,20 @@ int pool_init(pool_t *pool,
 	      uv_loop_t *loop,
 	      uint32_t threads_nr,
 	      uint32_t qos_prio);
-void pool_fini(pool_t *pool);
+/**
+ * Start the closing sequence for the thread pool.
+ *
+ * This signals the threads of the pool to finish their work and exit. The pool
+ * threads will not be joined, and the resources held by the pool will not be
+ * released, until pool_fini is called. Before calling pool_fini, the event loop
+ * that was used to create this pool must be run to completion (that is, until
+ * uv_run returns 0).
+ */
 void pool_close(pool_t *pool);
+/**
+ * Finish the closing sequence for the thread pool and release resources.
+ */
+void pool_fini(pool_t *pool);
 void pool_queue_work(pool_t *pool,
 		     pool_work_t *w,
 		     uint32_t cookie,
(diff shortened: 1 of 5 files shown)


cole-miller:
I suspect this was behind the elusive hangs we were seeing in CI as well...

cole-miller:
Need to fix pool setup/teardown in server.c as well.

cole-miller:
I discussed this offline with @just-now, who pointed out that the pool was designed based on the assumption that pool_fini would not be called until the associated event loop had been run to completion. This makes it sound to free the `pool_impl_t` in pool_fini. I've updated the tests to use pool_fini as intended and documented the thread pool's closing sequence.

src/lib/threadpool.h
```diff
@@ -104,8 +104,20 @@ int pool_init(pool_t *pool,
...
 	      uint32_t qos_prio);
-void pool_fini(pool_t *pool);
+/**
+ * Start the closing sequence for the thread pool.
+ *
+ * This signals the threads of the pool to finish their work and exit. The pool
+ * threads will not be joined, and the resources held by the pool will not be
+ * released, until pool_fini is called. Before calling pool_fini, the event loop
+ * that was used to create this pool must be run to completion (that is, until
+ * uv_run returns 0).
```
letFunny, line 114: Just to check my understanding, here we are exploiting the fact that ([source](https://docs.libuv.org/en/v1.x/handle.html#c.uv_close)):
> Handles that wrap file descriptors are closed immediately

So calling uv_close will close the handle and we can free it immediately after. However, we still have to run the event loop to completion prior to doing that because the handles might be in use resulting in UAF.

Is that right?

letFunny (approved):

just-now (approved):

src/lib/threadpool.h
cole-miller: > So calling uv_close will close the handle and we can free it immediately after.

No, it won't be closed immediately (`uv_async_t` doesn't wrap an fd). It'll be closed asynchronously, but we don't free it until after the event loop has exited, which implies that all handles including the pool's async handle have finished closing.

Issue #684, opened 2024-08-15
MicroOVN built with latest Dqlite crashes on cluster formation.


mkalcok:
Today we noticed that when MicroOVN is build with the dqlite from master branch it crashes when the third member joins the cluster with following message in `microovn.daemon` service logs:
```
Aug 15 09:31:19 microovn-basic-cluster-3 microovn.daemon[2258]: time="2024-08-15T09:31:19Z" level=info msg=" - binding https socket" network="10.112.216.95:6443"
Aug 15 09:31:19 microovn-basic-cluster-3 microovn.daemon[2258]: microovnd: ./src/unix/core.c:302: uv__finish_close: Assertion `handle->flags & UV_HANDLE_CLOSING' failed.
Aug 15 09:31:19 microovn-basic-cluster-3 systemd[1]: snap.microovn.daemon.service: Main process exited, code=dumped, status=6/ABRT
```
This problem did not occur before today.

**Steps to reproduce:**
* Get [MicroOVN code that uses Dqlite from master](https://github.com/canonical/microovn/tree/e58d1e03a220dbbf823cdeb4606e18578f0bca80) (rev. e58d1e0)
* [Build and install](https://canonical-microovn.readthedocs-hosted.com/en/latest/developers/building/) MicroOVN on three hosts.
  * LXD containers can be used for running MicroOVN, they just need `openvswitch` kernel module. You can launch them with `lxc launch ubuntu:noble <container_name> --config linux.kernel_modules=openvswitch`
Bootstrap the cluster and add new members
```
# On host 1
microovn cluster bootstrap
microovn cluster add <host2-hostname>
microovn cluster add <host3-hostname>
```
Join the cluster with 2nd member
```
# On host 2
microovn cluster join <token>
```
Join the cluster with 3rd member
```
# On host 3
microovn cluster join <token>
# Following message will be displayed:
# Error: Post "http://control.socket/core/control": EOF
```
Observe MicroOVN logs on the 3rd node
```
# On host 3
journalctl -xfeu snap.microovn.daemon
```
The service on the 3rd host will be in restart loop, but the first attempt to start will show above mentioned error.

**Workaround:**
[Pin the dqlite dependency](https://github.com/canonical/microovn/commit/5802dc9a545e87df8c94a4f5c78b805ed7a6b9c5)

cole-miller:
Thanks, seeing this in the musl CI runs for #682 as well. I'm investigating.

cole-miller:
[...]

PR #687, opened 2024-08-18
Break recursion in handle_exec_sql_next (second attempt)

cole-miller:
Revised version of the reverted #681, fixing the regression #684. The bug in that PR was of the same type as that fixed in #686: passing a `NULL` callback when `uv_close`-ing a handle and then freeing the allocation containing the handle immediately.

To fix this instance of the bug, I put the `uv_timer_t` in the `conn` object instead of the gateway, since `conn` already has an asynchronous closing sequence. This means the gateway goes back to not knowing that it's running on an event loop, which is convenient for the tests.

It's probably most efficient to review this by looking at the diff against the previous PR, as merged:

```console
$ git remote add cole https://github.com/cole-miller/dqlite
$ git fetch cole
$ git diff 7dea935c44a32d1a4006ae23e072aef6f9cce509 cole/exec-sql-suspend-v2
```

### DIFF
--- a/src/conn.c
+++ b/src/conn.c
@@ -100,14 +100,21 @@ abort:
 	conn__stop(c);
 }
 
-static void closeCb(struct transport *transport)
+static void conn_defer_close_cb(uv_handle_t *t)
+{
+	struct conn *c = t->data;
+	if (c->close_cb != NULL) {
+		c->close_cb(c);
+	}
+}
+
+static void conn_transport_close_cb(struct transport *transport)
 {
 	struct conn *c = transport->data;
 	buffer__close(&c->write);
 	buffer__close(&c->read);
-	if (c->close_cb != NULL) {
-		c->close_cb(c);
-	}
+	c->defer.data = c;
+	uv_close((uv_handle_t *)&c->defer, conn_defer_close_cb);
 }
 
 static void raft_connect(struct conn *c)
@@ -127,7 +134,8 @@ static void raft_connect(struct conn *c)
 	/* Close the connection without actually closing the transport, since
 	 * the stream will be used by raft */
 	c->closed = true;
-	closeCb(&c->transport);
+	gateway__close(&c->gateway);
+	conn_transport_close_cb(&c->transport);
 }
 
 static void read_request_cb(struct transport *transport, int status)
@@ -289,6 +297,28 @@ static int read_protocol(struct conn *c)
 	return 0;
 }
 
+static void conn_defer_cb(uv_timer_t *defer)
+{
+	struct conn *c = defer->data;
+	void (*cb)(void *) = c->defer_cb;
+	void *arg = c->defer_arg;
+	c->defer_cb = NULL;
+	c->defer_arg = NULL;
+	PRE(cb != NULL);
+	cb(arg);
+}
+
+static void conn_defer(void (*cb)(void *arg), void *arg, void *data)
+{
+	struct conn *c = data;
+	PRE(c->defer_cb == NULL);
+	PRE(c->defer_arg == NULL);
+	c->defer_cb = cb;
+	c->defer_arg = arg;
+	c->defer.data = c;
+	uv_timer_start(&c->defer, conn_defer_cb, 0, 0);
+}
+
 int conn__start(struct conn *c,
 		struct config *config,
 		struct uv_loop_s *loop,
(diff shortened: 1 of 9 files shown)


cole-miller:
Looking into the assertion failures in CI---seems like I didn't account for what happens in the case of a transaction that's split across several `EXEC`/`EXEC_SQL` requests.

cole-miller:
Assertions fixed, it was a silly error. This should be good to go now.

src/leader.h
```diff
@@ -63,8 +66,12 @@ struct exec {
 	uint64_t id;
 	int status;
 	queue queue;
-	exec_cb cb;
 	pool_work_t work;
+	/* When the callback is invoked, this field is `true` if raft_barrier
+	 * or raft_apply was called and `false` if the callback was invoked
+	 * immediately. */
```
letFunny, line 72: IMO this comment is talking about how this field is set from outside not about what the field means or how it is used by `struct exec`.

src/conn.c
```diff
@@ -311,7 +341,10 @@ int conn__start(struct conn *c,
 	c->transport.data = c;
 	c->uv_transport = uv_transport;
 	c->close_cb = close_cb;
-	gateway__init(&c->gateway, config, registry, raft, seed);
+	uv_timer_init(loop, &c->defer);
+	c->defer_cb = NULL;
+	c->defer_arg = NULL;
```
letFunny, line 346: Just to triple-check, there is always going to be a maximum of one exec request per connection, right? So this fields are never going to be overwritten while they are in use.

src/conn.c
```diff
@@ -127,7 +134,8 @@ static void raft_connect(struct conn *c)
 	/* Close the connection without actually closing the transport, since
 	 * the stream will be used by raft */
```
letFunny, line 135: This comment is a little bit outdated now that we call `conn_transport_close_cb`. Either the function name is misleading (but it is in fact a callback for closing the transport) or the comment is a little bit short now and should say something but we still free the resources.

However, this leads me to believe you can call `conn_transport_close_cb` without closing the transport which means that maybe it is clearer if we changed the name of the function and we didn't use it as the callback for uv_close(transport, ...).

src/conn.c
```diff
@@ -100,16 +100,23 @@ static void gateway_handle_cb(struct handle *req,
 	conn__stop(c);
 }
 
-static void closeCb(struct transport *transport)
+static void conn_defer_close_cb(uv_handle_t *t)
```
letFunny, line 103: It is a pity that we have to create so many callbacks just to move data around and to call the functions with the appropriate arguments. This is mostly unavoidable because you cannot inline functions in C, but there are also some cases that we could simplify. For example, we could change `close_cb` to get a `uv_handle_t` if it makes sense. That we we can remove one layer of callbacks. Probably there are other opportunities like this but I am not sure if those involve a larger refactor.

letFunny (commented):
Thanks Cole! It is looking very good, I just left a few comments.

src/conn.c
cole-miller: Yep.

src/leader.h
cole-miller: I think it makes sense as-is, the field is set by code in leader.c and it's part of the API of that translation unit as seen by the gateway. Maybe there's a way I could reword it that would make more sense?

src/conn.c
cole-miller: Indeed, `conn_transport_close_cb` is not responsible for closing the transport, rather it's the callback that runs after the transport has been closed. This naming convention is confusing but it's also used a lot in the existing code. In this case I like the idea of renaming that function to make more clear what it does, but I think we should still re-use it as the callback passed to `uv_close(transport)` because the two uses are conceptually connected. I can update the comment to explain this better as well.

src/conn.c
cole-miller: I agree it's a pity to have so many little adapter callbacks but I feel like it's the least bad option in this case. I'm reluctant to change the signature of the conn close callback because it sort of leaks implementation details about the closing sequence into the interface of this module. You shorten conn.c by one function, but then all callsites of conn__start have to do the work of retrieving the conn from the handle, and if you want to know why the handle is there you have to go read conn.c.

src/leader.h
cole-miller: Rewrote this, also removed the obsolete comment on leader__exec.

letFunny (approved):
I think I reviewed the diff correctly even with the force push. Thanks Cole for this, LGTM!

PR #690, opened 2024-08-26
Remove option to link separately-built libraft

cole-miller:
This removes the option to link libraft, so the bundled raft sources will be used unconditionally.


just-now (approved):
Looks good to me.

PR #694, opened 2024-08-30
Grab stack traces when unit test hangs

cole-miller:
We know which binary is hanging now, try to get some backtraces using GDB if it looks like it's hanging.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -28,6 +28,8 @@ jobs:
       run: |
           sudo apt update
           sudo apt install -y lcov libsqlite3-dev liblz4-dev libuv1-dev
+          # TODO: remove once the mysterious hang is fixed
+          sudo apt install -y gdb
 
     - name: Build dqlite
       env:
@@ -48,6 +50,7 @@ jobs:
         CC: ${{ matrix.compiler }}
         LIBDQLITE_TRACE: 1
       run: |
+          # TODO: return to just `make check` once the mysterious hang is fixed
           tests="integration-test \
                  raft-core-fuzzy-test \
                  raft-core-integration-test \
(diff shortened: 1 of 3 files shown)


cole-miller:
Running the unit test with `--no-fork` exposed a pre-existing minor test issue where we weren't using the `test/lib/sqlite.c` fixture everywhere we should---I just fixed this quickly as well.

.github/workflows/build-and-test.yml
```diff
@@ -48,20 +48,18 @@ jobs:
...
 
-    - name: OnTestFailure
-      if: ${{ failure() }}
-      uses: actions/upload-artifact@v3
-      with:
-        name: test-suite.log
-        path: test-suite.log
+          ./unit-test --no-fork &
+          pid=$!
+          bash -c "sleep 10m; sudo gdb -p $pid -batch -ex 'thread apply all bt' -ex quit; false" &
```
letFunny, line 61: Are you sure gdb is in the container image?

.github/workflows/build-and-test.yml
```diff
@@ -48,20 +48,18 @@ jobs:
...
         LIBDQLITE_TRACE: 1
       run: |
-          for bin in unit-test integration-test \
-                     raft-core-fuzzy-test raft-core-integration-test \
-                     raft-core-unit-test raft-uv-integration-test \
-                     raft-uv-unit-test
-          do ./$bin || touch any-failed
-          done 2>&1 | tee -a test-suite.log
-          test '!' -e any-failed
+          tests="integration-test \
```
letFunny, line 51: Can you add a TODO somewhere to revert this once we figure out why it hangs?

cole-miller: Done

.github/workflows/build-and-test.yml
cole-miller: It is not, good catch! Fixed.

PR #695, opened 2024-09-02
Fixes for sm.c

cole-miller:
Small fixes and one addition for sm.c, pulled from #678 to make review easier.


just-now (approved):
Looks good.

PR #696, opened 2024-09-02
Fixed query test. Added more iterations to uv_run() to deliver pool's callbacks.

just-now:


cole-miller (approved):
I feel like the magic loop iteration counts are a sign that we need to rethink these tests, still, glad to have a fix (hopefully) for the hanging test, thanks!

PR #697, opened 2024-09-02
Refactor leader.c to fix stack growth in handle_exec_sql

cole-miller:
This PR is another attempt to fix the stack growth issue with handle_exec_sql (related to #679), this time in a more principled way that involves a significant refactor of leader.c. Details follow.

Recall that the issue arises with a call stack like this (callees at the top)

```
handle_exec_sql_next
handle_exec_sql_cb
leaderExecV2
leader__barrier
leader__exec
handle_exec_sql_next
...
```

that is, indirect recursion that can generate a number of stack frames proportional to the number of `;`-separated statements in an `EXEC_SQL` request. This happens because leader__barrier and leader__exec can invoke their callbacks synchronously when suspending is not required (respectively, because the FSM is up to date with the raft log and and because `sqlite3_step` generated no changed pages).

This PR rewrites those two functions, establishing a new contract: the callback is only invoked if we yielded to the event loop at least once while processing the request; if the request was processed synchronously, a magic value `LEADER_NOT_ASYNC` is returned to indicate that the *caller* should invoke the callback. Existing callsites are updated to reflect this new contract.

With these new implementations available, we can fix the original problem by rewriting `handle_exec_sql_next` to *iteratively* process as many statements as possible until one of them suspends or returns an error. The PR includes a regression test to validate this fix.

The new implementations of the leader.c functions are intentionally different in style from the old ones. Effectively, each barrier request and exec request is a coroutine driven by a state machine. With this approach the `LEADER_NOT_ASYNC` feature falls out rather naturally, and we get the added benefits of more compact code and built-in observability.


cole-miller:
Jepsen run at https://github.com/canonical/jepsen.dqlite/actions/runs/10672883562.

cole-miller:
Status: trying to identify and fix the regression seen here https://github.com/canonical/lxd/pull/14034, namely `sqlite3_close` returning nonzero in `leader__close` (indicating that a prepared statement was not finalized).

cole-miller:
I believe the LXD issue has been fixed, the problem was returning 0 from `exec_async` in a case where we didn't suspend. This caused `handle_exec_sql_cb` to not be invoked, leaking a prepared statement and tripping `assert(sqlite3_close(conn) == 0)`.

src/leader.c
```diff
@@ -392,9 +332,129 @@ static int leaderApplyFrames(struct exec *req,
...
+	barrier->cb = NULL;
+
+	if (state == BARRIER_PASSED) {
+		cb(barrier, status);
+	}
+}
+
+static void barrier_raft_cb(struct raft_barrier *, int);
+
+static int barrier_async(struct barrier *barrier, int status)
```
letFunny, line 384: Nit: Isn't this more like a `barrier_tick`? It moves the state machine and is called when needed both from aync and sync contexts.

src/gateway.c
```diff
@@ -513,10 +519,11 @@ static int handle_exec(struct gateway *g, struct handle *req)
 	}
 	req->stmt_id = stmt->id;
 	g->req = req;
-	req_id = idNext(&g->random_state);
-	rv = leader__exec(g->leader, &g->exec, stmt->stmt, req_id,
-			  leader_exec_cb);
-	if (rv != 0) {
+	rv = leader_exec_v2(g->leader, &g->exec, stmt->stmt, leader_exec_cb);
+	if (rv == LEADER_NOT_ASYNC) {
+		/* XXX */
```
letFunny, line 524: ?

src/gateway.c
```diff
@@ -513,10 +519,11 @@ static int handle_exec(struct gateway *g, struct handle *req)
 	}
 	req->stmt_id = stmt->id;
 	g->req = req;
-	req_id = idNext(&g->random_state);
```
letFunny, line 516: Why are we removing the concept of the id? I can still see several instances of `req->id` for example in leader.c:281 and we are not removing the field from the struct, why?

src/gateway.c
```diff
@@ -719,66 +737,66 @@ static void handle_exec_sql_next(struct gateway *g,
...
+		rv = sqlite3_prepare_v2(l->conn, sql, -1, &stmt, &tail);
 		if (rv != SQLITE_OK) {
-			failure(req, rv, "bind parameters");
-			goto done_after_prepare;
+			tracef("exec sql prepare failed %d", rv);
+			failure(req, rv, sqlite3_errmsg(l->conn));
+			goto done;
+		}
+		if (stmt == NULL) {
+			goto success;
```
letFunny, line 771: I don't get the combination with `/* stmt will be set to NULL by sqlite when an error occurs. */`.

test/integration/test_client.c
```diff
@@ -145,3 +145,29 @@ TEST(client, querySql, setUp, tearDown, 0, client_params)
 
 	return MUNIT_OK;
 }
+
+/* Stress test of an EXEC_SQL with many ';'-separated statements. */
+TEST(client, semicolons, setUp, tearDown, 0, NULL)
```
letFunny, line 150: How do we know from this test that it is executing all of the statements separated by a semicolon instead of only the first one. Maybe we could change it to use an `INSERT` and count the total number of rows at the end.

src/gateway.c
```diff
@@ -719,66 +737,66 @@ static void handle_exec_sql_next(struct gateway *g,
...
-	rv =
-	    leader__exec(g->leader, &g->exec, stmt, req_id, handle_exec_sql_cb);
-	if (rv != SQLITE_OK) {
-		failure(req, rv, sqlite3_errmsg(g->leader->conn));
-		goto done_after_prepare;
+		req->sql = tail;
+		g->req = req;
+		rv = leader_exec_v2(g->leader, &g->exec, stmt, handle_exec_sql_cb);
+		if (rv == 0) {
+			return;
```
letFunny, line 785: Can we add a comment here saying that the function suspended and hence we are returning. Even better, can we add a comment to the top of this function that it is similar to the PR description, I find that it makes the code here much more readable:
> handle_exec_sql_next to iteratively process as many statements as possible until one of them suspends or returns an error

letFunny (commented):
Looks good to me, thanks Cole. I just have a few questions to improve my understanding and some very small suggestions.

test/integration/test_client.c
cole-miller: Seems like a good idea!

src/gateway.c
cole-miller: Yes, a comment here is a good idea.

src/gateway.c
cole-miller: It's not a great comment, if an error occurs then `rv != SQLITE_OK` and also `stmt` is set to `NULL`. If `rv == SQLITE_OK` and also `stmt == NULL` it means there was nothing to prepare in the given string, so we've executed all statements successfully. I'll fix the comment

src/gateway.c
cole-miller: `idNext` et al are code that I introduced a while ago to provide a way of tracking requests in traces. Now we have `sm.c` that does the same thing better (generalizes to all kinds of requests, `sm_relate`, no RNG), so I thought it was time to get rid of it. I'll clean up the remaining mentions of the field, thanks for spotting

src/leader.c
cole-miller: Yeah, `tick` is a good name.

src/gateway.c
cole-miller: Done

test/integration/test_client.c
cole-miller: Ah, a good trick here would be to add a single `INSERT` at the end and check that you can select it out again. That does the job of demonstrating that the whole string is executed.

src/gateway.c
cole-miller: Left over from experimenting with different ways to write the new "invoke callback manually" codepaths, no defect that I forgot to remove afaict.

test/integration/test_client.c
cole-miller: Done!

test/unit/test_gateway.c
```diff
@@ -918,8 +918,11 @@ TEST_CASE(exec, close_while_in_flight, NULL)
 		EXEC("INSERT INTO test(n) VALUES(1)");
 	}
 
+
 	/* Trigger a second page cache flush to the WAL, and abort before it's
 	 * done. */
+	/* FIXME(cole) it seems that this may no longer be successfully triggering
+	 * the page cache flush */
```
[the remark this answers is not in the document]
cole-miller: I don't remember why I thought this, but I just did some quick printf investigation and verified that in this test we see writes to the WAL both before and after line 924. So I'm satisfied that this test is working as intended, will remove the FIXME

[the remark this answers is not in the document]
letFunny: It is not ideal that the test does not include the check you had to do with printf so we know when it is not working as expected without manual investigation.

test/integration/test_client.c
letFunny: You are right! Good idea, test is looking perfect now.

PR #704, opened 2024-09-16
Remove dqlite-next build configuration

cole-miller:


src/raft/uv_encoding.c
```diff
@@ -90,11 +90,7 @@ size_t uvSizeofBatchHeader(size_t n, bool with_local_data)
 {
 	size_t res = 8 + /* Number of entries in the batch, little endian */
 		16 * n; /* One header per entry */;
-	if (with_local_data) {
-#ifdef DQLITE_NEXT
-		res += 8; /* Local data length, applies to all entries */
-#endif
-	}
+	(void)with_local_data;
```
letFunny, line 93: Can you add a TODO or something in each of these methods saying that this argument is not used and should be removed, same with the local data struct declaration.

letFunny (approved):

src/raft/uv_encoding.c
letFunny: I am okay removing them later, but we need to keep track of it.

cole-miller: Yep, good catch.

PR #709, opened 2024-10-02
Opt-in querying of non-leaders

cole-miller:
This PR is a first stab at allowing non-leader nodes to serve PREPARE, QUERY, and QUERY_SQL requests. When served by a non-leader, such requests will in general return stale data (QUERY, QUERY_SQL), or fail unhelpfully (PREPARE and QUERY_SQL, see #395). But when these limitations are tolerable, the option to query a non-leader can be helpful to spread load or (when targeting the local node) avoid network round-trips.

Clients must opt into relaxing the current leader checks on a per-connection basis. This is done by repurposing the old, unused `flags` field on the OPEN request (which non-leaders already accept). We relax the leader checks when the least-significant bit is set in `flags`; this is backwards-compatible since extant versions of go-dqlite always set this field to 0. The specific relaxations are:

- PREPARE is always allowed. (FINALIZE is also allowed, but it didn't have a leader check in the first place.)
- QUERY and QUERY_SQL are allowed when the statement in question is readonly (recall #477).

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -35,8 +35,6 @@ void gateway__init(struct gateway *g,
 
 void gateway__leader_close(struct gateway *g, int reason)
 {
-	bool allow_stale;
-
 	if (g == NULL || g->leader == NULL) {
 		tracef("gateway:%p or gateway->leader are NULL", g);
 		return;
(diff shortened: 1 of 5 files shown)


cole-miller:
Note to self/reviewer: might be worth discussing whether we should skip in leader__barrier if allow_stale is set but we happen to be the leader. In the current version we do skip in this situation, and I think that's the right choice, but I could be convinced otherwise.

cole-miller:
I just realized there's an issue I didn't deal with---if you connect to a node that happens to be the leader and open a DB with with the `ALLOW_STALE` flag, the connection will still be shut down when the node loses leadership. I'll fix that quickly...

cole-miller:
Pushed a new revision incorporating feedback from @marco6---the semantics of the OPEN flag have changed, in particular it now blocks modifications even when the server is currently the leader. This makes the behavior of connections using this flag more predictable and pushes clients to choose the right kind of connection for their use-case. The flag has been renamed from `ALLOW_STALE` to `READONLY` to reflect this.

PR #714, opened 2024-10-18
Dynamic trailing computation

marco6:
This PR adds a non-default strategy to dynamically compute the amount of entries to keep in memory based on the size of the snapshot.

The problem with using a fixed number is that each entry has a different size (depending on the number of pages written for each transaction). As such, the memory used (and then streamed to a lagging follower) might be way bigger than the size of the snapshot itself (even by order of magnitude).


marco6:
This PR lacks testing for now.

letFunny:
Thanks Marco for this PR, it looks very good to me, no comments so far. For the testing part of the story, we have the snapshot tests at `test/raft/integration/test_snapshot.c` so we could probably add this alternate strategy as a parameter for the existing tests there (where it makes sense). We can also add a new test for the dynamic strategy asserting that the entries in the snapshot match the calculation for the expected size. Actually I was going to suggest changing existing test to use both strategies by abstracting it away but I see that most of them use a threshold of 1 so it will probably require changing all of them, so adding a new one seems easier.

There are other tests where we can also add the new strategy for completeness, for example ones for replication (`test_replication.c`).

Lastly, do not look at the tests under `test/raft/unit/test_snapshot.c` because those are for the new snapshot procedure that was never released.

marco6:
After the conversation I had with @just-now, I changed slightly the strategy so that the computed trailing always sits in the range `[threshold, trailing]` so that the previous snapshot is still useful.

I also added some tests which hopefully cover the functionality now. I'm marking this as ready.

src/raft.h
```diff
@@ -765,6 +765,28 @@ struct raft_transfer; /* Forward declaration */
...
+	 * The static strategy is to just keep a fixed amount of entries
+	 * in the log cache. The number of entries can be tuned with 
+	 * @raft_set_snapshot_trailing. This is the default.
+	 */
+	RAFT_TRAILING_STRATEGY_STATIC,
+	
+	/**
+	 * The dynamic strategy computes the number of entries to keep by 
+	 * comparing the size held by each entry to the size of the snapshot.
+	 * The idea behind this is that if the entries are too big, streaming
```
just-now, line 783: > The idea behind this is that if the entries are too big, streaming the snapshot might be better.

Could you consider to rephrase it? This is a possible variant but it's not ideal:
If the amount of memory (on-disk or RAM) needed to store log entities exceeds the amount of memory for snapshot, streaming the snapshot is more efficient.

src/raft/raft.c
```diff
@@ -172,6 +174,19 @@ void raft_set_snapshot_trailing(struct raft *r, unsigned n)
...
+void raft_set_snapshot_trailing_strategy(struct raft *r, int strategy)
+{
+	switch (strategy) {
+	case RAFT_TRAILING_STRATEGY_STATIC:
+	case RAFT_TRAILING_STRATEGY_DYNAMIC:
+		r->snapshot.trailing_strategy = (uint8_t)strategy;
+		break;
+	default:
+		r->snapshot.trailing_strategy = RAFT_TRAILING_STRATEGY_STATIC;
+		break;
```
just-now, line 186: nit: here, `break;` is not required.

src/raft/replication.c
```diff
@@ -1674,6 +1674,50 @@ static void takeSnapshotClose(struct raft *r, struct raft_snapshot *s)
...
+static unsigned dynamicTrailingIndex(struct raft *r, struct raft_snapshot *snapshot) {
+	struct raft_log *l = r->log;
+	size_t snapshot_size = 0, num_entries = logNumEntries(l);
+	unsigned trailing, i;
+	/** 
+	 * This should never happen as it would mean that the snapshot
+	 * contains no new entry (i.e. either the snapshot is empty
+	 * or equal to the previous one).
+	 */
+	assert(num_entries > 0);
```
just-now, line 1687: If it's a precondition of the function it's required to use `PRE()`.

src/raft/replication.c
```diff
@@ -1674,6 +1674,50 @@ static void takeSnapshotClose(struct raft *r, struct raft_snapshot *s)
 	r->fsm->snapshot_finalize(r->fsm, &s->bufs, &s->n_bufs);
 }
 
+
+static unsigned dynamicTrailingIndex(struct raft *r, struct raft_snapshot *snapshot) {
```
just-now, line 1678: It seems the last argument of the function could be `const`.

src/raft/replication.c
```diff
@@ -1674,6 +1674,50 @@ static void takeSnapshotClose(struct raft *r, struct raft_snapshot *s)
...
+	struct raft_log *l = r->log;
+	size_t snapshot_size = 0, num_entries = logNumEntries(l);
+	unsigned trailing, i;
```
just-now, line 1681: > unsigned

I was triggered here because I started to think about the relation between number of bytes needed to represent `snapshot->n_bufs` and `i` which is used for indexing here. Could you either change `unsigned` to `size_t` or explain why it is sufficient.

src/raft/replication.c
```diff
@@ -1674,6 +1674,50 @@ static void takeSnapshotClose(struct raft *r, struct raft_snapshot *s)
...
+		trailing = r->snapshot.threshold;
+	} else if (trailing > r->snapshot.trailing) {
+		trailing = r->snapshot.trailing;
+	}
+
+	return trailing;
+}
+
+
+static unsigned trailingSize(struct raft *r, struct raft_snapshot *snapshot) {
```
just-now, line 1711: Could you think what's needed to make this function to take just `const`-s:
`static unsigned trailingSize(const struct raft *r, const struct raft_snapshot *snapshot)`
?
In case if it's hard please or I'm missing something please tell me.

just-now (commented):
Looks very good. Just minor nit-picks.

src/raft/replication.c
marco6: Ok fair. I used `unsigned` because `snapshot->n_bufs` is an `unsigned` and the return value is and `unsigned`, too.

However you are right, `num_entries` is a `size_t`, so we should really be consistent there to avoid weird stuff.

Given that the return value must be an `unsigned`, but we are capping the value to the settings `trailing` I think it's fair to convert after the capping has taken place as if `unsigned` is narrower than `size_t` we still know that no truncation happens.

src/raft/replication.c
marco6: I added this in a separate commit (a145249) as the const propagatest (nicely) as usual in 6 other methods (and some more `static` ones, but I don't really care about those)

just-now (approved):
Looks good. Merge it (preferably on Monday).

PR #728, opened 2024-11-25
In replication scenario, entry_sm gets updated every time it is used.

just-now:
It's not expected that the reference to the `entry` is taken in all places in the code. That's why every time it's accessed without taking a reference it's needed to save term and index of the `entry` to access it later through `log_get_entry_sm()`.

### DIFF
--- a/src/raft/replication.c
+++ b/src/raft/replication.c
@@ -1838,7 +1838,10 @@ int replicationApply(struct raft *r)
 			tracef("replicationApply - ENTRY NULL");
 			return 0;
 		}
-		struct sm *entry_sm = log_get_entry_sm(r->log, entry->term, index);
+		raft_term entry_sm_term = entry->term;
+		raft_index entry_sm_index = index;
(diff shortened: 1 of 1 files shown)


src/raft/replication.c
```diff
@@ -1865,6 +1874,9 @@ int replicationApply(struct raft *r)
 		}
```
letFunny, line 1875: For clarity I would add `entry_sm = NULL; entry = NULL` and move the comment here instead, rephrasing it like: "Note that applyCommand may changed the log and thus the references to `entry` and `entry_sm` are no longer valid."

letFunny (approved):
Thanks for this Anatoliy, I just left a comment that hopefully makes the intent more clear and makes it harder for people to misuse it in the future.

src/raft/replication.c
just-now: `entry` itself comes from a different structure, managed in a different way and it's not needed to update it.
I moved the comment below. I agree that it could be shortened as what happens with reference counting here is obvious and doesn't need a lecture.

Issue #530, opened 2023-08-17
Consider periodically using VACUUM to reduce memory footprint


freeekanayaka:
I've experimented using [`VACUUM`](https://www.sqlite.org/lang_vacuum.html) against a real-world medium-size Incus/LXD database (~700 containers) that had a size of 41 Megabytes. After running the `VACUUM` command, the size of the database was down to 3 Megabytes.

Depending on the workflow, if there are large databases that create issues with memory/snapshotting, this might be a relatively cheap and effective measure to explore. It might not have such a dramatic effect in all cases though.

freeekanayaka:
The database I mentioned had been running for several years, most probably performing a number of INSERTs and DELETEs over the course of time, which presumably slowly increased database fragmentation more and more. That means that running VACUUM regularly is advisable, especially for workloads with a lot of data updates.

I also tried to run `.dump` on that database and create it from scratch with the same data, and I got the same result: database size went from 41 MB to 3 MB, as it would be expected.

MathieuBordere:
Haven't looked into it yet, but weren't there issues with running `VACUUM` on a dqlite database, did you have to adapt some things?

MathieuBordere:
```
The VACUUM command works by copying the contents of the database into a temporary database file and then overwriting the original with the contents of the temporary file. When overwriting the original, a rollback journal or [write-ahead log](https://www.sqlite.org/wal.html) WAL file is used just as it would be for any other database transaction.
```

I'm just wondering how it would work in dqlite's case

freeekanayaka:
I can think of several design options. But in my mind it's most probably something that's needed for long-running deployments.

freeekanayaka:
If you are looking for something cheap, this could initially be implemented "offline", e.g. when the process restarts.

cole-miller:
VACUUM needs to be coordinated using a Raft log entry, since it affects page numbers in the future.

marco6:
VACUUM is now supported as a query. All should however bear in mind that it potentially create a very big transaction which has effects both on the RAFT protocol *and* on resource usage.

I would recommend not to run it unsupervised.

PR #734, opened 2025-02-19
Allow vacuum on main

marco6:
As discussed in these days on MM, VACUUM seems to be a transactional operation.

While it is true that it needs to attach a temporary database, new pages are then copied *in the WAL* and, as such, the attachment of the temporary file does not violate the VFS assumption about WAL and main file being unique after every statement.

The tricky part is however to allow sqlite to attach a database only during VACUUM. This PR achieves that by parsing the sql statement.

### DIFF
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -637,7 +637,6 @@ struct vfsFile
 	enum vfsFileType type;        /* Associated file (main db or WAL). */
 	struct vfsDatabase *database; /* Underlying database content. */
 	int flags;                    /* Flags passed to xOpen */
-	sqlite3_file *temp;           /* For temp-files, actual VFS. */
 	sqlite3_file *db;             /* For on-disk DB files, actual VFS. */
 };
 
@@ -764,14 +763,6 @@ static int vfsFileClose(sqlite3_file *file)
 	int rc = SQLITE_OK;
 	struct vfsFile *f = (struct vfsFile *)file;
 	struct vfs *v = (struct vfs *)(f->vfs);
-
-	if (f->temp != NULL) {
-		/* Close the actual temporary file. */
-		rc = f->temp->pMethods->xClose(f->temp);
-		sqlite3_free(f->temp);
-
-		return rc;
-	}
 
 	if (f->flags & SQLITE_OPEN_DELETEONCLOSE) {
 		rc = vfsDeleteDatabase(v, f->database->name);
@@ -933,11 +924,6 @@ static int vfsFileRead(sqlite3_file *file,
 	assert(offset >= 0);
 	assert(f != NULL);
 
-	if (f->temp != NULL) {
-		/* Read from the actual temporary file. */
-		return f->temp->pMethods->xRead(f->temp, buf, amount, offset);
-	}
-
 	switch (f->type) {
 		case VFS__DATABASE:
 			rv = vfsDatabaseRead(f->database, buf, amount, offset);
@@ -1087,11 +1073,6 @@ static int vfsFileWrite(sqlite3_file *file,
 	assert(buf != NULL);
 	assert(amount > 0);
 	assert(f != NULL);
-
-	if (f->temp != NULL) {
-		/* Write to the actual temporary file. */
-		return f->temp->pMethods->xWrite(f->temp, buf, amount, offset);
-	}
 
 	switch (f->type) {
 		case VFS__DATABASE:
@@ -1813,16 +1794,6 @@ static int vfsOpen(sqlite3_vfs *vfs,
 	 */
 	assert(!exclusive || create);
 
-	v = (struct vfs *)(vfs->pAppData);
-	f = (struct vfsFile *)file;
-
-	/* This tells SQLite to not call Close() in case we return an error. */
-	f->base.pMethods = 0;
-	f->temp = NULL;
-
-	/* Save the flags */
-	f->flags = flags;
-
 	/* From SQLite documentation:
 	 *
 	 * If the zFilename parameter to xOpen is a NULL pointer then xOpen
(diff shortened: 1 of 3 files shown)


marco6:
While I don't like a lot the implementation (the parsing of vacuum) this would fix  #530.

marco6:
The implementation seems now better after the call with @just-now. I'm marking this as ready.

src/db.c
```diff
@@ -119,11 +134,8 @@ static int open_follower_conn(const char *filename,
 	 * each connection will operate on only one DB file/WAL file
 	 * pair. Make sure that the client can't use ATTACH DATABASE to
 	 * break this assumption. We apply the same limit in openConnection
-	 * in leader.c.
-	 *
-	 * Note, 0 instead of 1 -- apparently the "initial database" is not
-	 * counted when evaluating this limit. */
-	sqlite3_limit(*conn, SQLITE_LIMIT_ATTACHED, 0);
+	 * in leader.c.*/
+	sqlite3_set_authorizer(*conn, dqlite_follower_authorizer, NULL);
```
letFunny, line 138: This is very neat, I didn't know there were mechanisms for such fine control in SQLite.

src/vfs.c
```diff
@@ -1836,24 +1795,15 @@ static int vfsOpen(sqlite3_vfs *vfs,
...
-		rc = vfs->xOpen(vfs, NULL, f->temp, flags, out_flags);
-		if (rc != SQLITE_OK) {
-			sqlite3_free(f->temp);
-			return rc;
-		}
-
-		f->base.pMethods = &vfsFileMethods;
-		f->vfs = NULL;
-		f->database = NULL;
+	assert((flags & SQLITE_OPEN_DELETEONCLOSE) == 0);
```
letFunny, line 1801: Should we document why this is the case? To check my understanding, SQLITE_OPEN_DELETEONCLOSE is only set for temporary files and cannot be used as a flag in sqlite_open_v2, is that correct?

src/vfs.c
```diff
@@ -2067,15 +2017,18 @@ int VfsInit(struct sqlite3_vfs *vfs, const char *name)
...
+	struct vfs *v = vfsCreate();
+	if (v == NULL) {
 		return DQLITE_NOMEM;
 	}
-
+	vfs->pAppData = v;
+	vfs->szOsFile = sizeof(struct vfsFile);
+	if (vfs->szOsFile < v->base_vfs->szOsFile) {
+		vfs->szOsFile = v->base_vfs->szOsFile;
+	}
```
letFunny, line 2031: What is this fixing?

src/vfs.c
```diff
@@ -3239,25 +3161,15 @@ static int vfsDiskOpen(sqlite3_vfs *vfs,
...
 
-		f->base.pMethods = &vfsDiskFileMethods;
-		f->vfs = NULL;
-		f->database = NULL;
+	assert((flags & SQLITE_OPEN_DELETEONCLOSE) == 0);
+	f = (struct vfsFile *)file;
 
-		return SQLITE_OK;
-	}
+	/* This tells SQLite to not call Close() in case we return an error. */
```
letFunny, line 3170: Are we purposefully changing the behavior by moving this to after the temporary file is created? I think the context I am missing is why didn't we want to call Close when an error happened, I thought the reason was that the temporary file didn't use the same vfs calls as a regular db file and why close was going to throw an error.

test/unit/test_gateway.c
```diff
@@ -1060,6 +1060,122 @@ TEST_CASE(exec, undo_not_leader_pending_other_elected, NULL)
...
+		"    SELECT 1 UNION ALL     "
+		"    SELECT n+1 FROM seq    "
+		"    WHERE  n < 10000       "
+		")                          "
+		"INSERT INTO test(n)        "
+		"SELECT n FROM seq          "
+	)
+
+	const char* vacuum_variants[] = {
+		" VACUUM \r\t\n ",
```
letFunny, line 1118: Isn't this case covering the previous TEST_CASE as well?

test/unit/test_gateway.c
```diff
@@ -1060,6 +1060,122 @@ TEST_CASE(exec, undo_not_leader_pending_other_elected, NULL)
...
+		"    SELECT 1 UNION ALL     "
+		"    SELECT n+1 FROM seq    "
+		"    WHERE  n < 10000       "
+		")                          "
+		"INSERT INTO test(n)        "
+		"SELECT n FROM seq          "
+	)
+	EXEC("DELETE FROM test WHERE n <= 5000");
+
+	const char* vacuum_variants[] = {
```
letFunny, line 1161: These are not used in the tests below, right?

test/unit/test_gateway.c
```diff
@@ -1060,6 +1060,122 @@ TEST_CASE(exec, undo_not_leader_pending_other_elected, NULL)
...
+	PREPARE("VACUUM");
+	f->request.db_id = 0;
+	f->request.stmt_id = stmt_id;
+	ENCODE(&f->request, exec);
+	HANDLE(EXEC);
+	WAIT;
+	ASSERT_CALLBACK(0, RESULT);
+	DECODE(&f->response, result);
+	munit_assert_int(f->response.last_insert_id, ==, 10000);
+	munit_assert_int(f->response.rows_affected, ==,   5000);
```
letFunny, line 1093: Q: Do we need any other check to ensure vacuum is working correctly? I was thinking about it like checking whether the contents were moved & compacted but I think that would be testing SQLite itself and not dqlite.

letFunny (commented):
Thanks for this Marco! Just a few comments and questions.

test/unit/test_gateway.c
letFunny: I am okay with both approaches. If it takes a lot of time then you can limit it to only check the syntax, if it is fast then we can leave it like this.

marco6:
Rebased on main so that tests can run to completion.

src/vfs.c
letFunny: > I think it's more like in the "we do not support that" sense.

Okay, that's what I meant. There is no way the flag could be set by using the public API, perfect.

src/vfs.c
letFunny: That makes sense, thanks for the explanation

marco6:
After extensive talking about the effect of a big transaction on the RAFT protocol, I think this is likely a decline.

While clearly this is not a "sqlite" issue and the logic there works, having big transactions is an issue.

The question that I leave here is the following: today we still have space for very big transactions (for example a `DROP TABLE` or an `ALTER TABLE` or an `UPDATE`) and it is up to the user not to do that or deal with the consequences. Should we follow that same route here saying "up to you, but if the compaction gets big then up to you)?

test/unit/test_gateway.c
```diff
@@ -1080,6 +1080,114 @@ TEST_CASE(exec, undo_not_leader_pending_other_elected, NULL)
...
+		"    SELECT 1 UNION ALL     "
+		"    SELECT n+1 FROM seq    "
+		"    WHERE  n < 10000       "
+		")                          "
+		"INSERT INTO test(n)        "
+		"SELECT n FROM seq          "
+	)
+	EXEC("DELETE FROM test WHERE n <= 5000");
+
+	PREPARE("VACUUM");
```
just-now, line 1104: Is it worth checking that VACUUM actually did something? 

The following queries might give us the picture of changes.
```
sqlite> PRAGMA freelist_count;
sqlite> PRAGMA page_count;
sqlite> PRAGMA page_size;
```

test/unit/test_gateway.c
```diff
@@ -1080,6 +1080,114 @@ TEST_CASE(exec, undo_not_leader_pending_other_elected, NULL)
...
+		WAIT;
+		ASSERT_CALLBACK(0, RESULT);
+		DECODE(&f->response, result);
+		munit_assert_int(f->response.last_insert_id, ==, 10000);
+		munit_assert_int(f->response.rows_affected, ==,   1000);
+	}
+	return MUNIT_OK;
+}
+
+/* Fail to vacume into a file */
```
just-now: nit: `s/vacume/vacuum/`

just-now (approved):
LGTM.

test/unit/test_gateway.c
marco6: Yes, makes sense.

PR #735, opened 2025-02-25
fix: initialize leader's queue

marco6:


marco6:
Not a real bug as the element inserted is single.

PR #738, opened 2025-03-10
fix: Prevent dqlite_server_stop() called multiple times

just-now:


marco6 (approved):
+1 on this not causing a crash.

PR #741, opened 2025-04-08
chore: update CLA check job

### LINKED ISSUE #739: Update CLA action to v2
You need to update action to v2. See https://github.com/canonical/has-signed-canonical-cla

marco6:
Resolves #739.

### DIFF
--- a/.github/workflows/cla-check.yml
+++ b/.github/workflows/cla-check.yml
@@ -1,13 +1,8 @@
-name: Canonical CLA
-
-on:
-  pull_request:
-    branches: [master]
+name: cla-check
+on: [pull_request]
 
 jobs:
   cla-check:
-    permissions:
-      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - name: Check if CLA signed
--- a/.github/workflows/cla-check.yml
+++ b/.github/workflows/cla-check.yml
@@ -2,7 +2,7 @@ name: Canonical CLA
 
 on:
   pull_request:
-    branches: [master, default]
+    branches: [master]
 
 jobs:
   cla-check:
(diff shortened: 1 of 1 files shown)


.github/workflows/cla-check.yml
```diff
@@ -1,11 +1,14 @@
 name: Canonical CLA
 
 on:
-  - pull_request
+  pull_request:
+    branches: [master, default]
```
letFunny, line 5: Does `default` work here? I was reading the documentation and I could not find a reference to using `default`, I only new about using the macro `$default-branch` ([example stackoverflow question](https://stackoverflow.com/questions/64781462/github-actions-default-branch-variable)).

.github/workflows/cla-check.yml
```diff
@@ -1,11 +1,14 @@
...
 
 jobs:
   cla-check:
-    runs-on: ubuntu-20.04
+    permissions:
+      pull-requests: write
```
letFunny, line 10: Is this needed? I don't see it in https://github.com/canonical/has-signed-canonical-cla.

letFunny (approved):
Thanks for this Marco! Minor comments below, approval is dependent on those.

.github/workflows/cla-check.yml
marco6: Ah TBH I copy-pasted this from another project. I'll remove that!

.github/workflows/cla-check.yml
marco6: Ok, I now copy-pasted the example. I think we want to check CLA also for other branches (like v1.17)

just-now (approved):
LGTM.

PR #742, opened 2025-04-24
perf: merge raft barrier requests

marco6:
This PR attempts to reduce the number of barriers written to the actual log by merging them together when possible (i.e. when there is an outstanding barrier already in place).

No changes to the API are required.


src/raft/client.c
```diff
@@ -74,53 +74,88 @@ int raft_apply(struct raft *r,
...
+	req->type = RAFT_BARRIER;
+	req->index = logLastIndex(r->log);
+	req->cb = cb;
+
+	/* Is the last entry an unreplicated barrier itself? 
+	 * If so, it is possible to merge the requests together
+	 * and avoid another roundtrip. */
+	struct raft_barrier *existing_req = (struct raft_barrier *)getRequest(r, req->index, RAFT_BARRIER);
+	if (existing_req != NULL) {
+		req->next = existing_req->next;
```
letFunny, line 121: I guess this line is needed for when there are 2 existing barriers and we are adding a third one. If that is the case we are missing a test that fails when removing this line.

test/unit/test_gateway.c
```diff
@@ -551,26 +551,22 @@ TEST_CASE(prepare, closing, NULL)
...
 	struct prepare_fixture *f = data;
-	uint64_t stmt_id;
 	(void)params;
 
 	/* Set up an uncommitted exec operation */
 	CLUSTER_ELECT(0);
-	PREPARE("CREATE TABLE test (n INT)");
-	EXEC_SUBMIT(stmt_id);
-	CLUSTER_DEPOSE;
-	ASSERT_CALLBACK(0, FAILURE);
```
letFunny, line 562: Why are we removing these lines from all the tests in this PR?

test/unit/test_gateway.c
```diff
@@ -551,26 +551,22 @@ TEST_CASE(prepare, closing, NULL)
...
-	/* Submit a prepare request, forcing a barrier, which fails */
-	CLUSTER_ELECT(0);
+	CLUSTER_APPLIED(2);
 	f->request.db_id = 0;
 	f->request.sql = "SELECT n FROM test";
 	ENCODE(&f->request, prepare);
 	/* We rely on leader_barrier_v2 (called by handle_prepare) attempting
 	 * an allocation using raft_malloc. */
+	/* TODO this is hacky, but I can't seem to hit the codepath otherwise */
+	raft_fixture_get(&f->cluster, 0)->last_applied--;
```
letFunny, line 565: Maybe we can achieve the same results by using the lower level macros to control the execution, e.g. instead of calling `PREPARE` that calls `WAIT` internally we call `ENCODE` and `HANDLE`.

If that is not feasible or it is flaky (which is my hunch), I am okay leaving it like this but with a better comment, something like:
```c
/* Test that several barriers are merged together by unapplying the last entry which is a a barrier. */
```

I also fail to see how this is testing that merging barriers work, I would assume we would need to check the log index to see that we are only replicating one.

test/unit/test_gateway.c
```diff
@@ -551,26 +551,22 @@ TEST_CASE(prepare, closing, NULL)
...
-	ASSERT_CALLBACK(0, FAILURE);
-
-	/* Submit a prepare request, forcing a barrier, which fails */
-	CLUSTER_ELECT(0);
+	CLUSTER_APPLIED(2);
 	f->request.db_id = 0;
 	f->request.sql = "SELECT n FROM test";
 	ENCODE(&f->request, prepare);
 	/* We rely on leader_barrier_v2 (called by handle_prepare) attempting
 	 * an allocation using raft_malloc. */
```
letFunny, line 563: This comment applies to lines 566 and 567, let's move it to be only above those lines. The new logic to unapply is not related to the comment and is a bit confusing to read about allocations before it.

test/unit/test_gateway.c
```diff
@@ -1855,25 +1851,21 @@ TEST_CASE(exec_sql, closing, NULL)
...
-	/* Submit an EXEC_SQL request, forcing a barrier, which fails */
-	CLUSTER_ELECT(0);
+	CLUSTER_APPLIED(2);
 	f->request.db_id = 0;
 	f->request.sql = "INSERT INTO test VALUES(123)";
 	ENCODE(&f->request, exec_sql);
-	/* We rely on leader_barrier_v2 (called by handle_exec_sql) attempting
+	/* We rely on leader_barrier_v2 (called by handle_prepare) attempting
 	 * an allocation using raft_malloc. */
+	/* TODO this is hacky, but I can't seem to hit the codepath otherwise */
```
letFunny, line 1863: Same two conversations. Let's discuss them in the other comments.

test/unit/test_gateway.c
```diff
@@ -2229,22 +2221,17 @@ TEST_CASE(query_sql, manyClosing, NULL)
...
-
 	/* Submit a QUERY_SQL request, forcing a barrier, which fails */
-	CLUSTER_ELECT(0);
+	CLUSTER_APPLIED(2);
 	f->request.db_id = 0;
 	f->request.sql = "SELECT n FROM test";
 	ENCODE(&f->request, query_sql);
 	/* We rely on leader_barrier_v2 (called by handle_query_sql) attempting
 	 * an allocation using raft_malloc. */
+	/* TODO this is hacky, but I can't seem to hit the codepath otherwise */
```
letFunny, line 2233: Same.

letFunny (commented):
This is looking great. I just had a few questions but consider it almost-approved from my side, the only thing I want to discuss is testing.

src/raft/client.c
```diff
@@ -74,53 +74,88 @@ int raft_apply(struct raft *r,
 	return rv;
 }
 
+/* Get the request matching the given @index and @type, if any.
+ * The type check is skipped when @type == -1. */
+static struct request *getRequest(struct raft *r,
```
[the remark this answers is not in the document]
marco6: Done.

src/raft/client.c
marco6: Done.

test/unit/test_gateway.c
marco6: I changed the way these test work so that they don't rely any more on a failed allocation (which was weird and fragile to begin with).

These tests now make the *raft append* related to the barrier fail.

I also added the `barrier_error` test for `query` and `exec` which somehow were forgotten.

src/raft/client.c
```diff
@@ -74,53 +74,85 @@ int raft_apply(struct raft *r,
 	return rv;
 }
 
+/* Get the request matching the given @index and @type, if any.
+ * The type check is skipped when @type == -1. */
```
letFunny: The comment is outdated now :)

src/raft.h
```diff
@@ -1288,6 +1288,7 @@ struct raft_barrier
 {
 	RAFT__REQUEST;
 	raft_barrier_cb cb;
+	struct raft_barrier *next;
```
letFunny, line 1294: Can we add a comment here to say that we are using this field as a linked list of barriers that are replicated together to avoid unnecessary round-trips?

src/raft/client.c
```diff
@@ -74,53 +74,85 @@ int raft_apply(struct raft *r,
...
+	req->type = RAFT_BARRIER;
+	req->index = logLastIndex(r->log);
+	req->cb = cb;
+
+	/* Is the last entry an unreplicated barrier itself?
+	 * If so, it is possible to merge the requests together
+	 * and avoid another roundtrip. */
+	struct request *existing_req = getRequest(r, req->index);
+	if (existing_req != NULL && existing_req->type == RAFT_BARRIER) {
+		struct raft_barrier *barrier_req =
```
letFunny:
```suggestion
		struct raft_barrier *existing_barrier =
```

src/raft/client.c
```diff
@@ -74,53 +74,85 @@ int raft_apply(struct raft *r,
...
+		    (struct raft_barrier *)existing_req;
+		req->next = barrier_req->next;
+		barrier_req->next = req;
+		return RAFT_OK;
+	}
+
+	req->index++;
+	req->next = NULL;
+	
+
```
letFunny, line 123: Nit: extra newline.

test/raft/integration/test_barrier.c
```diff
@@ -37,58 +37,57 @@ static void tearDown(void *data)
...
+    int _rv = raft_barrier(CLUSTER_RAFT(0), &req, barrierCbAssertResult);
+    munit_assert_int(_rv, ==, 0);
+    CLUSTER_STEP_UNTIL(barrierDone, &(result), 2000);
+
+    return MUNIT_OK;
+}
+
+TEST(raft_barrier, multiple, setUp, tearDown, 0, NULL)
+{
+	struct fixture *f = data;
```
letFunny, line 106: Tabs and spaces here.

test/raft/integration/test_barrier.c
```diff
@@ -37,58 +37,57 @@ static void tearDown(void *data)
...
+    struct raft_barrier reqs[REQ_N] = {};
+    for (int i = 0; i < REQ_N; i++) {
+        reqs[i] = (struct raft_barrier) {
+            .data = &result,
+        };
+        result.pending++;
+        int rv = raft_barrier(CLUSTER_RAFT(0), &reqs[i], barrierCbAssertResult);
+        munit_assert_int(rv, ==, 0);
+        munit_assert_int(result.pending, ==, i+1);
+    }
```
letFunny: I wonder if this is enough. I think this wouldn't catch the error in replication.c of doing:
```c
	cb = req->cb; // Note here we are only invoking the first cb.
	while (req != NULL) {
		struct raft_barrier *next = req->next;
		if (cb != NULL) {
			cb(req, 0);
		}
		req = next;
	}
```

letFunny: Options:
* We can store individual results in each barrier and wait for all of them (or the last one) and then check all of them are called.
* Clear the callback from `req` when `barrierCbAssertResult` is called.
* Something more sophisticated.

letFunny (approved):
Thanks for this, Marco. I see you addressed all comments. Approved pending the minor review comments I added in this round.

src/raft/client.c
marco6: I removed the comment.

src/raft.h
marco6: Done.

test/raft/integration/test_barrier.c
marco6: I re-formatted the whole file as it was mixed already, so the editor got confused. I'm basically taking ownership of the file per git blame, but I think that's ok as all tests in it were touched.

test/raft/integration/test_barrier.c
marco6: Discussed on MM. Now there is a test for multiple callbacks.

src/raft/client.c
```diff
@@ -74,53 +74,82 @@ int raft_apply(struct raft *r,
 	return rv;
 }
 
+static struct request *getRequest(struct raft *r, const raft_index index)
```
just-now: nit: `const` `struct raft *r`.

src/raft/replication.c
```diff
@@ -518,8 +518,12 @@ static void appendLeaderCb(struct raft_io_append *append, int status)
 				case RAFT_BARRIER: {
 					struct raft_barrier *barrier =
 					    (struct raft_barrier *)req;
-					if (barrier->cb) {
-						barrier->cb(barrier, status);
+					while (barrier != NULL) {
+						struct raft_barrier *next = barrier->next;
+						if (barrier->cb != NULL) {
+							barrier->cb(barrier, status);
```
just-now, line 524: UPD: ~~It's not clear how the failure path is handled. For example, one of the barriers fails and others are still in the queue. status variable gets rewritten every time we call `cb()`. If the first of two barriers fails, how would we report the failure?~~

just-now (commented):
Code itself is good. 

Yet, it's not clear what happens when one of the barrier fails.

just-now (approved):

PR #743, opened 2025-04-24
fix: reduce trace stack usage

marco6:
This PR wraps the logic to write traces in a noinline function so that stack message will not take space in the stack frame of the caller.

Related: #679. I'm really not sure if it is enough to fix it though.

### DIFF
--- a/src/tracing.c
+++ b/src/tracing.c
@@ -96,3 +96,15 @@ void stderrTracerEmit(const char *file,
 	if (level >= tracer__level)
 		tracerEmit(file, line, func, level, message);
 }
+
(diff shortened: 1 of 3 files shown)


letFunny:
Makes sense, do you have any measurement about consumption for #679 or is this about a general improvement that might also help that issue?

marco6:
> Makes sense, do you have any measurement about consumption for #679 or is this about a general improvement that might also help that issue?

This arose in the `new-exec` as on older versions of GCC the stack was growing too much and making some unit test fail. I am not sure this is the only issue for #679, we should check.

I am sure this helps though.

letFunny (approved):
Thanks Marco for clarification, I would also like to hear if @just-now has something to say from his experience

PR #744, opened 2025-04-24
fix: free shallow snapshot headers correctly

marco6:
This PR changes a bit the logic to free snapshot headers as the stress test in #740 showed a bug: while it is true that pages are never freed and can be shallowly referenced, the size of the database might change as writes are still allowed to continue during snapshot. This results in freeing the wrong buffer as the size of the database is used to "skip" buffers during finalize time.

The solution here uses the information stored *in the header* for each database instead.

### DIFF
--- a/src/fsm.c
+++ b/src/fsm.c
@@ -7,6 +7,7 @@
 #include "tracing.h"
 #include "vfs.h"
 
+#include <assert.h>
 #include <sys/mman.h>
 
 struct fsm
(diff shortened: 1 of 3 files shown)


letFunny:
Is there a simple way to include a test here or something that verifies the new behavior? If this bug was found by a stress test it might be very hard to replicate the full E2E behavior but I wonder if we can reproduce it in a simpler way.

just-now (approved):
Looks good.

PR #745, opened 2025-04-28
chore: drop support for 20.04

marco6:
Given that Ubuntu Focal Fossa (20.04) is EOL and jobs cannot run anymore this PR removes support for it in v1.18.

v1.17 can still support that version.


just-now (approved):
LGTM.

PR #746, opened 2025-04-28
chore: rename float_t to real_t

marco6:
This PR renames the `float_t` type to `real_t` so that it doesn't clash with `float_t` defined in `math.h`.

While the build issue is caused by libuv now importing `math.h` in their header (?!) this was a mistake since the start: a program that redefines a standard type is ill-formed.

I chose `real_t` to match SQL types listed in the SQLite manual at https://www.sqlite.org/datatype3.html


letFunny (approved):
Thanks!

just-now (approved):
That's the reason why it's good to have prefixes for all data structs in the project.
LGTM.

Issue #481, opened 2023-03-03
Moving sqlite3_step and other database operations off the main thread


cole-miller:
Right now, dqlite runs sqlite3_step and other database operations on the libuv main thread. This is problematic for disk mode, because such operations can take a long time when they need to write to disk, so that the event loop is blocked from handling other tasks (like those associated with incoming Raft messages). We also have some data from canonical/microk8s#3227 indicating that calls to sqlite3_step in particular frequently run for milliseconds to tens of milliseconds even for an an in-memory database, which is long enough to potentially cause problems when run on the main thread. We should fix our code to run sqlite3_step and perhaps other functions on the libuv thread pool, or in a dedicated thread.

freeekanayaka:
Ok, everything I'm going to say starts from the assumption that this issue is caused by:

1) disk-mode being turned on in microk8s
2) kine implementation being inefficient at emulating etcd in terms of SQL

If either 1) or 2) is wrong, please let me know and probably what I'm going to say can be discarded and reading it just a waste of time :)

So, it feels that the concerns that we had raised when discussing disk-mode are actually realizing, I'd encourage everybody to re-read https://github.com/canonical/dqlite/issues/368 carefully, because I believe it mentions pretty much everything, from the issues that could happen with the "cheap" approach of disk-mode that has been put in place (i.e. simply storing the database on disk and using blocking I/O to access it) to the reasons why just moving `sqlite3_step` (or any other part of the call stack) to a thread won't work per se.

If you bear with letting me to take again a step back, and talk from a high-level architectural/design point of view, I'll just re-iterate what I've said since the beginning of the microk8s/kine/dqlite idea years ago: 

- SQL is intrinsically not a great way to emulate a watchable and linearizable key/value store like etcd
- kine's implementation of such emulation is **probably** very inefficient in its own way and might need a more careful and clever approach to work better (for example I suspect there is away to avoid needing to store so many gigabytes of data into SQL tables and just store some metadata)
- dqlite should be used for different kind of workloads, because the choice of it being in-memory is dictated by precise technical circumstances beyond our control (in short, the fact that SQLite's VFS interface is synchronous, see below).

I know I'm beating a dead horse, since years now, and apologies for the directness but unfortunately I feel things have been kind of put under the carpet and duck-taped all along, and eventually reality knocks the doors. I'm well aware that all the developers involved have little choice, since this all comes from management. But probably management never fully realized the technical ramifications of this endeavor.

[...]

PR #747, opened 2025-04-28
feat: enable multithreaded execution

marco6:
This PR offloads the *execution* of SQLite statements (i.e. the `sqlite3_step` call) in the thread pool.

As usual with SQLite, this is only useful for read loads as SQLite can have as many readers as the machine can handle, but only one writer.

This PR is actually split in many semantic commits as the work to do that required some refactoring of the existing VFS. Only the "original" VFS is supported both in disk mode and in memory mode.

## The idea

SQLite is already great at synchronizing the computation among different threads and processes. However, the synchronization protocol, while not difficult to explain, contains in practice a lot of subtleties that must be taken into account.

Thinking about "what dqlite needs", it can be summarized with the following properties:
 - once a transaction is committed (from the PoV of SQLite), dqlite still needs some time to replicate the changes to the followers. In this period, readers should not be able to read data which is stable yet. This can be also thought as "make changes privately to the connection".
 - the write lock is not released so that other transactions cannot apply non-idemptent changes 
 - once a transaction is applied (in the RAFT sense), then connection should be able to read that (even if it wasn't originated by this replica).

The original way of achieving that was to:
 - take backe the write lock when polled
 - edit part of the shared memory to "hide" the transaction
 - "restart" (= corrupt) the shared memory when appending new entries, so that SQLite would try to rebuild it from scratch (which is BTW an expensive operaiton as it reads the WAL entirely).
 
Both the write lock and the editing generate a contention point. While write lock can be easily overcome int the VFS layer (by just not releasing the lock when asked), the editing part using normal C semantics cannot be solved as readers are allowed to read the transaction even when the write lock is held (i.e. the commit is visible not when the lock is released, but when the shared memory is updated!).

A way out would be to make each connection have its own shared memory. In this way each change will never be seen by anyone. The problem with that solution is that SQLite expects the VFS to hand out a pointer to the shared memory (which is then managed by SQLite without ever reaching the VFS again) and that pointer contains 2 main segments:
 - the header: written also by readers and used to orchestrate connections
 - the hash map: a hash map containing pointers to the segments. This could in theory be hidden.

These two requirements are conflicting with each other and cannot cohegsist in general.

As typical in logic domain though, two conflicting senteces can be both true if another variable is added: time. If we could switch that memory segment between "shared" (= changes are visible to all connections) to "private" (changes are not propagated) without SQLite knowing about that, then it could work.

### Memory is a lie

While in general when thinking about a pointer one pictures in his head an address to a region in memory, this is not really what happens with virtual memory on modern hardware. An address is actually translated to the phyisical address thanks to a table set up by the OS and the physical location might even move during the lifetime of a process.

That said it seems clear that if it is possible to somehow "remap" those addresses to back and forth a private/shared mapping then it would be possible to have both requirement fulfilled.

The idea is basically that each connection will have their shared memory pointing to the same physical pages by default like:

```mermaid
flowchart BT
    shared["shared memory"]
    reader1["reader"] --> shared
    reader2["reader"] --> shared
    reader3["reader"] --> shared
    reader4["reader"] --> shared
```

However when a connection becomes a writer (i.e. takes the write lock), memory is re-allocated and the *existing pointer* is remapped to a different *physical location*.

```mermaid
flowchart BT
    shared["shared memory"]
    reader1["reader"] --> shared
    reader2["reader"] --> shared
    reader3["reader"] --> shared
    private["private memory"]
    writer["writer"] --> private
    writer["writer"] -.-> shared

    linkStyle 4 stroke:red;
    linkStyle 3 stroke:green
```

On linux the above can be achieved with an in-memory file and the two syscalls `mmap` and `mremap`.

The technique used in this PR is strictly *Linux only* and it is not porable on other systems AFAIK. If anyone has ideas on how to support that on MacOS or Windows or whatever other system, please share.

The following sections explain the changes in this PR splitting them in phases and sections.

### Separation of concerns in the VFS

The first few commits are about splitting the behavior of each file in its own datastructure. This effort started already on #734 where the temporary file was returned "naked" from the underlying VFS instead of being wrapped.

In particulare, the once lone `struct vfsFile` is split in 3 separate structs, depending on the scope:
 - a `vfsNoopFile` used for journals (as they are necessary to enter WAL mode, but functionally useless)
 - a `vfsWalFile` used for in-memory WAL files
 - a `vfsMainFile` used for in-memory database files (the default)
 - a `vfsDiskMainFile` used for disk-mode database files.

This separation made the code longer, but simplified the logic in each case as the routines only need to care about the specific behavior of the file handled.

### File-backed shared memory

We need a file descriptor to correctly point to shared memory. A commit takes care of switching from an `malloc`-based shared memory to a `mmap` based one. Thie file used is an in-memory file (see `memfd_create`). This commit does not make the transactions private yet, but only deals with backing the shared memory with an mmapped file without breaking the system.

### Private maps

The next phase is to still keep everything in a single thread, but make sure changes to the shared memory are done privately by a write transaction. Please note that this only works because the write lock does not write to the index part related to the access protocol after taking the write lock.

Once the lock is released, this PR also "publishes" changes done to the private patch to the shared memory, but only during a `VfsApply`. `VfsAbort` discardes changes instead. This only works because there is a single writer that already took the highes read mark possible.

### Synchronizing access

The last bit for the VFS is to make sure that all its structures can be accessed concurrently. This is usually done by the OS (for example, locking a segment in the file is an atomic operation - we need to make sure it is for our VFS as well).

While there probably exists some lock-free schemes possible, I opted for a C11 mutex as its sematics guarantees acquire-release semantics on memory written while holding the lock.

### Using the thread pool

The last bit of this work is to take the work callbacks for the `exec` state machine and move them in the thread pool. This is easy given the redesign that already happened on #740.

The rest of this PR is adapting white-box tests to the new structure.


marco6:
This PR is currently being split in many smaller PRs as a lot of this code was preparatory refactoring which is not 100% related to multithreading and is good in general on its own.

marco6:
This will also resolve #481. This *does not* mean however that disk mode is a good idea still: the `sqlite3_prepare_v2` call clearly needs to access the disk (to read the schema) and that might still create issues (i.e. the idea that moving `sqlite3_step` to another thread is enough is wrong).

src/vfs.c
```diff
@@ -1540,8 +1477,118 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+}
+
+static int vfsRedirectShm(struct vfsMainFile *f) {
+	for (int regionIndex = 0; regionIndex < f->mappedShmRegions.len; regionIndex++) {
+		void *region = f->mappedShmRegions.ptr[regionIndex];
+		void *new_region = mmap(NULL, VFS__WAL_INDEX_REGION_SIZE,
+		PROT_READ | PROT_WRITE, MAP_PRIVATE, f->database->shm.fd, regionIndex * VFS__WAL_INDEX_REGION_SIZE);
+		if (new_region == MAP_FAILED) {
+			/* This should never happen. Also, this means that we might leave the 
+			* connection in a werid state. */
```
letFunny:
```suggestion
			* connection in a weird state. */
```

src/vfs.c
```diff
@@ -1544,6 +1479,111 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
 	w->n_tx = 0;
 }
 
+static int vfsRedirectShm(struct vfsMainFile *f) {
```
letFunny, line 1580: The two functions need a little bit of documentation

src/vfs.c
```diff
@@ -1544,6 +1479,111 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+	for (int regionIndex = 0; regionIndex < f->mappedShmRegions.len; regionIndex++) {
+		void *region = f->mappedShmRegions.ptr[regionIndex];
+		void *new_region = mmap(NULL, VFS__WAL_INDEX_REGION_SIZE,
+		PROT_READ | PROT_WRITE, MAP_PRIVATE, f->database->shm.fd, regionIndex * VFS__WAL_INDEX_REGION_SIZE);
+		if (new_region == MAP_FAILED) {
+			/* This should never happen. Also, this means that we might leave the 
+			* connection in a werid state. */
+			return SQLITE_NOMEM;
+		}
+		void *remapped = mremap(new_region,
```
letFunny, line 1593: Somewhere we need to record why we do this, the fact that the sqlite connection contains a pointer to the region that we cannot change so what we are doing instead of to change the mapping in the virtual memory so, even if the pointer is fixed, we can make the changes private when a writing transaction is running and then publish them.

src/vfs.c
```diff
@@ -1544,6 +1479,111 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+	if (f->mappedShmRegions.len == 0) {
+		return SQLITE_OK;
+	}
+
+	/* Copy the hash map array in all the secondary regions */
+	for (int i = 1; i < f->mappedShmRegions.len; i++) {
+		void *region = mmap(NULL, VFS__WAL_INDEX_REGION_SIZE,
+			PROT_READ | PROT_WRITE, MAP_SHARED, f->database->shm.fd, i * VFS__WAL_INDEX_REGION_SIZE);
+		assert(region != MAP_FAILED);
+		memcpy(region, f->mappedShmRegions.ptr[i], VFS__WAL_INDEX_REGION_SIZE);
```
letFunny, line 1629: I guess this is implementation dependent but this is almost surely not atomic. Do we have any problem where the connection might see a partially updated hash table?

EDIT: outdated, I wrote this comment based only on the commit before adding synchronization.

src/vfs.c
```diff
@@ -1544,6 +1479,111 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+		void *remapped = mremap(region, 
+			VFS__WAL_INDEX_REGION_SIZE, VFS__WAL_INDEX_REGION_SIZE, 
+			MREMAP_MAYMOVE | MREMAP_FIXED, f->mappedShmRegions.ptr[i]);
+		assert(remapped == f->mappedShmRegions.ptr[i]);
+	}
+
+	const size_t headerSize = VFS__WAL_INDEX_HEADER_SIZE * 2 + 40;
```
letFunny: Where is the `40` coming from? I would create a constant for documentation or write a comment, whatever makes more sense. At the same time `headerSize` is a bit of a weird name because it is actually `2 * headerSize + something`.

From the code below it seems to be `hashTableOffset`.

src/vfs.c
```diff
@@ -1544,6 +1479,111 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+	 * only be started after a read transaction and as such the marker and the read lock is
+	 * already acquired.
+	 * For the other fields, quoting the documentation:
+	 *  > The nBackfill can only be increased while holding the WAL_CKPT_LOCK. However, nBackfill
+	 *  > is changed to zero during a WAL reset, and this happens while holding the WAL_WRITE_LOCK. 
+	 * And also:
+	 *  > The mxFrame value is always greater than or equal to both nBackfill and nBackfillAttempted.
+	 * Which means that only when holding the checkpoint lock those fields should be published */
+	if (f->exclMask & (1 << VFS__WAL_CKPT_LOCK)) {
+		const size_t nBackfillOffset          =  96;
```
letFunny, line 1667: Nice to see the constant!

src/vfs.c
```diff
@@ -1544,6 +1479,111 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+		int32_t *privateBackfill              = (first_region_private + nBackfillOffset);
+		int32_t *sharedBackfill               = (first_region_shared + nBackfillOffset);
+		*sharedBackfill                       = *privateBackfill;
```
letFunny: Can we update both fields one after the other?

src/vfs.c
```diff
@@ -1540,8 +1477,118 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+	return SQLITE_OK;
+}
+
+static int vfsRollbackShm(struct vfsMainFile *f) {
+	for (int i = f->mappedShmRegions.len - 1; i >= 0; i--) {
+		void *new_region = mmap(NULL, VFS__WAL_INDEX_REGION_SIZE,
+			PROT_READ | PROT_WRITE, MAP_SHARED, f->database->shm.fd, i * VFS__WAL_INDEX_REGION_SIZE);
+		if (new_region == MAP_FAILED) {
+			/* This should never happen. Also, this means that we might leave the 
+				* connection in a werid state. */
```
letFunny:
```suggestion
				* connection in a weird state. */
```

src/vfs.c
```diff
@@ -1554,7 +1594,6 @@ static int vfsMainFileShmLock(sqlite3_file *file, int ofst, int n, int flags)
 	/* Legal values for the offset and the range */
 	PRE(n > 0);
 	PRE(ofst >= 0 && ofst + n <= SQLITE_SHM_NLOCK);
-	PRE(n == 1 || (flags & SQLITE_SHM_EXCLUSIVE) != 0);
```
letFunny, line 1560: Why? I thought this was an assumption about sqlite usage, not about our own usage.

src/vfs.c
```diff
@@ -1626,12 +1661,25 @@ static int vfsMainFileShmLock(sqlite3_file *file, int ofst, int n, int flags)
...
 			}
 		}
 	}
+
+	if (rv == SQLITE_OK && ofst == VFS__WAL_WRITE_LOCK && n == 1 && flags & SQLITE_SHM_EXCLUSIVE) {
+		if (flags & SQLITE_SHM_LOCK) {
+			/* When taking a write lock also make all mappings private. 
+			 * This effectively shadows the transaction as all other
+			 * connections cannot see the update to the mxFrames field. */
+			rv = vfsRedirectShm(f);
```
letFunny: What about `vfsHideShm` and `vfsPublishShm`?

src/vfs.c
```diff
@@ -1626,12 +1661,25 @@ static int vfsMainFileShmLock(sqlite3_file *file, int ofst, int n, int flags)
...
+		} else {
+			/* Releasing memory means "publishing" changes back to the
+			 * shared map and remove the COW behavior from the regions. */
+			rv = vfsCommitShm(f);
```
letFunny: I think we should make a note here about ROLLBACKs because when I was reading it I was expecting to see the rollback handled here, not calling `vfsShmUnlock` directly and bypassing the "public function logic".

src/vfs.c
```diff
@@ -2292,16 +2342,6 @@ int VfsPoll(sqlite3 *conn, struct vfsTransaction *transaction)
 	f->database->wal.n_tx = 0;
 	f->database->wal.tx = NULL;
 
-	/* If some frames have been written take the write lock. */
```
letFunny, line 2298: Sanity-check for me: we no longer have to take the write lock because we are not releasing it to being with, so there is no window when the lock is not taken.

src/vfs.c
```diff
@@ -1089,44 +1052,60 @@ static int vfsWalFileWrite(sqlite3_file* file, const void* buf, int amount, sqli
...
 	}
 
-	page_size = vfsWalGetPageSize(f->wal);
+	/* There is no need to lock here are only one writer is alloed to
+	 * ever change the wal and this routine is that writer. */
+	page_size = vfsWalGetPageSize(&f->database->wal);
 	assert(page_size > 0);
 
+	/* It is expected that once committed the WAL cannot be changed. 
+	 * As such, it's index must be above the already committed part */
```
letFunny: `its`

src/vfs.c
```diff
@@ -1089,44 +1052,60 @@ static int vfsWalFileWrite(sqlite3_file* file, const void* buf, int amount, sqli
...
+	index = (unsigned)formatWalCalcFrameIndex((int)page_size, offset);
+	if (index <= f->database->wal.n_frames) {
+		return SQLITE_IOERR_WRITE;
+	}
```
letFunny, line 1107: Why not an assert?

src/vfs.c
```diff
@@ -1089,44 +1052,60 @@ static int vfsWalFileWrite(sqlite3_file* file, const void* buf, int amount, sqli
...
+		assert(tx_index == f->database->wal.n_tx + 1);
+		/* Also, new frames always start by writing the header first */
+		assert(amount == FORMAT__WAL_FRAME_HDR_SIZE);
+
+		frame = vfsFrameCreate(page_size);
+		if (frame == NULL) {
+			return SQLITE_NOMEM;
+		}
+		unsigned new_n_tx = f->database->wal.n_tx + 1;
+		struct vfsFrame **new_tx = sqlite3_realloc(f->database->wal.tx, (int)(sizeof(*new_tx) * new_n_tx));
```
letFunny, line 1123: So for every frame that a write transaction creates we are reallocating all the pages and _possibly_ moving them in memory?

src/vfs.c
```diff
@@ -1089,44 +1052,60 @@ static int vfsWalFileWrite(sqlite3_file* file, const void* buf, int amount, sqli
...
+		if (new_tx == NULL) {
+			return SQLITE_NOMEM;
+		}
+		new_tx[f->database->wal.n_tx] = frame;
+		f->database->wal.tx = new_tx;
+		f->database->wal.n_tx = new_n_tx;
+	}
+
+	frame = f->database->wal.tx[tx_index - 1];
```
letFunny, line 1132: It is a bit weird that the index is not really an index and you have to substract `-1`. We could either change it to be a frame number or, alternatively, I suggest to substract `-1` directly in definition and then you can remove the +1 here: `f->database->wal.n_tx + 1`.

src/vfs.c
```diff
@@ -1143,12 +1122,14 @@ static int vfsWalFileTruncate(sqlite3_file* file, sqlite3_int64 size)
 		return SQLITE_PROTOCOL;
 	}
 
-	formatWalRestartHeader(f->wal->hdr);
-	vfsWalClose(f->wal);
-	f->wal->frames = NULL;
-	f->wal->n_frames = 0;
-	f->wal->tx = NULL;
-	f->wal->n_tx = 0;
+	/* No readers can ever touch this database. Also not VfsAppend can
```
letFunny: `not` -> `no`

src/vfs.c
```diff
@@ -1348,17 +1337,30 @@ static int vfsFileControlPragma(struct vfsMainFile *f, char **fcntl)
...
+			    sqlite3_mprintf("page size must be a power of 2");
+			return SQLITE_IOERR;
+		}
+
+		mtx_lock(&f->database->lock);
+		if (f->database->n_pages > 0) {
+			int db_page_size =
+			    (int)vfsDatabaseGetPageSize(f->database);
+			if (db_page_size != page_size) {
 				fcntl[0] = sqlite3_mprintf(
```
letFunny, line 1442: Shouldn't we unlock the locks also on error conditions? We don't know how the upper layer will handle it and if it might recover from it and continue. This is a general question, the code highlighted is only one example.

src/vfs.c
```diff
@@ -1547,17 +1556,15 @@ static int vfsCommitShm(struct vfsMainFile *f) {
...
 		int32_t *privateBackfill              = (first_region_private + nBackfillOffset);
-		int32_t *sharedBackfill               = (first_region_shared + nBackfillOffset);
-		*sharedBackfill                       = *privateBackfill;
+		_Atomic int32_t *sharedBackfill       = (first_region_shared + nBackfillOffset);
+		atomic_store_explicit(sharedBackfill, *privateBackfill, memory_order_relaxed);
 
 		const size_t nBackfillAttemptedOffset = 128;
 		int32_t *privateBackfillAttempted     = (first_region_private + nBackfillAttemptedOffset);
-		int32_t* sharedBackfillAttempted      = (first_region_shared + nBackfillAttemptedOffset);
+		int32_t *sharedBackfillAttempted      = (first_region_shared + nBackfillAttemptedOffset);
```
letFunny, line 1677: Why is this one not atomic?

src/vfs.c
```diff
@@ -1533,6 +1536,12 @@ static int vfsCommitShm(struct vfsMainFile *f) {
 
 	/* Copy the hash map array for the first region */
 	memcpy(first_region_shared + headerSize, first_region_private + headerSize, VFS__WAL_INDEX_REGION_SIZE - headerSize);
+	atomic_thread_fence(memory_order_relaxed);
+
+	/* Finally publish the two copies of the WAL Index information and remap. */
```
letFunny, line 1648: This comment is outdated. Also why do we update the headers first now.

src/vfs.c
```diff
@@ -1533,6 +1536,12 @@ static int vfsCommitShm(struct vfsMainFile *f) {
...
+	memcpy(first_region_shared + VFS__WAL_INDEX_HEADER_SIZE, f->mappedShmRegions.ptr[0] + VFS__WAL_INDEX_HEADER_SIZE, VFS__WAL_INDEX_HEADER_SIZE);
+	atomic_thread_fence(memory_order_seq_cst);
+	memcpy(first_region_shared, f->mappedShmRegions.ptr[0], VFS__WAL_INDEX_HEADER_SIZE);
```
letFunny, line 1652: Why do you copy them in two memcpy across the fence? This guarantees that the second header is seen but the first one might not be seen by the other barrier (assuming I got it right). If someone reads the shared memory and the headers do not match, isn't it treated as corrupt?

src/vfs.c
```diff
@@ -2193,7 +2206,11 @@ static int vfsSleep(sqlite3_vfs *vfs, int microseconds)
 {
 	(void)vfs;
 
-	/* TODO (is this needed?) */
```
letFunny, line 2156: hahahaa :)

src/vfs.c
```diff
@@ -2193,7 +2206,11 @@ static int vfsSleep(sqlite3_vfs *vfs, int microseconds)
...
+	struct timespec sp;
+	sp.tv_sec = microseconds / 1000000;
+	sp.tv_nsec = (microseconds % 1000000) * 1000;
```
letFunny, line 2318: Isn't the conversion from micro to nano done by `/1000`?

src/vfs.c
```diff
@@ -64,6 +66,8 @@ const int vfsOne = 1;
 #define VFS__WAL_WRITE_LOCK 0
 #define VFS__WAL_CKPT_LOCK  1
 #define VFS__WAL_RECOVER_LOCK  2
+#define VFS__WAL_READ_LOCK(I) (3+(I))
```
just-now: Is it used somewhere?

src/vfs.c
```diff
@@ -448,27 +384,46 @@ static uint32_t vfsWalGetPageSize(struct vfsWal *w)
...
+	int64_t size = 0;
+	if (w->n_frames > 0) {
+		uint32_t page_size;
+		page_size = vfsWalGetPageSize(w);
+		size += VFS__WAL_HEADER_SIZE;
+		size += (int64_t)w->n_frames *
+			(int64_t)(FORMAT__WAL_FRAME_HDR_SIZE + page_size);
+	}
+	/* TODO dqlite is limited to a max database size of SIZE_MAX */
+	assert((size >= 0) && ((uint64_t)size <= SIZE_MAX));
```
just-now: nit: extra braces in `((size >= 0)...` are not needed. 
```
assert(size >= 0 && (uint64_t)size <= SIZE_MAX);
```
man operator

src/vfs.c
```diff
@@ -448,27 +384,46 @@ static uint32_t vfsWalGetPageSize(struct vfsWal *w)
...
 static void vfsWalClose(struct vfsWal *w)
 {
 	unsigned i;
 	for (i = 0; i < w->n_frames; i++) {
 		vfsFrameDestroy(w->frames[i]);
 	}
-	if (w->frames != NULL) {
-		sqlite3_free(w->frames);
-	}
+	sqlite3_free(w->frames);
```
just-now, line 436: Thanks (free(0)).

src/vfs.c
```diff
@@ -362,84 +366,16 @@ static struct vfsFrame *vfsWalFrameLookup(struct vfsWal *w, unsigned n)
 	assert(w != NULL);
 	assert(n > 0);
 
-	if (n > w->n_frames + w->n_tx) {
+	if (n > w->n_frames) {
```
just-now: In `vfsWalFrameLookup()`, could you add a comment with an expectation that database->lock is locked in it?
In previous live we had a special function `mutex_is_locked(mtx)` for invariant checks.

src/vfs.c
```diff
@@ -1150,44 +1054,60 @@ static int vfsWalFileWrite(sqlite3_file* file, const void* buf, int amount, sqli
...
 		assert(amount == VFS__WAL_HEADER_SIZE);
 
-		memcpy(f->wal->hdr, buf, (size_t)amount);
+		memcpy(f->database->wal.hdr, buf, (size_t)amount);
 		return SQLITE_OK;
 	}
 
-	page_size = vfsWalGetPageSize(f->wal);
+	/* There is no need to lock here are only one writer is alloed to
+	 * ever change the wal and this routine is that writer. */
```
just-now: There are a few mistakes in the comment.

src/vfs.c
```diff
@@ -1150,44 +1054,60 @@ static int vfsWalFileWrite(sqlite3_file* file, const void* buf, int amount, sqli
...
+		 * one page after the end. */
+		assert(tx_index == f->database->wal.n_tx + 1);
+		/* Also, new frames always start by writing the header first */
+		assert(amount == FORMAT__WAL_FRAME_HDR_SIZE);
+
+		frame = vfsFrameCreate(page_size);
+		if (frame == NULL) {
+			return SQLITE_NOMEM;
+		}
+		unsigned new_n_tx = f->database->wal.n_tx + 1;
```
just-now: [nit] You can put this declaration before ln. 1077 and update the assert().

src/gateway.c
```diff
@@ -700,8 +725,37 @@ static void handle_query_work_cb(struct exec *exec)
 		return;
 	}
 
-	leader_exec_result(exec, rc == SQLITE_DONE ? RAFT_OK : RAFT_ERROR);
-	TAIL return leader_exec_resume(exec);
+	leader_exec_result(exec, rc == SQLITE_DONE ? RAFT_OK :RAFT_ERROR);
```
just-now: nit: s/:RAFT_ERROR/: RAFT_ERROR/

src/gateway.c
```diff
@@ -700,8 +725,37 @@ static void handle_query_work_cb(struct exec *exec)
...
+	struct gateway *g = exec->data;
+	PRE(g->req != NULL && g->leader != NULL);
+	struct handle *req = g->req;
+
+	if (req->cancellation_requested) {
+		/* Nothing else to do. */
+		TAIL return leader_exec_resume(exec);
+	}
+
+	if (!sqlite3_statement_empty(exec->leader->conn, exec->tail)) {
```
just-now: It is not a comment to the current line.
`sqlite3_statement_empty()` pretends to be a part of `sqlite3` still it's defined by us. Is it possible to rename it so that nobody has an impression that it's `sqlite3_` namespace.

src/vfs.c
```diff
@@ -1108,7 +990,29 @@ static int vfsWalFileRead(sqlite3_file* file, void* buf, int amount, sqlite3_int
...
 	}
 
-	frame = vfsWalFrameLookup(f->wal, index);
+	/* If ever really needed, there exists lock-free algorithms. I think
+	 * that here contention will always be very small the only thing that
+	 * can be contended is the access to the wal page pointer and not
+	 * on the content: it is guaranteed by the sqlite3 WAL protocol that
+	 * the WAL will not be restarted or checkpointed if a READ lock is
+	 * held, which is always true when reading the WAL. */
+	mtx_lock(&f->database->lock);
```
just-now: Could you explain in a comment around the database lock itself, that this lock appeared because of an added source of parallelism and as some of vfs functions are called from new threads, uv thread needs a certain serialization with others.

src/vfs.c
```diff
@@ -1409,17 +1343,30 @@ static int vfsFileControlPragma(struct vfsMainFile *f, char **fcntl)
 		 */
 		int page_size = atoi(right);
 
-		if (page_size >= FORMAT__PAGE_SIZE_MIN &&
-		    page_size <= FORMAT__PAGE_SIZE_MAX &&
-		    ((page_size - 1) & page_size) == 0) {
-			if (f->database->n_pages > 0 &&
-			    page_size !=
-				(int)vfsDatabaseGetPageSize(f->database)) {
+		if (page_size < FORMAT__PAGE_SIZE_MIN ||
```
just-now, line 1425: For maintainability, shorter code is better:
`(page_size < FORMAT__PAGE_SIZE_MIN || page_size > FORMAT__PAGE_SIZE_MAX) && !is_po2((unsigned long)page_size)`

src/vfs.c
```diff
@@ -1626,22 +1687,33 @@ static int vfsMainFileShmLock(sqlite3_file *file, int ofst, int n, int flags)
...
 		if (rv == SQLITE_OK && ofst == VFS__WAL_WRITE_LOCK) {
-			assert(n == 1);
 			if (flags == (SQLITE_SHM_LOCK | SQLITE_SHM_EXCLUSIVE)) {
 				assert(f->database->wal.n_tx == 0);
 			}
 		}
 	}
+
+	if (rv == SQLITE_OK && ofst == VFS__WAL_WRITE_LOCK && n == 1 && flags & SQLITE_SHM_EXCLUSIVE) {
+		if (flags & SQLITE_SHM_LOCK) {
```
just-now: [nit]: It's better to put comments below into the function themselves and shorten the code
```
rc = flags & SQLITE_SHM_LOCK ? vfsRedirectShm(f) : vfsCommitShm(f);
```

just-now (commented):
I like the patch and it's clear to me. 
We need to phrase the expectation for some vfs functions that now they will be called under the database->lock.

src/vfs.c
marco6: Removed.

src/vfs.c
marco6: Done. Please note that the right hand of this check means nothing as any value is less than or equal the max representable value.

src/vfs.c
marco6: I added a lot of (rather long) comments on locking and data synchronization among threads. I hope that helps.

src/vfs.c

src/gateway.c
marco6: Renamed to `is_statement_empty`.

src/vfs.c
marco6: As I said above, I changed few things:
 - now each datastructure has its own mutex which makes things a bit cleaner.
 - I added comments explaining the reason something can proceed without a lock or needs a lock.
 
 Hope that helps.

src/vfs.c

src/vfs.c
marco6: The error string is different though. I think it's best to give a meaningful message in this case.

marco6:
We likely want to test this also on arm. I don't have a raspberry with me, but we can likely wait for #793 and rebase/merge on master.

src/vfs.c
```diff
@@ -1687,9 +1700,7 @@ static int vfsMainFileShmLock(sqlite3_file *file, int ofst, int n, int flags)
 static void vfsMainFileShmBarrier(sqlite3_file *file)
 {
 	(void)file;
-	/* This is a no-op since we expect SQLite to be compiled with mutex
-	 * support (i.e. SQLITE_MUTEX_OMIT or SQLITE_MUTEX_NOOP are *not*
-	 * defined, see sqliteInt.h). */
+	atomic_thread_fence(memory_order_seq_cst);
```
letFunny, line 1810: Maybe a comment around the fences to say which ones are supposed to synchronize would be helpful

src/gateway.c
```diff
@@ -655,39 +675,44 @@ static int handle_exec_sql(struct gateway *g, struct handle *req)
...
-		    tuple_decoder__remaining(&req->decoder) > 0) {
+		rv = bind__params(exec->stmt, &req->decoder);
+		if (rv != DQLITE_OK) {
+			tracef("handle exec bind failed %d", rv);
 			leader_exec_result(exec, RAFT_GATEWAY_PARSE);
-			TAIL return leader_exec_resume(exec);
+			return SQLITE_ERROR;
+		}
+		if (tuple_decoder__remaining(&req->decoder) > 0) {
+			leader_exec_result(exec, RAFT_GATEWAY_PARSE);
```
letFunny, line 693: Maybe a tracef would also be helpful if the error message is generic

test/unit/test_concurrency.c
```diff
@@ -372,11 +373,14 @@ TEST_CASE(exec, busy_wait_transaction_dropped, NULL)
 	/* start another write */
 	PREPARE(f->c2, "INSERT INTO test(n) VALUES(1)", &f->stmt_id2);
 	EXEC(f->c2, f->stmt_id2);
-	
+	munit_assert_false(f->c2->context.invoked);
```
letFunny, line 376: AFAIK is `ASSERT_CALLBACK` is not called then `invoked` is not cleared so a single assertion will also catch the mistake. Is the intention to know exactly when it happens, or is there another reason?

src/vfs.c
```diff
@@ -1612,6 +1618,13 @@ static int vfsMainFileShmLock(sqlite3_file *file, int ofst, int n, int flags)
 
 	PRE((f->exclMask & f->sharedMask) == 0);
 
+	/* In case a checkpoint is running, the locking logic is overridden, to
+	 * make sure nothing can happen concurrently. As such, it is just enough
+	 * to always return SQLITE_OK in this case. */
+	if (f->exclMask == VFS__CHECKPOINT_MASK) {
+		return SQLITE_OK;
+	}
```
letFunny, line 1737: Why do we need this? If the checkpoint is taking all the locks, the logic below should work as is, having this tells me something is not right.

For context, my understanding is that this function attempts to take the lock which is a no-op if it is already taken. If when checkpointing we take all the locks (or we should), then the regular logic should succeed intact.

src/vfs.c
```diff
@@ -1150,14 +1152,18 @@ static const sqlite3_io_methods vfsWalFileMethods = {
...
+
 /* Implementation of the abstract sqlite3_file base class.
  * for the main database file */
 struct vfsMainFile {
 	sqlite3_file base;            /* Base class. Must be first. */
 	struct vfs *vfs;              /* Pointer to volatile VFS data. */
 	struct vfsDatabase *database; /* Underlying database content. */
 	uint16_t sharedMask;          /* Mask of shared locks held */
-	uint16_t exclMask;            /* Mask of exclusive locks held */
+	/* Mask of exclusive locks held. The special value `CHECKPOINT_MASK` is
```
letFunny: CHECKPOINT_MASK -> VFS__CHECKPOINT_MASK

src/vfs.c
```diff
@@ -362,82 +374,29 @@ static struct vfsFrame *vfsWalFrameLookup(struct vfsWal *w, unsigned n)
...
-		if (tx == NULL) {
-			rv = SQLITE_NOMEM;
-			goto err_after_vfs_frame_create;
+		frame = NULL;
+	}
+	mtx_unlock(&w->mtx);
+
+	/* In case of a cache spill, it is possible that SQLite will try to
+	 * re-populate the cache reading from the uncommitted part of the WAL. 
+	 * unfortunately, it is not possible to reliably link a WAL file to
```
letFunny:
```suggestion
	 * Unfortunately, it is not possible to reliably link a WAL file to
```

src/vfs.c
```diff
@@ -362,82 +374,29 @@ static struct vfsFrame *vfsWalFrameLookup(struct vfsWal *w, unsigned n)
...
+	 * its main file and checking this way that this is the connection
+	 * holding the write lock. As such, the best we can do is to trust
+	 * SQLite. This is also the reason why this code is in an unprotected
```
letFunny, line 390: > As such, the best we can do is to trust SQLite.

Is there a guarantee or an expectation that SQLite will only re-populate the cache on a connection that holds the write lock? The text is not very clear in translating what we expect to happen.

src/vfs.c
```diff
@@ -362,82 +374,29 @@ static struct vfsFrame *vfsWalFrameLookup(struct vfsWal *w, unsigned n)
...
+	 * section (outside the protection of the mtx). */
```
letFunny, line 391: I thought this code was unprotected by the mutex because we are accessing the "private" `w->{n_tx,tx}` fields. Again, it might be that the comment is not clear and we are saying the same thing.

src/vfs.c
```diff
@@ -1251,6 +1212,35 @@ static int vfsMainFileRead(sqlite3_file *file,
...
+	 *   we get from SQLite is that pages accessed are never changed. While
+	 *   for a file that is enough, it is not enough for an array in memory
+	 *   that might be resized (as in realloc). So, while the content of
+	 *   each page in f->database->pages can be considered constant, the
+	 *   array holding each page might change. The resizing of this array
+	 *   and/or the overwriting of unread pages is always done during a
+	 *   checkpoint, holding the checkpoint lock. On top of these guarantees
+	 *   from SQLite, dqlite adds a more strict behaviour: only TRUNCATE
+	 *   checkpoint are allowed and they always block all readers (i.e. they
+	 *   hold all locks exclusively) and are always TRUNCATE checkpo. This
```
letFunny:
```suggestion
	 *   hold all locks exclusively). This
```

src/vfs.c
```diff
@@ -1140,6 +1077,9 @@ static int vfsWalFileRead(sqlite3_file* file, void* buf, int amount, sqlite3_int
 static int vfsWalFileWrite(sqlite3_file* file, const void* buf, int amount, sqlite3_int64 offset)
 {
 	struct vfsWalFile *f = (struct vfsWalFile *)file;
+	/* == Safety ==
+	 * See vfsWalFile read.
```
letFunny:
```suggestion
	 * See [vfsWalFileRead].
```
In general I am not sure about whether it's better to use `[identifier]` everywhere or not.

src/vfs.c
```diff
@@ -1251,6 +1212,35 @@ static int vfsMainFileRead(sqlite3_file *file,
...
+	 *   hold all locks exclusively) and are always TRUNCATE checkpo. This
+	 *   means that during this call it is not necessary to synchronize at
+	 *   all.
+	 *
+	 * Regarding reading values from memory shared across multiple threads,
+	 * it is guaranteed that after all checkpoints, a full memory barrier is
+	 * issued from the checkpoint thread (the libuv main thread). A full
+	 * memory barrier is also issued by the thead the transaction runs in
+	 * when starting a read transaction. This creates a thread-fence chain
+	 * that guarantees that no stale data is read after a checkpoint by read
```
letFunny, line 1240: Let me double-check that I understand this because I think I am missing something. The memory barrier in `vfsCommitShm` which is used to release the memory when a transaction unlocks the write lock. When you say here that "a full memory barrier is issued from checkpoint thread" it is because that thread belongs to a writer that has taken the lock as I say above and will release it triggering the barrier. If you say the full memory barrier is issued from the main libuv thread that is because releasing the lock calling the vfs method happens in the main thread when calling apply.

Is all of the above correct?

src/vfs.c
marco6: In SQLite code that 40-byte structure is referred as [`WalCkptInfo`](https://github.com/sqlite/sqlite/blob/master/src/wal.c#L394). I called the constant `ckptInfoSize`.

src/vfs.c
marco6: That is true, the assumption about SQLite usage is correct. However, we are taking *all locks* (see `vfs.c:2681`) in one go as a way to make sure nothing happens during a checkpoint. This is something that does not happen normally in SQLite as locks are taking only when needed.

However, we need a way to make the checkpoint operation as a "either do nothing or do the whole truncate" and taking all locks exclusively fulfils that purpose.

src/vfs.c
marco6: This was true for the previous implementation as well. We can sure improve this in a later step.

src/vfs.c
marco6: I think this came in the original code following the same pattern found in the main file, where pages are numbered starting at page 1.

I couldn't find a reference to "WAL page numbers" and in the SQLite source they are using a normal index (usually called `iPage`).

I am so-so on this one: on one side, using normal indexing follows the normal pattern and makes reading easier. On the other, given that there are really only 2 files, having the indexing working in the same way for the main and the wal might make sense?

src/vfs.c
marco6: Am I missing something? The code above is not holding any locks... Could you point me to a place where this happens?

src/vfs.c
marco6: BackfillAttempted is never accessed atomically (i.e. the readers), but only by the logic running the checkpoint/the write transaction (= when you already got the write lock).

As such, no reason to create a synchronization point,

src/vfs.c
marco6: How is that outdated? Two memcpys for the WAL index copies follow... 

Regarding the order: backfill related stuff are only read (and considered) when doing a checkpoint and when starting a read transaction. There is no harm in a read transaction reading an old backfill as the worst it can happen is that a reader will take a read mark even if the frames have already been backfilled (and as such it could have taken the 0 read mark and let concurrent checkpoints run to completion and the WAL restart). As such, it is more important to quickly publish the "transaction data" (WAL headers and hash maps) than to update those fields.

On the other side, with the new locking mechanism for checkpoints, I think the second part (the backfill one) is now not necessary any more as we are not redirecting the memory during checkpoints. We could likely have an assert instead, but maybe it is ok to have a dead branch (`f->exclMask & (1 << VFS__WAL_CKPT_LOCK)` should never be true) so that the "publishing" logic is a bit more "future proof". WDYT?

src/vfs.c
marco6: This is part of the (many) [little details](https://github.com/sqlite/sqlite/blob/master/src/wal.c#L2585) of a (almost) lock-free algorithm like the one used by SQLite and it is a bit counter-intuitive.

Readers read the two regions in order, writers do that the other way around. This way, if the thread fence synchronizes on the first read then it is guaranteed they also synchronize correctly on the second one. This would not be true when writing and reading in the same order.

src/vfs.c
marco6: I copied this from [SQLite source](https://github.com/sqlite/sqlite/blob/master/src/os_unix.c#L6850).

src/vfs.c
marco6: You are missing the fact that SQLite will still try to take the locks as from its PoV no locks have been taken yet. And also unlock them.

Instead, we should hold them for the entire duration of the checkpoint so that the checkpoint logic becomes "atomic": either it runs or it doesn't.

src/vfs.c
marco6: The part that is after `mx_frames` should never be read by any connection that is not the one performing the transaction, otherwise a reader might read uncommitted pages.

src/vfs.c
marco6: Yes that was the intent of the "outside of the protection of the mtx" part: the `mtx` is necessary as a `VfsApply` might reallocate the frames array, but the rest is "private".

src/vfs.c
marco6: For now, let's just use normal names.

src/vfs.c
marco6: Do we want to terminate the host application in this case? Not against it, but not sure either..

src/gateway.c
marco6: Is that generic though? It will return `SQLITE_ERROR` (generic) but with the text `bind parameters` to the user.

test/unit/test_concurrency.c
marco6: It will (i.e. the test will fail) but we wouldn't know which step did it.

src/vfs.c
marco6: Kinda. This method will run on many different threads *unsynchronized*. So, just reading values from the database structur (`n_pages`, `pages` and *the content* of the pages) might read stale data.

There is only one routine updating those fields: the checkpoint logic. SQLite will call `xShmBarrier` (`vfsMainFileShmBarrier`) after a checkpoint (actually after all updates to the shared memory, including writes). All barriers are expected (by sqlite interface) to be full memory barriers.

Now the question though: this method will *not* call any barrier, so, is it enough? (barriers are effective only if both reader and writers use them, otherwise stale data is still a thing).

And the answer is "yes" because SQLite will also issue barrier when starting the transaction.

src/vfs.c
letFunny: You are right, I got it the other way around

src/vfs.c
letFunny: I think this is one of the cases where an error indicates a bug in our code and it's better to crash than to risk some corruption but I may be wrong here.

src/vfs.c
letFunny: > We can sure improve this in a later step.

Sure, if it is not a regression then no problem. Can you write a TODO in the code so the next person can see it just by grepping? Something saying TODO this could be optimized to avoid reallocating...

src/vfs.c
letFunny: I don't mind then but what about making the distinction between `index` and `number`, so `tx_pageno` is better than `index` because it is the page number which happens to be represented in a 0-based array.

src/vfs.c
letFunny: Forget about this one, it was almost surely based on the previous version of the code. I reviewed commits individually and then we change the locking so that each struct contained a mutex instead of having a "broad" one.

src/vfs.c
letFunny: > How is that outdated?

I guess my complaint was about being more precise. We say that _finally_ we are going to copy the information and remap while in reality we have to publish more fields and we have to be careful about those. Re-reading it, it looks more like something very minor. The original intention was to highlight that between commits we changed how we publish the data and we didn't update the comment.

To summarize: I am happy keeping it this way.

> is now not necessary any more as we are not redirecting the memory during checkpoints. We could likely have an assert instead, but maybe it is ok to have a dead branch

Let's discuss this tomorrow in our meeting.

src/gateway.c
letFunny: What is the different in the user-facing error message between this line and line 691?

test/unit/test_concurrency.c
letFunny: That makes sense, then keep it.

src/vfs.c
marco6: I added a `FIXME` for now.

src/vfs.c
marco6: Discussed offline.

src/vfs.c
marco6: Discussed offline, it wouldn't be very useful in this case.

src/gateway.c
marco6: Discussed offline. We are actually removing the other trace.

src/vfs.c
marco6: Discussed offline, seems easier this way,

marco6:
> The only thing pending is the assert we added in the meeting that was failing a test.

This actually found a hard-to-catch bug. I thing it couldn't happen as the interrupt does not work at all, but it's still good to fix it. It is addressed in #798.

just-now (approved):
+1. Looks good.

PR #748, opened 2025-05-11
fix: compute commit tag correctly

marco6:
This PR fixes the computation of the commit tag in the WAL frame so that SQLite truncates the main file at the correct size. Before this PR, the tag could always only increase.


src/vfs.c
```diff
@@ -2280,8 +2280,8 @@ static int vfsWalAppend(struct vfsWal *w,
 			goto oom_after_frames_alloc;
 		}
 
-		if (page_number > database_size) {
-			database_size = page_number;
+		if (page_number == 1) {
```
letFunny, line 2283: Can you write a comment here explaining a bit of the reasoning?

letFunny (approved):
Thanks for this Marco, LGTM, this is quite simple actually.

PR #751, opened 2025-05-19
refactor: remove deprecated vfs interface

marco6:


marco6:
Depends on #740.

marco6:
Ready as #740 was merged.

include/dqlite.h
```diff
@@ -7,7 +7,11 @@
 #include <stdint.h>
 
 #ifndef DQLITE_API
-#define DQLITE_API __attribute__((visibility("default")))
+# if defined(__has_attribute) && __has_attribute(visibility)
+#  define DQLITE_API __attribute__((visibility("default")))
+# else
+#  define DQLITE_API
+# endif
```
letFunny, line 14: Do we really need this? If I understand it correctly, if the attribute is not supported then `-fvisibility=hidden` in the Makefile will make everything hidden. Maybe it's better to fail here instead if not supported and know about it.

include/dqlite.h
```diff
@@ -673,90 +677,6 @@ DQLITE_API const char *dqlite_node_errmsg(dqlite_node *n);
  */
 DQLITE_API dqlite_node_id dqlite_generate_node_id(const char *address);
 
-/**
```
letFunny, line 679: I don't think we can move this fix to the LTS version if we are changing the public interface and removing functions which means removing the VFS might lead to a lot of conflicts when cherry-picking the PRs that are to come. It might end up being more work than not removing the interface in the first place, thoughts?

src/vfs.h
```diff
@@ -18,19 +18,22 @@ int VfsEnableDisk(struct sqlite3_vfs *vfs);
  * SQLite global registry. */
 void VfsClose(struct sqlite3_vfs *vfs);
 
+struct vfsTransaction {
```
letFunny, line 21: This is indeed much more semantically meaningful, thanks!

src/vfs.c
```diff
@@ -2124,37 +2123,41 @@ static int vfsWalPoll(struct vfsWal *w, dqlite_vfs_frame **frames, unsigned *n)
...
 	if (commit == 0) {
-		*frames = NULL;
-		*n = 0;
+		*transaction = (struct vfsTransaction){ };
 		return 0;
 	}
 
-	*frames = sqlite3_malloc64(sizeof **frames * w->n_tx);
-	if (*frames == NULL) {
+	uint64_t *numbers = sqlite3_malloc64(sizeof *numbers * w->n_tx);
```
letFunny: Nit: I prefer parenthesis around sizeof even though it has precedence and it is not a function and whatever, but I think it is just easier to read `sizeof(T) * something`.

include/dqlite.h
```diff
@@ -673,90 +677,6 @@ DQLITE_API const char *dqlite_node_errmsg(dqlite_node *n);
...
- * This function is DEPRECATED and will be removed in a future major release.
- *
- * Initialize the given SQLite VFS interface object with dqlite's custom
- * implementation, which can be used for replication.
- */
-DQLITE_API int dqlite_vfs_init(sqlite3_vfs *vfs, const char *name);
-
-DQLITE_API int dqlite_vfs_enable_disk(sqlite3_vfs *vfs);
```
letFunny, line 687: [IMPORTANT]: This function in particular is part of the public API and nor marked as deprecated.

src/fsm.c
```diff
@@ -213,67 +174,30 @@ static int apply_frames(struct fsm *f, const struct command_frames *c)
...
+		rv = DQLITE_PROTO;
+		goto error;
 	}
 
-	sqlite3_free(page_numbers);
 	maybeCheckpoint(db);
-	return 0;
+error:
+	sqlite3_free(c->frames.page_numbers);
+	sqlite3_free(c->frames.pages);
```
letFunny, line 198: I think we could use a function to free the transaction object and set its pointers to NULL. That way the logic is uniform across the code and abstracted.

src/fsm.c
```diff
@@ -213,67 +174,30 @@ static int apply_frames(struct fsm *f, const struct command_frames *c)
...
 		}
+	} else {
+		rv = DQLITE_PROTO;
+		goto error;
 	}
 
-	sqlite3_free(page_numbers);
 	maybeCheckpoint(db);
-	return 0;
+error:
```
letFunny, line 196: Nit: weird that we have an `error` tag that is also executed as part of the happy path.

src/fsm.c
```diff
@@ -213,67 +174,30 @@ static int apply_frames(struct fsm *f, const struct command_frames *c)
...
-			if (rv != 0) {
-				tracef("malloc");
-				sqlite3_free(page_numbers);
-				return DQLITE_NOMEM;
-			}
-			rv =
-			    VfsApply(db->vfs, db->path, f->pending.n_pages,
-				     f->pending.page_numbers, f->pending.pages);
-			if (rv != 0) {
-				tracef("VfsApply failed %d", rv);
```
letFunny, line 245: Why did we remove the traces? Comment applies to all occurrences.

src/command.c
```diff
@@ -31,39 +32,96 @@ static size_t frames__sizeof(const frames_t *frames)
...
+	return DQLITE_OK;
+}
+
 static int frames__decode(struct cursor *cursor, frames_t *frames)
 {
-	int rc;
-	rc = uint32__decode(cursor, &frames->n_pages);
-	if (rc != 0) {
-		return rc;
+	int rv;
```
letFunny: hahaha `return value` vs `return code`, I see that you have a very strong dislike to the latter

src/command.c
```diff
@@ -134,36 +192,3 @@ int command__decode(const struct raft_buffer *buf, int *type, void **command)
 	*type = h.type;
 	return 0;
 }
-
-int command_frames__page_numbers(const struct command_frames *c,
```
letFunny, line 138: This code was much much worse IMO, thanks for replacing it

src/command.h
```diff
@@ -16,16 +16,11 @@ enum { COMMAND_OPEN = 1, COMMAND_FRAMES, COMMAND_UNDO, COMMAND_CHECKPOINT };
...
-	/* TODO: because the sqlite3 replication APIs are asymmetrics, the
-	 * format differs between encode and decode. When encoding data is
-	 * expected to be a sqlite3_wal_replication_frame* array, and when
-	 * decoding it will be a pointer to raw memory which can be further
-	 * decoded with the command_frames__page_numbers() and
-	 * command_frames__pages() helpers. */
-	const void *data;
+	uint32_t   n_pages;
+	uint16_t   page_size;
+	uint16_t   __unused__;
```
letFunny, line 21: Is it really necessary to do padding manually?

src/command.h
```diff
@@ -16,16 +16,11 @@ enum { COMMAND_OPEN = 1, COMMAND_FRAMES, COMMAND_UNDO, COMMAND_CHECKPOINT };
...
 {
-	uint32_t n_pages;
-	uint16_t page_size;
-	uint16_t __unused__;
-	/* TODO: because the sqlite3 replication APIs are asymmetrics, the
-	 * format differs between encode and decode. When encoding data is
-	 * expected to be a sqlite3_wal_replication_frame* array, and when
-	 * decoding it will be a pointer to raw memory which can be further
-	 * decoded with the command_frames__page_numbers() and
-	 * command_frames__pages() helpers. */
```
letFunny, line 27: Nice to see this go, unnecessary optimization

src/fsm.c
```diff
@@ -213,67 +174,30 @@ static int apply_frames(struct fsm *f, const struct command_frames *c)
...
+		rv = VfsApply(db->vfs, db->path, &transaction);
 		if (rv != 0) {
-			tracef("add pending pages failed %d", rv);
-			sqlite3_free(page_numbers);
-			return DQLITE_NOMEM;
+			goto error;
 		}
+	} else {
+		rv = DQLITE_PROTO;
+		goto error;
```
letFunny, line 192: Also breaking change.

test/unit/test_vfs_extra.c
```diff
@@ -255,10 +231,13 @@ struct tx
...
-	do {                           \
-		free(TX.frames);       \
-		free(TX.page_numbers); \
+#define DONE(TX)                                         \
+	do {                                                 \
+		for (unsigned _i = 0; _i < (TX).n_pages; _i++) { \
+			sqlite3_free((TX).pages[_i]);                \
+		}                                                \
+		sqlite3_free((TX).pages);                        \
+		sqlite3_free((TX).page_numbers);                 \
```
letFunny, line 240: Another place where destructor can be used.

letFunny (changes requested):
Thanks for this Marco! It's very nice to see old code and a significant source of complexity go. However, I have some questions regarding support and backporting to the LTS which are quite important.

include/dqlite.h
marco6: If it is not supported, then likely the compiler will not have `-fvisibility` at all. Which means that everything will be *visible* not *hidden*.

include/dqlite.h
marco6: I don't think any of the code we have been merging will ever go to the LTS honestly. Why should we? New features should never get into that branch, only bug fixes.

src/vfs.c
marco6: Me too, but that was the style in the original PR. I think the readability concern is high enough for me to not respect the original way. I added the parenthesis.

include/dqlite.h
marco6: This is not part of the public API IMHO as you would need to access internal datastructures (the `vfs`) and there is no way to understand which VFS a particular node refers to as there are no getters and the name is automatically generated by [`config_init`](https://github.com/canonical/dqlite/blob/master/src/config.c#L47).

src/command.h
marco6: Not necessary IMHO, but I didn't want to touch that.

src/fsm.c
marco6: This is actually a breaking change in two ways:
 - we don't support coexisting from the "experimental" v1
 - we return an error in that case.
 
 I think it is a fair assumption.

marco6: As such I also updated the comment above the if.

src/command.c
marco6: changed to `rc`. We should really settle to one of those.

src/fsm.c
marco6: This is kind of normal in C. See [an example](https://github.com/sqlite/sqlite/blob/master/src/pager.c#L5434) from  a well-known library :-)

src/fsm.c
marco6: Because it is duplicated: `VfsApply` will trace with more information if the only operation that can fail (` vfsWalAppend`) fails. See [here](https://github.com/canonical/dqlite/blob/master/src/vfs.c#L2461).

(and yes, it would be nice if we had a way to error with more information and log at the top-level operation instead of relying on knowing internals. It seems like outside the scope of this PR however).

letFunny: I agree it is better to log information at more levels. In this case however I would prefer to keep the log because it **adds** information to the otherwise more low-level `wal append failed rv:%d n_pages:%u n:%u", rv`. Could we add it back?

src/fsm.c
letFunny: Discussed offline, sometimes the data is owned and some times it is not and it is not "easy" to model the difference because we don't follow neither pattern of callee owns and frees the memory or object itself owns and frees the memory.

Additionally, in this case you could argue we are not freeing the tx and it is a "coincidence". With that said, can we please add a comment in the definition of `vfsTransaction` saying the data can be either owned if you are the leader or not if you are the follower? I fear the next one touching this code will free it incorrectly.

src/fsm.c
marco6: I added back the trace. I'm not really convinced that having the log with:

```
wal append failed rv: 10 n_pages:1 n:2
VfsApply failed: 10
```

with same error code, same operation and less information helps. dqlite log is already spammy enough IMHO. Still, I added it back.

src/vfs.h
```diff
@@ -18,19 +18,24 @@ int VfsEnableDisk(struct sqlite3_vfs *vfs);
...
+	uint32_t    n_pages;      /* Number of pages in the transaction. */
+	uint64_t   *page_numbers; /* Page number for each page. */
+	void      **pages;        /* Content of the pages. */
+};
+
+void VfsTransactionClose(struct vfsTransaction *transaction);
```
letFunny: I think this one is not used and we agreed to not have it

src/fsm.c
```diff
@@ -203,18 +200,9 @@ static int apply_frames(struct fsm *f, const struct command_frames *c)
 static int apply_undo(struct fsm *f, const struct command_undo *c)
```
letFunny, line 203: Why keep the function and make it a noop instead of removing it? Also, why have a command_undo if it is not used.

letFunny (commented):
Almost there Marco, thanks for this one!

src/vfs.h
marco6: Right, it was a prototype I had in mind before discussing it with you on MM. :-( removing it!

src/fsm.c
marco6: I think it makes sense from the code perspective to have a method for each command type. These types are no-ops now, maybe they will not in a future release.

PR #752, opened 2025-05-20
chore: improve gateway readability

marco6:
This PR moves the order of the methods in `gateway.c` so that they are semantically close. The aim of this PR is to improve patch readability for #740 .

It also removes usage of `sqlite3_{malloc,free}` for RAFT related operations in favour of `raft_{malloc,free}`.

Last, it also includes an improved version of `START_V0` that can now omit the response, in case it's not needed. This removes all the awkward `(void)response;` in the code.

### DIFF
--- a/src/gateway.c
+++ b/src/gateway.c
@@ -19,18 +19,19 @@ void gateway__init(struct gateway *g,
 		   struct raft *raft)
 {
 	tracef("gateway init");
-	g->config = config;
-	g->registry = registry;
-	g->raft = raft;
-	g->leader = NULL;
-	g->req = NULL;
-	g->exec.data = g;
+	*g = (struct gateway) {
+		.config = config,
+		.registry = registry,
+		.raft = raft,
+		.exec = {
+			.data = g,
+		},
+		.barrier = {
+			.data = g,
+		},
+		.protocol = DQLITE_PROTOCOL_VERSION,
+	};
 	stmt__registry_init(&g->stmts);
-	g->barrier.data = g;
-	g->barrier.cb = NULL;
-	g->barrier.leader = NULL;
-	g->protocol = DQLITE_PROTOCOL_VERSION;
-	g->client_id = 0;
 }
 
 /* FIXME: This function becomes unsound when using the new thread pool, since
@@ -88,7 +89,7 @@ void gateway__leader_close(struct gateway *g, int reason)
 	}
 	stmt__registry_close(&g->stmts);
 	leader__close(g->leader);
-	sqlite3_free(g->leader);
+	raft_free(g->leader);
 	g->leader = NULL;
 }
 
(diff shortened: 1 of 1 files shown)


src/gateway.c
```diff
@@ -103,26 +104,38 @@ void gateway__close(struct gateway *g)
...
-		}                                                     \
+		}                                                 \
 	}
 
+/* Declare a request struct and a response struct of the appropriate types and
+ * decode the request. This is used in the common case where only one schema
+ * version is extant. */
+#define START_V0(...)        \
+	DECLARE_V0(__VA_ARGS__); \
+    INIT_V0(__VA_ARGS__)
```
letFunny, line 137: tabs, spaces.

src/gateway.c
```diff
@@ -103,26 +104,38 @@ void gateway__close(struct gateway *g)
...
+		rv_ = request_##REQ##__decode(cursor, &request);  \
+		if (rv_ != 0) {                                   \
 			return rv_;                                   \
-		}                                                     \
+		}                                                 \
 	}
 
+/* Declare a request struct and a response struct of the appropriate types and
+ * decode the request. This is used in the common case where only one schema
+ * version is extant. */
```
letFunny, line 134: We need to tune this comment to say that declare a request struct and possibly a response struct ... Probably the usage is shorter and more concise, something like:
```
Usage:
- `START_VO(request_type)` Declares a request struct ...
- `START_VO(request_type, response_type)` Declares request and response structs ...
In both usages the request is decoded. This is used in the common case...
```

letFunny (approved):
Nice and simple, thanks for splitting the PRs Marco, highly appreciated!

src/gateway.c
```diff
@@ -103,26 +104,38 @@ void gateway__close(struct gateway *g)
...
-#define START_V0(REQ, RES, ...)                                       \
-	struct request_##REQ request = { 0 };                         \
-	struct response_##RES response = { 0 };                       \
-	{                                                             \
-		int rv_;                                              \
-		if (req->schema != 0) {                               \
+#define DECLARE_REQUEST(REQ, ...) struct request_##REQ request = { 0 }
+#define DECLARE_RESPONSE(REQ, RES, ...)    \
+    DECLARE_REQUEST(REQ);                  \
+    struct response_##RES response = { 0 }
```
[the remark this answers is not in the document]
marco6: I think that the usual way of doing things is to write macros that require arguments "as if" they were functions, which is why a lot of those are wrapped in a `do { ... } while (0)`.

From the reader perspective IMHO it feels more natural to end a statement with semicolons.

src/gateway.c
```diff
@@ -103,26 +104,38 @@ void gateway__close(struct gateway *g)
...
- * decode the request. This is used in the common case where only one schema
- * version is extant. */
-#define START_V0(REQ, RES, ...)                                       \
-	struct request_##REQ request = { 0 };                         \
-	struct response_##RES response = { 0 };                       \
-	{                                                             \
-		int rv_;                                              \
-		if (req->schema != 0) {                               \
+#define DECLARE_REQUEST(REQ, ...) struct request_##REQ request = { 0 }
+#define DECLARE_RESPONSE(REQ, RES, ...)    \
```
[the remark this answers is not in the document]
marco6: I'll go with `DECLARE_REQUEST_RESPONSE`.

src/gateway.c
marco6: I went with slightly different wording. Hope that helps

letFunny (approved):
LGTM

PR #753, opened 2025-05-20
feat: use SQLITE_STATIC binding

marco6:
The whole protocol and the `gateway.c` implementation assume that the current request message is fixed in memory and is never overwritten until a response has been sent (including multiple message responses like the query rows). This is also why the part around interrupting long queries doesn't work.

Given that the solution to the above would be to allocate a message for each request, it is safe to use `SQLITE_STATIC` and let SQLite reference data in the message.

### DIFF
--- a/src/bind.c
+++ b/src/bind.c
@@ -6,43 +6,36 @@ static int bind_one(sqlite3_stmt *stmt, int n, struct value *value)
 {
 	int rc;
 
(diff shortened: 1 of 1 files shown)


letFunny (approved):
Thanks Marco, discussed offline and I am convinced of the flow now.

PR #756, opened 2025-06-04
fix: dump database without open

marco6:
This PR fixes a bug that was created by #740.


test/unit/test_gateway.c
```diff
@@ -2796,6 +2796,221 @@ TEST_CASE(request_cluster, unrecognizedFormat, NULL)
...
+{
+	(void)params;
+	struct request_dump_fixture *f = data;
+	uint64_t stmt_id;
+
+	OPEN;
+	EXEC("CREATE TABLE test (data BLOB)");
+	/* Make sure we force a checkpoint */
+	struct value pages = { .type = SQLITE_INTEGER,
+			       .integer =
```
letFunny: Nit: bad formatting.

test/unit/test_gateway.c
```diff
@@ -2796,6 +2796,221 @@ TEST_CASE(request_cluster, unrecognizedFormat, NULL)
...
+	HANDLE(DUMP);
+	ASSERT_CALLBACK(DQLITE_OK, FILES);
+	DECODE(&f->response, files);
+
+	munit_assert_int(f->response.n, ==, 2);
+	struct file main = {};
+	DECODE_FILE(&main);
+	munit_assert_string_equal(main.name, "test");
+	munit_assert_int(main.content.len, >=,
+			 f->gateway->config->checkpoint_threshold * 512);
```
letFunny: Where is the `512` coming from? I assume this is the value of page_size in the query above. I know there are platforms that can use bigger pages. I would suggest getting the proper value here to avoid test breaking unexpectedly.

letFunny (approved):
Thanks Marco, LGTM! And thanks for adding all the missing tests, amazing that we didn't have them.

test/unit/test_gateway.c
marco6: I was able to get clang-tidy do something better.

test/unit/test_gateway.c
marco6: This now is taken from the configuration instead.

PR #758, opened 2025-06-06
fix: unresponsive connection after leadership lost

marco6:


marco6:
Rebased on master now that #762 is in: that should fix the tests (except musl which is resolved by the multithreading)

marco6:
Tests are passing. I'll merge.

PR #760, opened 2025-06-06
refactor: separate VFS journal behavior

marco6:
This PR separates the behaviour of the Journal file in the VFS implementation from the logic of the other files. This way it is cleaner to understand what a Journal file does and when changing the behaviour of a file (for example, the WAL) we can be sure the the other behaviours are not accidentally changing.

This was part of #747 .


src/db.c
```diff
@@ -85,6 +85,14 @@ static int dqlite_authorizer(void *pUserData, int action, const char *third, con
 		if (third != NULL && third[0] != '\0') {
 			return SQLITE_DENY;
 		}
+	} else if (action == SQLITE_PRAGMA) {
+		if (strcasecmp(third, "journal_mode") == 0 && fourth) {
+			/* When the user executes 'PRAGMA journal_mode=x' we ensure
+			* that the desired mode is 'wal'. */
+			if (strcasecmp(fourth, "wal") != 0) {
+				return SQLITE_DENY;
+			}
```
letFunny: Why even allow this in the first place? When we run the initial queries for a connection are we also executing the authorizer or is it put in place later on?

src/vfs.c
```diff
@@ -75,6 +74,107 @@ const int vfsOne = 1;
...
+
+static int vfsJournalSectorSize(sqlite3_file* file)
+{
+	(void)file;
+	return 0;
+}
+
+static int vfsJournalRead(sqlite3_file* file, void* data, int iAmt, sqlite3_int64 iOfst) {
+	(void)file;
+	(void)iOfst;
```
letFunny: tabs and spaces, applies to the whole file.

src/vfs.c
```diff
@@ -1801,6 +1897,12 @@ static int vfsOpen(sqlite3_vfs *vfs,
 		vfs = sqlite3_vfs_find("unix");
 		assert(vfs != NULL);
 		return vfs->xOpen(vfs, NULL, file, flags, out_flags);
+	} else if (flags & SQLITE_OPEN_MAIN_JOURNAL) {
+		/* Journal file is just a noop file as only WAL mode is supported */
+		file->pMethods = &vfsNoopMethods;
+		return SQLITE_OK;
+	} else {
+		assert((flags & SQLITE_OPEN_DELETEONCLOSE) == 0);
```
letFunny: I think the else is redundant, the assert is two lines below

src/db.c
marco6: Ok, I changed the initialization order so that the authorizer comes last. No point in running it when we are in control of the commands.

src/vfs.c
marco6: I hope I fixed those. If I run clang-tidy on the file we will get just so many unrelated changes...

src/vfs.c
marco6: removed.

PR #761, opened 2025-06-09
fix: missing barrier events

marco6:
This PR fixes a bug where connection would be kept open, but in an unresponsive state after a node stop, if a barrier was still in progress.

The reason was a missing callback for barrier failed during leadership loss.

### DIFF
--- a/test/integration/test_node.c
+++ b/test/integration/test_node.c
@@ -276,6 +276,7 @@ static int register_notify(sqlite3 *connection, char **pzErrMsg, struct sqlite3_
 static sem_t write_sem;
 static void notify_transaction(sqlite3_context *context, int argc, sqlite3_value **argv)
 {
+	(void)argc;
 	/* Just return the same value */
 	sqlite3_result_value(context, argv[0]);
 	/* Signal the application that we were able to get to the write part */


test/integration/test_node.c
```diff
@@ -181,6 +186,123 @@ TEST(node, startInetStopTwice, setUpInet, tearDown, 0, node_params)
...
+	munit_assert_int(rv, ==, RAFT_OK);
+
+
+	for (int i = 0; i < CLIENT_N; i++) {
+		rv = clientSendExecSQL(&clients[i], 
+			"INSERT INTO test_table VALUES (1)", NULL, 0, NULL);
+		munit_assert_int(rv, ==, RAFT_OK);
+	}
+
+	usleep(1000);
```
just-now: Is it worth to try `sched_setaffinity()` for the whole process and test in-flight items cleanly?

test/integration/test_node.c
```diff
@@ -181,6 +186,123 @@ TEST(node, startInetStopTwice, setUpInet, tearDown, 0, node_params)
...
+	rv = clientRecvDb(c, NULL);
+	if (rv != RAFT_OK) {
+		clientClose(c);
+		return rv;
+	}
+	return RAFT_OK;
+}
+
+
+TEST(node, stopInflightReads, setUpInet, tearDown, 0, node_params)
```
letFunny, line 220: How was this failing before this change? The PR description is empty.

test/integration/test_node.c
```diff
@@ -181,6 +186,123 @@ TEST(node, startInetStopTwice, setUpInet, tearDown, 0, node_params)
...
+			NULL, 0, NULL
+		);
+		munit_assert_int(rv, ==, RAFT_OK);
+		
+		rv = clientRecvRows(&clients[i], &rows, &done, NULL);
+		munit_assert_int(rv, ==, RAFT_OK);
+		munit_assert_false(done);
+		munit_assert_int(rows.column_count, ==, 1);
+		munit_assert_string_equal(rows.column_names[0], "i");
+		clientCloseRows(&rows);
```
letFunny, line 253: For clarity can you move this to line 260 just before `clientClose`? When reading this you have to open the function just to make sure it is not closing anything, it is only freeing the rows.

test/integration/test_node.c
letFunny: Why do you need to yield here?

letFunny (commented):
Thanks for this Marco, just a couple of questions because I am missing some context.

test/integration/test_node.c
marco6: We yield because we need a write to be happening while changing leadership for this bug to manifest. I have no easy way out to understand that in a "integration" test

test/integration/test_node.c
marco6: Do we really want to have an array of rows? It seems to me fair to open and close them here...

test/integration/test_node.c
marco6: updated

test/integration/test_node.c
letFunny: What about changing the name to `clientFreeRows`?

test/integration/test_node.c
letFunny: Ty

test/integration/test_node.c
letFunny: Discussed offline, we are yielding hoping that there are some write requests in flight.

letFunny (approved):
Thanks Marco for the fix

PR #763, opened 2025-06-09
feat: keep connection open on leadership loss

marco6:
#740 introduces a change in protocol behaviour that closes the connection on leadership loss. This PR reverts that decision and let's the [client close the connection](https://github.com/canonical/go-dqlite/blob/v3/driver/driver.go#L850), if deemed necessary.

This was the "common" behaviour before.

## Rationale

I specified "common" as when no error occurs, everything returned a message and called the `gateway_handle_cb` except in case a raft transfer membership started (as in `raft_transfer`, see `handle_transfer` in `gateway.c`), but is not done yet then `raft->transfer != NULL` (set by `membershipLeadershipTransferInit`).

`leader_exec_v2` had 2 error-reporting paths:
 - the "async" path, which will return the error/success to the callback
 - the "sync" path, which will return the error/success to the caller.

In this case, the instance is still the leader, so:
 - `CHECK_LEADER` will pass.
 - `raft_barrier` will fail *synchronously* returning `RAFT_NOTLEADER`
    ```c
    int raft_barrier(struct raft *r, struct raft_barrier *req, raft_barrier_cb cb)
    {
	    raft_index index;
	    struct raft_buffer buf;
	    int rv;
    
	    if (r->state != RAFT_LEADER || r->transfer != NULL) {
		    rv = RAFT_NOTLEADER;
		    goto err;
	    }
	    ...
    ```

Now, following the execution path on the old `leader_exec_v2` (commit 8fcffef):

```c
static int handle_exec(struct gateway *g, struct handle *req)
{
	...
	CHECK_LEADER(req); /* will pass */
	...
	req->stmt_id = stmt->id;
	g->req = req;
	rv = leader_exec_v2(g->leader, &g->exec, stmt->stmt, leader_exec_cb);
	...
```

Which initialises the request and then calls `exec_tick`:

```C
int exec_tick(struct exec *req, int status)
{
	...
	switch (sm_state(&req->sm)) {
	case EXEC_START:
		PRE(status == 0);
		l = req->leader;
		PRE(l != NULL);
		barrier_rv = leader_barrier_v2(l, &req->barrier, exec_barrier_cb);
```

Which in turn does:

```C
int leader_barrier_v2(struct leader *l,
		      struct barrier *barrier,
		      barrier_cb cb)
{
	int rv;

	if (!needsBarrier(l)) {
		return LEADER_NOT_ASYNC;
	}

	sm_init(&barrier->sm, barrier_invariant, NULL, barrier_states, "barrier",
		BARRIER_START);
	barrier->cb = cb;
	barrier->leader = l;
	barrier->req.data = barrier;
	rv = barrier_tick(barrier, 0);
	POST(rv != LEADER_NOT_ASYNC);
	return rv;
}

static int barrier_tick(struct barrier *barrier, int status)
{
	int rv;

	if (sm_state(&barrier->sm) == BARRIER_START) {
		PRE(status == 0);
		rv = raft_barrier(barrier->leader->raft, &barrier->req, barrier_raft_cb);
		if (rv != 0) {
			barrier_done(barrier, rv);
		}
		return rv;
	}
	...
}

static void barrier_done(struct barrier *barrier, int status)
{
	...
	if (status != 0) {
		sm_fail(&barrier->sm, BARRIER_FAIL, status);
	} else {
		sm_move(&barrier->sm, BARRIER_DONE);
	}
	...
	if (state == BARRIER_PASSED) {
		cb(barrier, status);
	}
}
```

In case a barrier is needed, the callback is not called and `RAFT_NOTLEADER` is returned. Now, back on `exec_tick`:

```c
int exec_tick(struct exec *req, int status)
{
	...
	switch (sm_state(&req->sm)) {
	case EXEC_START:
		...
 		barrier_rv = leader_barrier_v2(l, &req->barrier, exec_barrier_cb);
		if (barrier_rv == 0) {
			/* suspended */
			ret = 0;
			break;
		} else if (barrier_rv != LEADER_NOT_ASYNC) {
			/* return error to caller, don't invoke callback,
			 * but set req->status so that that the SM will
			 * record the failure */
			req->status = barrier_rv;
			exec_done(req, LEADER_NOT_ASYNC);
			ret = barrier_rv;
			break;
		}
		...
	}

	return ret;
```

`exec_done` in this case does not call the callback, but sets the status to `EXEC_FAILED` and the error is *returned*. In turn `handle_exec` will push the error up to the conn (through `gateway__handle`):

```C
static int handle_exec(struct gateway *g, struct handle *req)
{
	...
	rv = leader_exec_v2(g->leader, &g->exec, stmt->stmt, leader_exec_cb);
	if (rv == LEADER_NOT_ASYNC) {
		leader_exec_cb(&g->exec, g->exec.status);
	} else if (rv != 0) {
		tracef("handle exec leader exec failed %d", rv);
		g->req = NULL;
		return rv;
	}
	return 0;
}
```

And last, the error is handled with a `conn__stop` from `conn.c`:

```c
	rv = gateway__handle(&c->gateway, &c->handle, c->request.type,
			     c->request.schema, &c->write, gateway_handle_cb);
	if (rv != 0) {
		tracef("read gateway handle error %d", rv);
		conn__stop(c);
	}
```

This behaviour was even documented in a test:

```c
	rv = clientSendExec(f->client, stmt_id, NULL, 0, NULL);
	munit_assert_int(rv, ==, 0);
	rv = clientRecvResult(f->client, &last_insert_id, &rows_affected, NULL);
	munit_assert_int(rv, ==, DQLITE_CLIENT_PROTO_ERROR);
```

`DQLITE_CLIENT_PROTO_ERROR` is returned when the connection is closed for some reason and the client is "unusable":

```c
static int readMessage(struct client_proto *c,
		       uint8_t *type,
		       struct client_context *context)
{
	...
	rv = doRead(c->fd, p, n, context);
	if (rv < 0) {
		return DQLITE_CLIENT_PROTO_ERROR;
	} else if (rv < (ssize_t)n) {
		return DQLITE_CLIENT_PROTO_SHORT;
	}
```

So, when #740 unified the error path by only using the callbacks only one behaviour was necessary and I chose to close the connection given that it was documented in a test.


letFunny (approved):
Discussed offline, this PR is good in the sense that it unifies the error path and the closing sequence. Not perfect but in the good direction.

PR #764, opened 2025-06-12
fix: use clang for musl builds

marco6:


contrib/build-static.sh
```diff
@@ -48,7 +48,9 @@ if [ ! -f "${INSTALL_DIR}/musl/bin/musl-gcc" ]; then
     rm -rf musl
     git clone "${REPO_MUSL}" --depth 1 --branch "${TAG_MUSL}" musl
     cd musl
-    ./configure --prefix="${INSTALL_DIR}/musl"
+    # workaround for missing stdatomic
+    cp /usr/lib/clang/18/include/stdatomic.h include/stdatomic.h
```
letFunny, line 52: Discussed offline this is due to a musl limitation that they won't fix. Can you put a link here to where you found the information? Then consider it approved.

letFunny (approved):
LGTM when the one comment is resolved.

PR #766, opened 2025-06-12
 refactor: group semantically vfs methods

marco6:
This PR reorders all methods and datastructures in vfs.c so that they are grouped in cathegories:
  - helpers
  - datastructures
  - sqlite-specific implementation

This PR is mainly useful to make the changes in #747 more readable and reduce conflict in subsequent work.

This PR also uses `.git-blame-ignore-revs` to make sure as many lines as possible were kept with the original author. As such, please use merge and not squash in this case.


src/vfs.c
```diff
@@ -1799,1624 +1797,1541 @@ static int vfsFileShmUnmap(sqlite3_file *file, int delete_flag)
...
+	cursor = *data;
 
-/* We always report that a lock is held. This routine should be used only in
- * journal mode, so it doesn't matter.
- * TODO does this need to be treated differently in disk-mode?
- * */
-static int vfsDiskFileCheckReservedLock(sqlite3_file *file, int *result)
-{
-	(void)file;
-	*result = 1;
```
letFunny, line 3087: I think the new one is different:
```c
static int vfsNoopCheckReservedLock(sqlite3_file *file, int *pResOut)
{
    (void)file;
    *pResOut = 0;
    return SQLITE_OK;
}
```

src/vfs.c
```diff
@@ -1278,34 +1439,6 @@ static int vfsFileSize(sqlite3_file *file, sqlite_int64 *size)
...
-	return SQLITE_OK;
-}
-
-/* We always report that a lock is held. This routine should be used only in
- * journal mode, so it doesn't matter. */
-static int vfsFileCheckReservedLock(sqlite3_file *file, int *result)
-{
-	(void)file;
-
-	*result = 1;
```
letFunny, line 1305: And this one, I forgot to comment both :P

src/vfs.c
marco6: The `xCheckReservedLock` is only ever used in DELETE mode (which clearly includes the initial transition from DELETE to WAL mode when opening a new database).

Given that the opening is done on the main thread and has no concurrency, this method is useless.

src/vfs.c
marco6: See the other comment.

PR #770, opened 2025-06-16
refactor: separate `{Wal,Main,DiskMain}File`s behaviour

marco6:
This PR further separates the behaviour of the many different files SQLite uses. In detail, it splits the behaviour in 3 different structs:

 - `vfsWalFile` which is only related to `WAL` operation. This file is always in memory for dqlite
 - `vfsMainFile` the real database file. This file is always in memory for dqlite. This file also indirectly manages the shared memory (`-shm` file).
 - `vfsDiskMainFile` again, the real database file, but in `disk` mode. This file is on disk.

### DIFF
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -253,7 +253,7 @@ static void vfsFrameDestroy(struct vfsFrame *f)
 struct vfsShm
 {
 	void **regions;     /* Pointers to shared memory regions. */
-	unsigned n_regions; /* Number of shared memory regions. */
+	int n_regions; /* Number of shared memory regions. */
 	unsigned refcount;  /* Number of outstanding mappings. */
 	unsigned shared[SQLITE_SHM_NLOCK];    /* Count of shared locks */
 	unsigned exclusive[SQLITE_SHM_NLOCK]; /* Count of exclusive locks */
@@ -262,16 +262,7 @@ struct vfsShm
 /* Initialize the shared memory mapping of a database file. */
 static void vfsShmInit(struct vfsShm *s)
 {
-	int i;
-
-	s->regions = NULL;
-	s->n_regions = 0;
-	s->refcount = 0;
-
-	for (i = 0; i < SQLITE_SHM_NLOCK; i++) {
-		s->shared[i] = 0;
-		s->exclusive[i] = 0;
-	}
+	*s = (struct vfsShm){0};
 }
 
 static int vfsShmLock(struct vfsShm *s, int ofst, int n, int flags)
@@ -349,7 +340,7 @@ static int vfsShmUnlock(struct vfsShm *s, int ofst, int n, int flags)
 static void vfsShmClose(struct vfsShm *s)
 {
 	void *region;
-	unsigned i;
+	int i;
 
 	assert(s != NULL);
 
@@ -485,42 +476,6 @@ static uint32_t vfsWalGetPageSize(struct vfsWal *w)
 	return vfsParsePageSize(ByteGetBe32(&w->hdr[8]));
 }
 
-/* Truncate a WAL file to zero. */
-static int vfsWalTruncate(struct vfsWal *w, sqlite3_int64 size)
-{
-	unsigned i;
-
-	/* We expect SQLite to only truncate to zero, after a
-	 * full checkpoint.
-	 *
-	 * TODO: figure out other case where SQLite might
-	 * truncate to a different size.
-	 */
-	if (size != 0) {
-		return SQLITE_PROTOCOL;
-	}
-
-	if (w->n_frames == 0) {
-		return SQLITE_OK;
-	}
-
-	assert(w->frames != NULL);
-
-	/* Restart the header. */
-	formatWalRestartHeader(w->hdr);
-
-	/* Destroy all frames. */
-	for (i = 0; i < w->n_frames; i++) {
-		vfsFrameDestroy(w->frames[i]);
-	}
-	sqlite3_free(w->frames);
-
-	w->frames = NULL;
-	w->n_frames = 0;
-
-	return SQLITE_OK;
-}
-
 /* Release all memory used by a WAL object. */
 static void vfsWalClose(struct vfsWal *w)
 {
@@ -537,6 +492,16 @@ static void vfsWalClose(struct vfsWal *w)
 	if (w->tx != NULL) {
 		sqlite3_free(w->tx);
 	}
+}
+
+static void vfsWalTruncate(struct vfsWal *w)
+{
+	formatWalRestartHeader(w->hdr);
+	vfsWalClose(w);
+	w->frames = NULL;
+	w->n_frames = 0;
+	w->tx = NULL;
+	w->n_tx = 0;
 }
 
 /* Database-specific content */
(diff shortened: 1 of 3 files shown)


src/vfs.c
```diff
@@ -253,7 +253,7 @@ static void vfsFrameDestroy(struct vfsFrame *f)
 struct vfsShm
 {
 	void **regions;     /* Pointers to shared memory regions. */
-	unsigned n_regions; /* Number of shared memory regions. */
+	int n_regions; /* Number of shared memory regions. */
```
letFunny, line 258: Why `int` if this is only > 0? Do you want to detect incorrect logic that does wrapping?

src/vfs.c
```diff
@@ -693,6 +648,85 @@ static uint32_t vfsDatabaseGetPageSize(struct vfsDatabase *d)
...
+ * that originated the transaction, in order to make the transaction visible.
+ *
+ * Note that the hash table contained in the WAL index does not get modified,
+ * and even after a rewind following a write transaction it will still contain
+ * entries for the frames committed by the transaction. That's safe because
+ * mxFrame will make clients ignore those hash table entries. However it means
+ * that in case the replication is not actually successful and
+ * dqlite_vfs_abort() is called the WAL index must be invalidated.
+ **/
+static void vfsAmendWalIndexHeader(struct vfsDatabase *d)
```
letFunny, line 683: Please as we discussed last time separate the commits that move code around from the ones that actually make changes. It is mixed right now so it takes triple the effort to review.

src/vfs.c
```diff
@@ -485,42 +476,6 @@ static uint32_t vfsWalGetPageSize(struct vfsWal *w)
...
-	 */
-	if (size != 0) {
-		return SQLITE_PROTOCOL;
-	}
-
-	if (w->n_frames == 0) {
-		return SQLITE_OK;
-	}
-
-	assert(w->frames != NULL);
```
letFunny, line 507: Why did we remove these checks? Did we change when truncate is called?

src/vfs.c
```diff
@@ -1184,103 +1172,9 @@ static int vfsWalRead(struct vfsWal *w,
...
-			break;
-	}
-
-	/* From SQLite docs:
-	 *
-	 *   If xRead() returns SQLITE_IOERR_SHORT_READ it must also fill
-	 *   in the unread portions of the buffer with zeros.  A VFS that
-	 *   fails to zero-fill short reads might seem to work.  However,
-	 *   failure to zero-fill short reads will eventually lead to
-	 *   database corruption.
```
letFunny, line 1240: I see that we copied this to `vfsWalRead`,  can you also copy it to `vfsMainFileRead`?

src/vfs.c
```diff
@@ -1772,108 +1630,77 @@ static int vfsFileShmLock(sqlite3_file *file, int ofst, int n, int flags)
...
-
-	if (f->db != NULL) {
-		/* Read from the actual database file. */
-		return f->db->pMethods->xRead(f->db, buf, amount, offset);
-	}
-
-	switch (f->type) {
-		case VFS__WAL:
-			rv = vfsWalRead(&f->database->wal, buf, amount, offset);
-			break;
```
letFunny, line 1880: Why did we need this case before?

src/vfs.c
```diff
@@ -2198,259 +1926,117 @@ static int vfsOpen(sqlite3_vfs *vfs,
...
 }
 
 static int vfsDelete(sqlite3_vfs *vfs, const char *filename, int dir_sync)
 {
-	struct vfs *v;
-
 	(void)dir_sync;
-
-	assert(vfs != NULL);
-	assert(vfs->pAppData != NULL);
```
letFunny, line 2434: Why removing the two assertions?

src/vfs.c
```diff
@@ -3306,23 +2870,20 @@ int VfsRestore(sqlite3_vfs *vfs,
 
 	wal = &database->wal;
 
-	/* Truncate any existing content. */
-	rv = vfsWalTruncate(wal, 0);
```
letFunny, line 3332: Why did we change the order of the truncate and `vfsDatabaseRestore`?

src/vfs.c
```diff
@@ -1335,113 +1229,184 @@ static int vfsWalWrite(struct vfsWal *w,
...
-
-	assert(f != NULL);
-
-	switch (f->type) {
-		case VFS__DATABASE:
-			rv = vfsDatabaseTruncate(f->database, size);
-			break;
+	formatWalRestartHeader(f->wal->hdr);
+	vfsWalClose(f->wal);
+	f->wal->frames = NULL;
```
just-now, line 1259: [nit]: it could be easier just to assign `(struct vfsWal){}` to `f->wal`.

src/vfs.c
```diff
@@ -1335,113 +1229,184 @@ static int vfsWalWrite(struct vfsWal *w,
...
-		page_size = vfsWalGetPageSize(w);
-		size += VFS__WAL_HEADER_SIZE;
-		size += (uint64_t)w->n_frames *
-			(uint64_t)(FORMAT__WAL_FRAME_HDR_SIZE + page_size);
+	/* If the main database file is not empty, we expect the
+	 * page size to have been set by an initial write. */
+	page_size = vfsDatabaseGetPageSize(f->database);
+	assert(page_size > 0);
+
+	if (offset < (int)page_size) {
```
just-now: [nit]: I'd put `(int)page_size` into an int `i_page_size` as it's used everywhere and IMVVVVHO is a bit noisy.

src/vfs.c
```diff
@@ -1335,113 +1229,184 @@ static int vfsWalWrite(struct vfsWal *w,
...
+
+	page = vfsDatabasePageLookup(f->database, pgno);
+
+	if (pgno == 1) {
+		/* Read the desired part of page 1. */
+		memcpy(buf, (char *)page + offset, (size_t)amount);
+	} else {
+		/* Read the page. */
+		memcpy(buf, page, (size_t)amount);
+	}
```
just-now: [nit][style]: I'd vote for the following style:
```
sqlite_int64 off = pgno == 1 : offset : 0;
memcpy(buf, (char *)page + off, (size_t)amount);
```

src/vfs.c
```diff
@@ -1772,108 +1630,77 @@ static int vfsFileShmLock(sqlite3_file *file, int ofst, int n, int flags)
...
-	vfsFileControl,
-	vfsNoopSectorSize,
-	vfsNoopDeviceCharacteristics,
-	vfsFileShmMap,
-	vfsFileShmLock,
-	vfsFileShmBarrier,
-	vfsFileShmUnmap,
-	0,
-	0,
+	.iVersion = 2,
```
just-now, line 1675: I was waiting for someone to change this table to C99 style. Thanks!

src/vfs.c
marco6: I generally dislike a lot using unsigned arithmetic as a way to "enforce" correct computation. That is a source of many many very hard-to-debug bugs (unless you have checked arithmetics, but C doesn't have that and I don't want to use everywhere compiler intrinsics).

I think I am willing to spend 1 bit, considering that:
 - `int` is at least 16bit in C
 - the minimum size of a page is  `512 B` (the default would be `4KiB`).
 - 1st region contains 4062 entries in the hash map
 - all other regions contain 4096 entries in the hash map.
 
 As such, with an int size of 16 bit you get to address a WAL which can be as big as:
  - (2^15 - 1) * 4096 + 4062 pages = ~64GiB of data as a maximum *single transaction* size which would already be enough.
 
 However, int size in linux is always 32 bit as it follows the ILP32/LP64 which both have int as 32 bit. In that case, the size of the transaction is way bigger than and disk produced so far AFAIK.

src/vfs.c
marco6: It is a bit subtle, but that would also zero the header which is something we should not do.

src/vfs.c
marco6: I inlined that directly and changed the type of `page` from `void *` to `const char *` so addition is defined without compiler extension.

src/vfs.c
marco6: We had this case as we only had one structure managing:
 - main files
 - in-memory WALs
 - disk files

As such the read and write (and other) methods had to deal with "which file is actually the current one" logic.

src/vfs.c
marco6: I think that it is best to always restart the wal header as the truncate operation actually does not "truncate the size to zero", but it truncates the size to the header[^1]. As such, I think that it's best to always call `formatWalRestartHeader` so that the WAL always restarts correctly.

[^1]: and it works only because SQLite cannot distinguish between "I truncated and another process already created the new header" from "I truncated but actually I didn't" (which is what is happening here).

src/vfs.c
marco6: For the first one, it's part of the SQLite interface contract and I don't see a reason to test that. I'm putting back the other one.

src/vfs.c
marco6: It makes things more "transactional" if that makes sense to you: `vfsDatabaseRestore` can fail while the WAL truncation can't.

This means that if we first do WAL and then database and we fail, then we leave the WAL truncated, but the database still there, which might lead to inconsistencies.

src/vfs.c
letFunny: I love the level of detail :)

Agree with checking for underflows instead of enforcing it using the "type".

src/vfs.c
letFunny: That makes sense to me but I hope no bugs are introduced hahaha

src/vfs.c
letFunny: I am not sure I understand fully the answer. I was asking why did we remove the assumption that `w->frames != NULL`

src/vfs.c
```diff
@@ -1144,15 +1146,21 @@ static int vfsWalFileRead(sqlite3_file* file, void* buf, int amount, sqlite3_int
 	}
 
 	if (index == 0) {
-		// This is an attempt to read a page that was
-		// never written.
+		/* This is an attempt to read a page that was never written. */
```
letFunny: Why is this one here not the same comment about `xRead()`, is it a different situation?

letFunny (approved):
Thanks Marco for the discussion offline, I am now convinced it _seems_ sound :P

src/vfs.c
letFunny: Discussed offline, actually the explanation is quite length so I will summarize it as follows. When restoring a snapshot we need to invalidate the shared memory so it is rebuilt after appending to the WAL. This needs to happen independently of whether there were frames in the WAL or not at the time when restoring the snapshot (obviously). The check above `w->n_frames == 0 => return SQLITE_OK` was NOT correct because it was not changing the header, meaning the shm was not rebuilt. This is a problem we didn't find in production but something that could happen.

In some more detail, if the shared memory is not rebuilt, then the pages appended to the WAL will not be seen by SQLite because the shared memory will contain the wrong length of the WAL.

src/vfs.c
marco6: I'll wait on #772 as it is fixed there already.

just-now (approved):
Looks good.

PR #772, opened 2025-06-18
fix: read main file when out of range

marco6:
This PR fixes the reading of the main file in the case of an "over the limit" read, which might happen when enabling autovacuum.

### DIFF
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -1178,28 +1178,15 @@ static int vfsWalRead(struct vfsWal *w,
 	}
 
 	if (index == 0) {
-		/* From SQLite docs:
-		*
-		*   If xRead() returns SQLITE_IOERR_SHORT_READ it must also fill
-		*   in the unread portions of the buffer with zeros.  A VFS that
-		*   fails to zero-fill short reads might seem to work.  However,
-		*   failure to zero-fill short reads will eventually lead to
-		*   database corruption.
-		*/
+		// This is an attempt to read a page that was
+		// never written.
 		memset(buf, 0, (size_t)amount);
 		return SQLITE_IOERR_SHORT_READ;
 	}
 
 	frame = vfsWalFrameLookup(w, index);
 	if (frame == NULL) {
-		/* From SQLite docs:
-		*
-		*   If xRead() returns SQLITE_IOERR_SHORT_READ it must also fill
-		*   in the unread portions of the buffer with zeros.  A VFS that
-		*   fails to zero-fill short reads might seem to work.  However,
-		*   failure to zero-fill short reads will eventually lead to
-		*   database corruption.
-		*/
+		// Again, the requested page doesn't exist.
 		memset(buf, 0, (size_t)amount);
 		return SQLITE_IOERR_SHORT_READ;
 	}
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -1063,6 +1063,14 @@ static int vfsDatabaseRead(struct vfsDatabase *d,
 	void *page;
 
 	if (d->n_pages == 0) {
+		/* From SQLite docs:
+		*
+		*   If xRead() returns SQLITE_IOERR_SHORT_READ it must also fill
+		*   in the unread portions of the buffer with zeros.  A VFS that
+		*   fails to zero-fill short reads might seem to work.  However,
+		*   failure to zero-fill short reads will eventually lead to
+		*   database corruption.
+		*/
 		memset(buf, 0, (size_t)amount);
 		return SQLITE_IOERR_SHORT_READ;
 	}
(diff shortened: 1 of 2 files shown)


test/unit/test_gateway.c
```diff
@@ -2381,6 +2389,36 @@ TEST_CASE(exec_sql, manyParams, NULL)
...
+TEST_CASE(exec_sql, autovacuum_full, NULL)
+{
+	struct exec_sql_fixture *f = data;
+	(void)params;
+
+	EXEC_SQL("PRAGMA auto_vacuum = FULL; VACUUM");
+	EXEC_SQL(
+		"CREATE TABLE IF NOT EXISTS test(                                   "
+		" id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT                     "
+		" -- When the definition of a table is big                          "
```
just-now, line 2402: Curious if it's worth putting string below in the SQL comment rather then just a comment around CREATE statement?

marco6: You mean moving the comment outside of the `CREATE` statement? If I got that correctly I think it might work, however that string is used to make the `CREATE` command "big". I am not sure if SQLite will split my query in 2 queries (one with just the `CREATE` and the other with just the comment) or if it will actually keep a single one.

src/vfs.c
```diff
@@ -1063,6 +1063,7 @@ static int vfsDatabaseRead(struct vfsDatabase *d,
 	void *page;
 
 	if (d->n_pages == 0) {
+		memset(buf, 0, (size_t)amount);
```
letFunny, line 1074: I think we are missing a comment here and there is probably a dependency with #770.

src/vfs.c
```diff
@@ -1092,6 +1093,11 @@ static int vfsDatabaseRead(struct vfsDatabase *d,
 
 	page = vfsDatabasePageLookup(d, pgno);
 
+	if (page == NULL) {
+		memset(buf, 0, (size_t)amount);
+		return SQLITE_IOERR_SHORT_READ;
```
letFunny, line 1114: Same comment.

test/unit/test_gateway.c
```diff
@@ -2381,6 +2389,36 @@ TEST_CASE(exec_sql, manyParams, NULL)
...
+		" -- enough to create an 'overflow page' and                        "
+		" -- auto_vacuum is enabled, the DROP TABLE                         "
```
letFunny: This comments says `DROP TABLE` can try to access a page number above the size of the db but we are not using that in the test. Can you change the next statement to `DROP TABLE` even if equivalent just to make it more clear.

letFunny (commented):
Thanks for this Marco!

test/unit/test_gateway.c
```diff
@@ -2381,6 +2389,36 @@ TEST_CASE(exec_sql, manyParams, NULL)
...
+		" -- logic can try to access a page number which                    "
+		" -- is above the size of the database. In this                     "
```
letFunny, line 2406: One more question, how do you know this? Can you put a link or was it something you read on the implementation?

test/unit/test_gateway.c
marco6: `DROP INDEX` is the same as `DROP TABLE` as far as the format is concerned.

marco6: Changed.

test/unit/test_gateway.c
marco6: As discussed offline, this come out when I tried to test autovacuum in [`k8s-dqlite`](https://github.com/canonical/k8s-dqlite/compare/master...autovacuum). I managed to reduce the example to this single case.

test/unit/test_gateway.c
just-now: Thanks for explanation.

marco6:
I updated the test @just-now @letFunny. Apparently I can reproduce even without the `DROP TABLE` as long as I still create a new page.

I decided to dig a little bit deeper, not sure if I got it 100%, but the idea should be the following: [here](https://github.com/sqlite/sqlite/blob/981022b/src/btree.c#L10083) (and in the other calls to `btreeGetPage` in there) the logic passes flags `0` (the last parameter), which mean "try to read the page, even if it might not be on disk". This is unusual because when the page is new, usually it is passed with "no-content" flag which will just initialize the page to 0, without asking the VFS. Looking around the explanation for that can be found in the [`allocateBtreePage`](https://github.com/sqlite/sqlite/blob/981022b/src/btree.c#L6696) function, quoting from there:

> Normally, new pages allocated by this block can be requested from the
pager layer with the 'no-content' flag set. This prevents the pager
from trying to read the pages content from disk. However, if the
current transaction has already run one or more incremental-vacuum
steps, then the page we are about to allocate may contain content
that is required in the event of a rollback. In this case, do
not set the no-content flag. This causes the pager to load and journal
the current page content before overwriting it.
>
> Note that the pager will not actually attempt to load or journal
content for any page that really does lie past the end of the database
file on disk. So the effects of disabling the no-content optimization
here are confined to those pages that lie between the end of the
database image and the end of the database file.

Which makes sense, even if I think the path of "journaling" will not be taken in case of WAL-mode (as we are really not journaling).

> the only thing missing is to add a comment describing the bug better

I think that the comment in the SQL command is enough. Do you think we should add more? Is the above rationale enough?

letFunny (approved):
Discussed offline, I am now even more convinced why this works, thanks Marco for PR and explanation!

PR #773, opened 2025-06-20
perf: transactional ordering for callbacks

marco6:
This PR changes the order in which callbacks are applied, in the case an apply manages to commit[^1] more than one entry in a single round.

What happened before was that each entry was applied and its callback was immediately fired (before applying the other entries, despite them being part of the log). The new code applies everything in the current batch *and then* calls (in order) the callbacks.

This helps some situations for example:
 - leadership change (I already know I'm not the leader anymore, as there is a committed entry in the log with a different term, coming from another server, and yet I try to perform a write transaction)
 - barriers (I call for a barrier as I want the latest version, but the latest version is already in the log - it's just not applied yet)

[^1]: commit in the RAFT sense, which means reaching quorum for the leader and being notified of the new commit index for the follower

### DIFF
--- a/src/fsm.c
+++ b/src/fsm.c
@@ -220,8 +220,7 @@ static int apply_checkpoint(struct fsm *f, const struct command_checkpoint *c)
 }
 
 static int fsm__apply(struct raft_fsm *fsm,
-		      const struct raft_buffer *buf,
-		      void **result)
+		      const struct raft_buffer *buf)
 {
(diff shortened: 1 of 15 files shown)


marco6:
As discussed offline, this will wait for #747.

marco6:
Given how rare barriers are once #821 gets in, this is likely a decline.

PR #774, opened 2025-06-20
 perf: improve barrier implementation 

marco6:
This is another PR that aims at reducing the footprint of creating a barrier and in particular aims at reducing the write amplification (and its dependency on an fsync and a quorum) for writes *and* reads.


marco6:
Depends on #773

PR #778, opened 2025-06-30
feat: improve db__open initialization

marco6:
This PR improves the db__open initialization sequence and, more importantly, makes sure that underlying files are open (especially the connection to the WAL) even when opening a database for the first time.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -1,3 +1,4 @@
+#include <sqlite3.h>
 #include <stdint.h>
 #include <string.h>
 
@@ -95,6 +96,14 @@ static int dqlite_authorizer(void *pUserData, int action, const char *third, con
 	return SQLITE_OK;
 }
 
+static int isWalMode(void* out, int argc, char **argv, char **unused)
+{
+	(void)unused;
+	assert(argc == 1);
+	*(int*)out = sqlite3_stricmp("WAL", argv[0]) == 0;
+	return SQLITE_OK;
+}
+
 int db__open(struct db *db, sqlite3 **conn)
 {
 	tracef("open conn: %s page_size:%u", db->path, db->config->page_size);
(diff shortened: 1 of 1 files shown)


src/db.c
```diff
@@ -109,38 +118,47 @@ int db__open(struct db *db, sqlite3 **conn)
...
-	/* Set WAL journaling. */
-	rc = sqlite3_exec(*conn, "PRAGMA journal_mode=WAL", NULL, NULL, &msg);
+	int initialized;
+	rc = sqlite3_exec(*conn, "PRAGMA journal_mode", isWalMode, &initialized, &msg);
 	if (rc != SQLITE_OK) {
-		tracef("journal_mode=WAL failed");
+		tracef("extended codes failed %d", rc);
 		goto err;
 	}
+	if (!initialized) {
```
letFunny, line 134: Isn't it simpler to do this every time we open a db? Does the performance suffer a lot to justify the extra complexity in having two opening sequences?

Example: page size is for some reason different from expected, if `initialized == true` we don't catch it.

Maybe the benefit is for updates on existing databses where we might not want to change the page_size for example, but supporting that scenario with different assumptions about page_size is tricky in itself.

letFunny (commented):
Thanks Marco for this PR, only one question and approving.

src/db.c
marco6: You cannot change page_size in SQLite unless you issue a `VACUUM` after in SQLite. Dqlite on the other side [does not support](https://github.com/canonical/dqlite/blob/improve-db__open/src/vfs.c#L1496) changing page size.

But the reason for this is clearly not performance: we need to make sure a connection to the WAL is opened, and the only way to do that is to start and rollback an empty transaction. However, we don't want to that blindly as another transaction might be happening for existing databases and the command `BEGIN IMMEDIATE; ROLLBACK` will try to acquire:
 - the `DELETE` mode `EXCLUSIVE` lock (for the transition `DELETE` -> `WAL`)
 - the `WAL` mode `WRITE` lock
 
 So, we have two choices here: write very complex code to sync with the `leader_exec` logic or do that only once during initialization, in a case where we are 100% sure we are the only users of that database as it wasn't there before and we are running on the main loop, which is the only one that can create databases.

letFunny: > we need to make sure a connection to the WAL is opened

Why? Isn't the first transaction going to open it anyway?

Agree roughly about the rest.

letFunny (approved):
Discussed offline, thanks Marco for walking me through it.

PR #780, opened 2025-07-01
test: fix wait times for gateway tests

marco6:
This PR fixes a bunch of tests in the gateway suite so that they always wait (where appropriate) and do not rely on some async operation to end synchronously.

It also adds the capability to change the expected async work duration so that it is possible to rely on that timing during test time.

### DIFF
--- a/test/unit/test_concurrency.c
+++ b/test/unit/test_concurrency.c
@@ -285,7 +285,8 @@ TEST_CASE(exec, busy_wait_statement, NULL)
 {
 	struct exec_fixture *f = data;
 	(void)params;
-	
+
+	raft_fixture_set_work_duration(&f->cluster, 0, 50);
 	f->servers[0].config.busy_timeout = 100;
 
 	/* Create a test table using connection 0 */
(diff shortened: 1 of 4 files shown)


test/unit/test_concurrency.c
```diff
@@ -285,7 +285,8 @@ TEST_CASE(exec, busy_wait_statement, NULL)
 {
 	struct exec_fixture *f = data;
 	(void)params;
-	
+
+	raft_fixture_set_work_duration(&f->cluster, 0, 50);
 	f->servers[0].config.busy_timeout = 100;
```
letFunny, line 290: Weird that this is done manually.

letFunny (approved):

test/unit/test_concurrency.c
marco6: Eh the problem when I wrote those tests in #740 is that `config` is supposed to be tweaked *before* starting the node as it is not guarded by atomics/locks and changes to it could be partially read.

So, either I added this "hack" or I changed the initialization sequence which is IMHO way harder to do.

letFunny: Understandable, as long as the tests are not finicky then it is okay. I meant we are creating a function function when in the past we did it inline, but it is not important.

PR #781, opened 2025-07-01
perf: improve barrier request logic

marco6:
This PR improves the logic around barrier request so that barriers are not requested for already started transactions as sequencing is not necessary anymore for them.

### DIFF
--- a/src/leader.c
+++ b/src/leader.c
@@ -36,6 +36,11 @@ static void exec_enqueue(struct db *db, struct exec *exec);
  * index. */
 static bool exec_needs_barrier(struct leader *l)
 {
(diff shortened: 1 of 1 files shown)


marco6:
TBF I actually think that the barrier shouldn't be there for writes, but for reads. Writes will either fail because of lost leadership or because a write lock is in place and the time in the queue runs out. The only case where we would need a barrier for writes is at the beginning of a term, where the leader does not know which part of the log are actually committed.

But again, I don't want to change too much the logic, just slightly improve it. We can put this PR on hold and see the effects on real use cases when #747 is in.

marco6:
This is superseded by #821.

PR #782, opened 2025-07-02
fix: enable foreign key support by default

marco6:
Following the report from LXD, this PR re-enables foreign key support by default.

It also adds a test as clearly that was a hidden promise our interface did which wasn't tested.

### DIFF
--- a/test/unit/test_gateway.c
+++ b/test/unit/test_gateway.c
@@ -2450,22 +2450,6 @@ TEST_CASE(exec_sql, foreignKeys, NULL)
 	/* Make sure I can insert proper data in the table */
 	EXEC_SQL("INSERT INTO b(a_id) VALUES (1), (1), (2)");
 
-	/* Make sure that cascading works */
-	EXEC_SQL("DELETE FROM a");
-	struct value value;
-	uint64_t n;
-	text_t column;
-	QUERY_SQL_SUBMIT("SELECT COUNT(*) FROM b");
-	WAIT;
-	ASSERT_CALLBACK(0, ROWS);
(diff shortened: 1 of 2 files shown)


tomponline:
Thanks!

test/unit/test_gateway.c
```diff
@@ -2427,6 +2427,48 @@ TEST_CASE(exec_sql, autovacuum_full, NULL)
...
+	QUERY_SQL_SUBMIT("SELECT COUNT(*) FROM b");
+	WAIT;
+	ASSERT_CALLBACK(0, ROWS);
+	uint64__decode(f->cursor, &n);
+	munit_assert_int(n, ==, 1);
+	text__decode(f->cursor, &column);
+	munit_assert_string_equal(column, "COUNT(*)");
+	DECODE_ROW(1, &value);
+	munit_assert_int(value.type, ==, SQLITE_INTEGER);
+	munit_assert_int(value.integer, ==, 0);
```
letFunny: Why do we need this? Aren't we only testing foreign_keys?

letFunny (commented):
Thanks Marco for being so quick

test/unit/test_gateway.c
marco6: Cascading is also enabled by foreign keys.

letFunny: Sure but the upper part already fails if foreign keys are not enabled. I am just trying to keep the test as minimal as possible.

marco6: Ok I'll remove that

letFunny (approved):

PR #783, opened 2025-07-02
refactor: improve `Vfs{Poll,Abort,Apply}` interface

marco6:
This PR aims at improving the interface for `Vfs{Poll,Abort,Apply}` in such a way that it uses a connection instead of fetching every time the VFS through a linear search in the registry.


src/fsm.c
```diff
@@ -93,19 +85,12 @@ static void maybeCheckpoint(struct db *db)
...
 	/* Get the database file associated with this db->follower connection */
-	rv = sqlite3_file_control(conn, "main",
-				  SQLITE_FCNTL_FILE_POINTER, &main_f);
-	assert(rv == SQLITE_OK); /* Should never fail */
-
-	/* Get the first SHM region, which contains the WAL header. */
-	rv = main_f->pMethods->xShmMap(main_f, 0, 0, 0, &region);
-	assert(rv == SQLITE_OK); /* Should never fail */
-
-	rv = main_f->pMethods->xShmUnmap(main_f, 0);
```
letFunny, line 108: I want to confirm my understanding per our discussion last time. This was needed only so that the shared memory was initialized (in fact `xShmUnmap` is not needed at all). Now, we don't need it anymore because when we open a connection to the db we are initializing it by virtue of running `BEGIN IMMEDIATE; COMMIT` per #778.

src/fsm.c
```diff
@@ -159,30 +140,28 @@ static int apply_frames(struct fsm *f, const struct command_frames *c)
...
 			tracef("open follower failed %d", rv);
 			return rv;
 		}
-		sqlite3_close(conn);
 	}
 
-	/* If the commit marker must be set as otherwise this must be an
+	/* The commit marker must be set as otherwise this must be an
 	 * upgrade from V1, which is not supported anymore. */
 	if (c->is_commit) {
```
letFunny: Suggestion: Let's invert the condition `if (!c->is_commit) { rv = DQLITE_PROTO; goto error; }` and remove one level of indentation in the happy path.

src/vfs.c
```diff
@@ -2246,98 +2246,74 @@ void VfsClose(struct sqlite3_vfs *vfs)
...
-		return SQLITE_NOTFOUND;
-	}
-
-	shm = &database->shm;
-	wal = &database->wal;
-
-	if (wal == NULL) {
-		*transaction = (struct vfsTransaction) {};
-		return 0;
-	}
```
letFunny, line 2322: Again, to check assumptions, why could this be NULL, when was `database->wal` supposed to be NULL? Is this something you changed in the initialization sequence or has it always been unnecessary check?

test/unit/test_vfs_extra.c
```diff
@@ -195,65 +195,52 @@ static void vfsFillDbPath(struct fixture *f,
 
 /* Poll the given VFS object and serialize the transaction data into the given
  * tx object. */
-#define POLL(VFS, TX)                                              \
-	do {                                                           \
-		sqlite3_vfs *vfs = sqlite3_vfs_find(VFS);                  \
-		memset(&(TX), 0, sizeof(TX));                              \
```
letFunny, line 201: memset not needed anymore?

test/unit/test_vfs_extra.c
```diff
@@ -464,7 +451,7 @@ TEST(vfs_extra, pollAfterWriteTransaction, setUp, tearDown, 0, vfs_params)
 	PREPARE(db, stmt, "CREATE TABLE test(n INT)");
 	STEP(stmt, SQLITE_DONE);
 
-	POLL("1", tx);
+	POLL(db, tx);
```
letFunny, line 454: Why does OPEN() initialize the db manually and set the WAL mode, page_size etc. couldn't we use db__open instead? It seems far less fragile and we would test the actual production initialization more.

marco6: Because this tests are meant to be unit test of the `vfs` layer and not of an upper layer like `db`.

TBH it is a bit strange as part of the vfs requirement is explicited in the authorizer (which is set in the `db__open`). Also some logic from the upper layer is somehow slipping here (for example, we are changing this because we know upper layer needs better isolation when doing multithreading...)

I think that in general the reason is that the abstraction is a lot leakier than this test think it is.

test/unit/test_vfs_extra.c
marco6: `VfsPoll` always sets the transaction, so (re-)initialization is somehow not needed anymore.

src/vfs.c
marco6: No, this check was simply useless and likely removed by the compiler: 

```C
wal = &database->wal;
```

That line is taking the address of a member of a struct known to be `!= NULL` (it is adding basically the offset of `wal` to the address pointed by `database`). As such, it can't ever be `0`.

test/unit/test_vfs_extra.c
letFunny: > I think that in general the reason is that the abstraction is a lot leakier than this test think it is.

Even if the test is meant to be a "unit test", it would be better to call db open and then get the raw connection if needed. On principle we should strive to use the real initialization, close down sequences, etc. in as much code as possible to prevent bugs slipping because different assumptions in different test suites.

src/vfs.c
letFunny: :)

src/fsm.c
marco6: Kind of. It will force to map (and create) the shared memory file (not the WAL) and that is necessary for the `xShmLock` calls to succeed (as they lock some bytes in that file). 

As you said, this is taken care now by making sure that we always have a connection to the WAL.

TBH I'm like 99% sure that the double-open that we discussed (one in `VfsApply`, one in `maybeCheckpoint`) would open a connection to the WAL, considering that the code above checks the size of the WAL! So I think it was just useless.

test/unit/test_vfs_extra.c
marco6: Can we do that in another PR? I think that the `OPEN` is not really related here...

marco6: As discussed on MM, there are benefits for using the "real" logic, but that requires a small refactor. It will happen after multithreading.

letFunny (approved):

PR #785, opened 2025-07-03
chore: fix pipeline

marco6:
This PR tries to fix the PPA publishing pipeline by running the job on 22.04 instead of 20.04. This will not work for the LTS. Please see #784 for a PR that does the same for the LTS 1.17.x.

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -14,7 +14,7 @@ jobs:
           - jammy
           - noble
           - oracular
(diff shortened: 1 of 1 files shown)


letFunny (commented):
If ubuntu-22.04 doesn't work shouldn't we use a container with ubuntu-20.04 instead? (github runner on ubuntu-22.04 running a container of ubuntu-20.04 if that makes sense)

PR #786, opened 2025-07-03
chore: fix test pipeline

marco6:
This PR fixes the pipeline *for the LTS branch*.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -16,10 +16,16 @@ jobs:
         compiler:
           - gcc
           - clang
-    runs-on: ${{ matrix.os }}
-
+    runs-on: ${{ matrix.os == 'ubuntu-20.04' && 'ubuntu-22.04' || matrix.os }}
+    container: ${{ matrix.os == 'ubuntu-20.04' && 'ubuntu:20.04' || '' }}
     steps:
     - uses: actions/checkout@v4
+
+    - name: Install sudo in container
+      if: ${{ matrix.os == 'ubuntu-20.04' }}
+      run: |
+          apt-get update
+          DEBIAN_FRONTEND=noninteractive apt-get install -y sudo tzdata build-essential autoconf clang automake libtool pkg-config
 
     - name: Set up dependencies
       run: |
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -16,16 +16,18 @@ jobs:
         compiler:
           - gcc
           - clang
-    runs-on: ${{ matrix.os == 'ubuntu-20.04' && 'ubuntu-22.04' || matrix.os }}
-    container: ${{ matrix.os == 'ubuntu-20.04' && 'ubuntu:20.04' || '' }}
+    runs-on: ${{ (matrix.os == 'ubuntu-20.04' && 'ubuntu-22.04') || matrix.os }}
+    container: ${{ (matrix.os == 'ubuntu-20.04' && 'ubuntu:20.04') || '' }}
     steps:
     - uses: actions/checkout@v4
 
     - name: Install sudo in container
       if: ${{ matrix.os == 'ubuntu-20.04' }}
+      env:
+          DEBIAN_FRONTEND: noninteractive
       run: |
           apt-get update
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -16,11 +16,17 @@ jobs:
         compiler:
           - gcc
           - clang
-    runs-on: ${{ matrix.os }}
-
+    runs-on: ${{ matrix.os == 'ubuntu-20.04' && 'ubuntu-22.04' || matrix.os }}
+    container: ${{ matrix.os == 'ubuntu-20.04' && 'ubuntu:20.04' || '' }}
```
letFunny: Can you add parenthesis? I presume this syntax works because ` matrix.os == 'ubuntu-20.04' && 'ubuntu-22.04'` evaluates to the second argument if the condition is true (which I didn't know btw). Could it instead be:
```
${{ (matrix.os == 'ubuntu-20.04' && 'ubuntu-22.04') || matrix.os }}
```

.github/workflows/build-and-test.yml
```diff
@@ -16,11 +16,17 @@ jobs:
...
     steps:
     - uses: actions/checkout@v4
 
+    - name: Install sudo in container
+      if: ${{ matrix.os == 'ubuntu-20.04' }}
```
letFunny, line 25: In general and I commented something similar in another PR I would prefer if we had the same process for all different versions. So why not always run on "ubuntu-22.04" and then on a container with each version and this step always runs?

By having less differences the build process is simpler to review and there are less surprises due to deviations and misconfigurations.

.github/workflows/build-and-test.yml
marco6: > evaluates to the second argument if the condition is true

That is the "truthy" way of doing thing is JS, you can try in your browser lol. This also tell us that likely it is nodejs under the hood of workflows.

I will update that matrix even if it shows a failure which will need to be fixed in a separate PR (unit-test not passing, but finally running).

.github/workflows/build-and-test.yml
marco6: The reasoning it that I really want to keep the tests running on a VM and not on a container. However, that is not possible for `20.04` so, for that alone, I would be ok to run in a container, as a work-around.

letFunny: > The reasoning it that I really want to keep the tests running on a VM and not on a container

Can I ask why? And if it justifies the extra complexity of two different build processes with possibly different tools installed?

.github/workflows/build-and-test.yml
letFunny: Yes, I still prefer having a parenthesis for easier readability.

.github/workflows/build-and-test.yml
marco6: Containers can behave in special ways and have no way to touch some part of the OS if ever needed. This is more a defensive pov rather than "in this case it will not work". It is more like: it worked so far, let's not change what's working if there is no need.

marco6: An [example](https://github.com/actions/runner-images/issues/9524) of things that can go wrong with containers, happening on my NUC with this very repo.

.github/workflows/build-and-test.yml
marco6: Done

.github/workflows/build-and-test.yml
letFunny: AFAIK the link you shared is actually on the GH runner image, so the VM. We are going to have that bug regardless, the discussion is vm+container vs vm.

>  it worked so far, let's not change what's working if there is no need.

This is true IMO, but to me the benefits outweigh the drawback, the vm is already virtualized, not sure if the container makes a difference.

marco6: > AFAIK the link you shared is actually on the GH runner image, so the VM

It is on the VM, but running the workaround in the container doesn't work while on the VM it does (clearly, allowing a container change a kernel parameter would be a huge vulnerability).

letFunny (approved):
Discussed offline about the having two build processes (vm+container for ubuntu-20.04 vs only vm for the rest). I see the arguments to justify it, I disagree and commit, there is no clear option, approved, thanks!

PR #787, opened 2025-07-04
fix: retry fallocate if EINTR is returned

marco6:
This PR tries again to call `posix_fallocate` in case the system call was interrupted by a signal.


marco6:
Note: this will likely need to be backported.

PR #789, opened 2025-07-07
test: fix checkpoint test

marco6:


test/unit/test_vfs_extra.c
```diff
@@ -1188,23 +1188,26 @@ TEST(vfs_extra, applyOnDifferentVfsCheckpointReclaimsSpace, setUp, tearDown, 0,
 	rv = sqlite3_file_control(db2, "main", SQLITE_FCNTL_FILE_POINTER, &main_f);
 	assert(rv == SQLITE_OK);
 	CHECKPOINT(db2);
-
 	rv = main_f->pMethods->xFileSize(main_f, &pre_vacuum_size);
 	assert(rv == SQLITE_OK);
+	CLOSE(db2);
```
letFunny, line 1193: Do you need to close because Apply() doesn't expect a connection to be opened?

letFunny (approved):
Just a question Marco

test/unit/test_vfs_extra.c
marco6: I need to `CLOSE` because the second time you call `CHECKPOINT` the version of SQLite in `20.04` will try to acquire the exclusive lock twice. This is actually not a bug: it is explicitly required for VFSs to not fail if the lock is already held by the connection (= if conn1 acquires twice the `EXCL` lock is ok, if conn1 and conn2 try to acquire it one each, then one of them should fail).

This is not supported by the old VFS and I have no plan on supporting it, considering that in the code of this branch a connection is opened just to checkpoint.

letFunny: I see so it is a bug in our code, can you write a comment there saying what you told me here?

PR #790, opened 2025-07-07
refactor: use an in-memory file to serve shared memory

marco6:
This PR removes the previous implementation of malloc-backed shared memory with a memfd-backed one.

It will also add state to each file descriptor to track which lock was acquired by the current connection, so that it can correctly implement the semantics of `xShmLock` (which require subsequent calls to lock acquire to be no-ops if the lock is already taken by that connection). Most of the logic on this check, as well as the names for it, is copied 1-1 from SQLite source code (see [`os_unix.c`](https://github.com/sqlite/sqlite/blob/master/src/os_unix.c#L4374)).


src/vfs.c
```diff
@@ -255,116 +260,72 @@ static void vfsFrameDestroy(struct vfsFrame *f)
...
 struct vfsShm
 {
-	void **regions;     /* Pointers to shared memory regions. */
-	int n_regions; /* Number of shared memory regions. */
-	unsigned refcount;  /* Number of outstanding mappings. */
-	unsigned shared[SQLITE_SHM_NLOCK];    /* Count of shared locks */
-	unsigned exclusive[SQLITE_SHM_NLOCK]; /* Count of exclusive locks */
+	int fd;
+	int size;
+	int lock[SQLITE_SHM_NLOCK];  /* -1==excl lock */
```
letFunny: I think we should be more verbose on the documentation, let's say that positive indexes indicate that the lock is taking by n readers, 0 lock is free, -1 lock is taken in an exclusive form. Do you think it is worth it or it is expected from sqlite and documentation is redundant?

src/vfs.c
```diff
@@ -255,116 +260,72 @@ static void vfsFrameDestroy(struct vfsFrame *f)
...
-	unsigned *these_locks;
-	unsigned *other_locks;
-	int i;
-
-	if (flags & SQLITE_SHM_SHARED) {
-		these_locks = s->shared;
-		other_locks = s->exclusive;
-	} else {
-		these_locks = s->exclusive;
-		other_locks = s->shared;
```
letFunny, line 324: The new implementation is much cleaner and easier to follow, thanks for this Marco!

src/vfs.c
```diff
@@ -255,116 +260,72 @@ static void vfsFrameDestroy(struct vfsFrame *f)
...
-	}
-}
-
-/* Revert the shared mamory to its initial state. */
-static void vfsShmReset(struct vfsShm *s)
-{
-	vfsShmClose(s);
-	vfsShmInit(s);
+	int rv = close(s->fd);
+	assert(rv == 0);
```
letFunny: I assume that there is no way that this is failing but do we want to panic in the shutdown sequence if this doesn't work?

src/vfs.c
```diff
@@ -561,8 +533,10 @@ static int vfsDatabaseGetPage(struct vfsDatabase *d,
 	bool pending_byte_page_reached =
 	    (page_size * d->n_pages == dq_sqlite_pending_byte);
 	if ((pgno > d->n_pages + 1) && !pending_byte_page_reached) {
-		rc = SQLITE_IOERR_WRITE;
-		goto err;
+		tracef("tried to write far from end (requested %d, last %d)",
```
letFunny: What is "far" in this context, I would say `page number greater than length` or something similar.

src/vfs.c
```diff
@@ -561,8 +533,10 @@ static int vfsDatabaseGetPage(struct vfsDatabase *d,
...
+		       pgno, d->n_pages);
+		*page = NULL;
+		return SQLITE_IOERR_WRITE;
```
letFunny, line 547: How did you find out this was wrong? Was it in the multithreading PR?

src/vfs.c
```diff
@@ -697,37 +675,28 @@ static void vfsAmendWalIndexHeader(struct vfsDatabase *d)
...
-	    checksum[0];
-	*(uint32_t *)__builtin_assume_aligned(&index[44], sizeof(uint32_t)) =
-	    checksum[1];
+	*(uint32_t *)(&index[40]) = checksum[0];
+	*(uint32_t *)(&index[44]) = checksum[1];
 
 	/* Update the second copy of the first part of the WAL index header. */
 	memcpy(index + VFS__WAL_INDEX_HEADER_SIZE, index,
 	       VFS__WAL_INDEX_HEADER_SIZE);
+	rv = pwrite(shm->fd, index, sizeof(index), 0);
```
letFunny, line 706: Maybe I am wrong but if the line above is updating the second copy of the WAL index header, shouldn't the prwrite here to be of `sizeof(index) * 2`?

src/vfs.c
```diff
@@ -1506,64 +1452,61 @@ static int vfsMainFileShmMap(sqlite3_file *file, /* Handle open on database file
...
-			assert(region_size == VFS__WAL_INDEX_REGION_SIZE);
-			assert(region_index == s->n_regions);
-			region = sqlite3_malloc64((sqlite3_uint64)region_size);
-			if (region == NULL) {
-				rv = SQLITE_NOMEM;
-				goto err;
-			}
 
-			memset(region, 0, (size_t)region_size);
+	PRE(region_size == VFS__WAL_INDEX_REGION_SIZE);
```
letFunny, line 1467: Nice! I was thinking of this before seeing the line, it is quite clear that we record properties and that it is readable for reviewer.

src/vfs.c
```diff
@@ -1595,60 +1538,90 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
-		/* When acquiring the write lock, make sure there's no
-		 * transaction that hasn't been rolled back or polled. */
-		if (flags == (SQLITE_SHM_LOCK | SQLITE_SHM_EXCLUSIVE)) {
-			assert(wal->n_tx == 0);
+	PRE(flags == (SQLITE_SHM_LOCK | SQLITE_SHM_SHARED) ||
+	    flags == (SQLITE_SHM_LOCK | SQLITE_SHM_EXCLUSIVE) ||
+	    flags == (SQLITE_SHM_UNLOCK | SQLITE_SHM_SHARED) ||
+	    flags == (SQLITE_SHM_UNLOCK | SQLITE_SHM_EXCLUSIVE));
+
+	PRE((f->exclMask & f->sharedMask) == 0);
```
letFunny, line 1567: I may be wrong here but I believe this is only checked in sqlite when SHM_UNLOCK is set, see https://github.com/sqlite/sqlite/blob/master/src/os_unix.c#L5109.

src/vfs.c
```diff
@@ -1595,60 +1538,90 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+
+	int rv = SQLITE_OK;
+	if (((flags & SQLITE_SHM_UNLOCK) &&
+	     ((f->exclMask | f->sharedMask) & mask)) ||
+	    (flags == (SQLITE_SHM_SHARED | SQLITE_SHM_LOCK) &&
+	     0 == (f->sharedMask & mask)) ||
+	    (flags == (SQLITE_SHM_EXCLUSIVE | SQLITE_SHM_LOCK))) {
```
letFunny: This is not very readable IMO, can you group conditions with a parenthesis in the same line like in the sqlite source code? We may need to turn clang-format off for this to happen.

src/vfs.c
```diff
@@ -1595,60 +1538,90 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+					vfsWalRollbackIfUncommitted(
+					    &f->database->wal);
+				}
+			}
+		} else if (flags & SQLITE_SHM_SHARED) {
+			rv = vfsShmLock(&f->database->shm, ofst, n, false);
+			if (rv == SQLITE_OK) {
+				f->sharedMask |= mask;
+			}
+		} else {  // (flags & SQLITE_SHM_EXCLUSIVE)
```
letFunny: IMO `flags & SQLITE_SHM_EXCLUSIVE` is better as an assert and works as documentation which is also what sqlite does.

src/vfs.c
```diff
@@ -1595,60 +1538,90 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+			PRE((f->exclMask & mask) == 0);
+
+			/* When acquiring a write lock, make sure there's no
+			 * transaction that hasn't been rolled back or polled.
+			 * If there is, just return SQLITE_BUSY. */
+			// FIXME: this check is strange because we need to
+			// support a weird way to check for checkpointing.
+			if (ofst == VFS__WAL_WRITE_LOCK &&
+			    f->database->wal.n_tx > 0) {
+				rv = SQLITE_BUSY;
```
letFunny, line 1616: We are changing the behavior here, if I am not wrong the old code used to crash an assert here instead.

src/vfs.c
```diff
@@ -1595,60 +1538,90 @@ static void vfsWalRollbackIfUncommitted(struct vfsWal *w)
...
+			}
 		}
-		/* When releasing the write lock, if we find a pending
-		 * uncommitted transaction then a rollback must have occurred.
-		 * In that case we delete the pending transaction. */
-		if (flags == (SQLITE_SHM_UNLOCK | SQLITE_SHM_EXCLUSIVE)) {
-			tracef("ROLLBACK TIME");
-			vfsWalRollbackIfUncommitted(wal);
+
+		if (rv == SQLITE_OK && ofst == VFS__WAL_WRITE_LOCK) {
```
letFunny, line 1628: I guess this is meant to preserve the old behavior and to separate it from the condition in the if statement, can you add a comment here to say the reason and how it is linked?

src/vfs.c
```diff
@@ -1664,11 +1637,16 @@ static int vfsMainFileShmUnmap(sqlite3_file *file, int delete_flag)
...
 	(void)delete_flag;
 	struct vfsMainFile *f = (struct vfsMainFile *)file;
-	struct vfsShm *s = &f->database->shm;
-	s->refcount--;
-	if (s->refcount == 0) {
-		vfsShmReset(s);
+	for (int i = 0; i < f->mappedShmRegions.len; i++) {
+		if (f->mappedShmRegions.ptr[i] != NULL) {
+			int rv = munmap(f->mappedShmRegions.ptr[i], VFS__WAL_INDEX_REGION_SIZE);
+			assert(rv == 0);
```
letFunny, line 1654: Same comment about failing, similar to checking `close()` in the shutdown path.

src/vfs.c
```diff
@@ -2438,23 +2416,22 @@ static void vfsInvalidateWalIndexHeader(struct vfsDatabase *d)
...
 
-	if (shm->n_regions == 0) {
-		return;
-	}
-
-	uint8_t *header = shm->regions[0];
-
-	assert(shm->exclusive[VFS__WAL_WRITE_LOCK] == 0);
-	assert(shm->exclusive[VFS__WAL_CKPT_LOCK] == 0);
-	assert(shm->exclusive[VFS__WAL_RECOVER_LOCK] == 0);
```
letFunny, line 2449: Why is this change needed? How did you find out?

For context there are two changes, checking the number of regions and early return, and the asserts using `==` instead of `>=`.

src/vfs.c
```diff
@@ -2492,11 +2469,12 @@ int VfsApply(sqlite3 *conn, const struct vfsTransaction *transaction)
 	 * originated the transaction (this can happen for example when applying
 	 * a Raft barrier and replaying the Raft log in order to serve a request
 	 * of a newly connected client). */
-	if (f->database->shm.exclusive[0] == 1) {
-		f->database->shm.exclusive[0] = 0;
+	// FIXME
```
letFunny: Context?

src/vfs.c
```diff
@@ -2438,23 +2416,22 @@ static void vfsInvalidateWalIndexHeader(struct vfsDatabase *d)
...
 	 * the first byte of each of the two copies is enough to make the check
 	 * fail. */
-	header[0] = 1;
-	header[VFS__WAL_INDEX_HEADER_SIZE] = 0;
+	uint8_t buffer[1] = { 1 };
+	ssize_t rv = pwrite(shm->fd, buffer, 1, 0);
+	assert(rv == 1);
+
+	buffer[0] = 0;
+	rv = pwrite(shm->fd, buffer, 1, VFS__WAL_INDEX_HEADER_SIZE);
```
letFunny, line 2445: We should assert shm->size >= VFS_WAL_INDEX_HEADER_SIZE * 2, right?

src/vfs.c
```diff
@@ -3061,38 +3059,51 @@ int VfsDiskRestore(sqlite3_vfs *vfs,
...
+	/* Lock the database. The locking scheme here is similar to the one used
+	 * when transitioning from WAL to DELETE mode. The WAL-Index recovery is
+	 * not enough as it lets connections read the content of the main file.
+	 * This means that:
+	 *  - all WAL-Index locks must be held exclusively, including READ(0).
+	 *  - a exclusive lock must be held on the main file.
+	 * Given that only WAL mode is supported, a lock on the database object
+	 * should be enough to emulate the file lock.
+	 */
+	rv = vfsShmLock(&database->shm, 0, SQLITE_SHM_NLOCK, true);
```
letFunny, line 3090: Why the change? Was the earlier logic incorrect?

src/vfs.c
```diff
@@ -3061,38 +3059,51 @@ int VfsDiskRestore(sqlite3_vfs *vfs,
...
 	if (rv != 0) {
 		tracef("database restore failed %d", rv);
-		return rv;
+		goto err_locked;
 	}
 
-	vfsInvalidateWalIndexHeader(database);
+	rv = ftruncate(database->shm.fd, 0);
+	assert(rv == 0);
+	database->shm.size = 0;
```
letFunny, line 3103: Why the change? What happens if we truncate the size of the fd underneath the database? And if it is equivalent, why is it better than invalidating?

test/unit/test_vfs.c
```diff
@@ -235,7 +235,7 @@ static uint32_t __wal_idx_mx_frame(sqlite3 *db)
 	rc = sqlite3_file_control(db, NULL, SQLITE_FCNTL_FILE_POINTER, &file);
 	munit_assert_int(rc, ==, SQLITE_OK);
 
-	rc = file->pMethods->xShmMap(file, 0, 0, 0, &region);
+	rc = file->pMethods->xShmMap(file, 0, 32768, 0, &region);
```
letFunny: Why this number for page size? And if it is meaningful, can we make it a constant?

test/unit/test_vfs.c
```diff
@@ -1147,7 +1147,7 @@ TEST(VfsTruncate, misaligned, setUp, tearDown, 0, NULL)
 
 SUITE(VfsShmMap);
 
-static char *test_shm_map_oom_delay[] = {"0", "1", NULL};
+static char *test_shm_map_oom_delay[] = {"0", NULL};
```
letFunny, line 1151: Do you know if this is used? I don't see `params` used below

test/unit/test_vfs.c
```diff
@@ -1227,7 +1230,8 @@ TEST(VfsShmLock, sharedBusy, setUp, tearDown, 0, vfs_params)
 TEST(VfsShmLock, exclBusy, setUp, tearDown, 0, vfs_params)
 {
 	struct fixture *f = data;
-	sqlite3_file *file = munit_malloc(f->vfs.szOsFile);
+	sqlite3_file *file1 = munit_malloc(f->vfs.szOsFile);
+	sqlite3_file *file2 = munit_malloc(f->vfs.szOsFile);
```
letFunny, line 1235: It is a bit concerning that there is only one test, for one use-case, that tests having two different connections accessing the locks.

letFunny (commented):
Thanks for this Marco, it's finally taking shape. I asked a bunch of questions because I see different properties in the previous implementation and in the new one, not necessarily that the new one is wrong, but I want to understand why it is different.

src/vfs.c
marco6: Does it matter? If during lock de-escalation that can't be true, than it can't be true for escalation (= lock) as well.

src/vfs.c
marco6: Reading the [manual](https://man7.org/linux/man-pages/man2/close.2.html), among the errors, the only possible one is `EINTR`, and we shall not retry after that. I will change this to `assert((rv == 0) || (errno == EINTR))`.

src/vfs.c
marco6: The `index` variable contains both versions of the header:

```c
uint8_t index[VFS__WAL_INDEX_HEADER_SIZE * 2];
```

Indeed, we modify the first copy, duplicate that thanks to the `memcpy` on line `696` and then write out both copies as a single `pwrite`. Also, in general, reading `sizeof(XXX) * 2` is a buffer over-read. as the size of `XXX` is `sizeof(XXX)`.

src/vfs.c

src/vfs.c
marco6: This is a bit different than `close` as the enumeration of [possible errors](https://man7.org/linux/man-pages/man2/munmap.2.html) is different (i.e. `EINTR` is surprisingly not part of the list):

```
       EACCES not possible when unmapping (the access has already taken place)
       EAGAIN not possible when unmapping
       EBADF  not possible when unmapping
       EEXIST not possible when unmapping
       EINVAL only possible if there is a bug.
       EINVAL not possible when unmapping
       ENFILE not possible when unmapping
       ENODEV not possible when unmapping
       ENOMEM not possible when unmapping, as we don't unmap regions in the middle of existing regions.
       EOVERFLOW not possible for x64 machines
       EPERM not possible when unmapping
       ETXTBSY not possible when unmapping
```

src/vfs.c
marco6: This was a comment I used to remind me that later this will need to change. There is no need for it now. Removed.

src/vfs.c
marco6: Right, the call is already guarded by:

```
		if (f->database->shm.size > 0) {
```

And the `shm.size` can only grow by at least `VFS__WAL_INDEX_REGION_SIZE` which is above the size of `VFS__WAL_INDEX_HEADER_SIZE * 2`, but still, it makes sense as a documentation point. Added.

test/unit/test_vfs.c
marco6: This number is `VFS__WAL_INDEX_REGION_SIZE`, and it is the only acceptable value for mapping. The mapping logic before wasn't taking into account the size, just the offset, and assumed that the size was `VFS__WAL_INDEX_REGION_SIZE`. So, in this sense, the test was wrong (= if a normal VFS was asked to do this, it would return error).

marco6: I'll name a constant in the test.

test/unit/test_vfs.c
marco6: it is used in `VfsShmMap/oom` test. In general, given that we compile with `-Werror -Wall -Wextra` and that enables `-Wunused-value`, it is not possible to have values which are not used as that would lead to compilation error.

src/vfs.c
marco6: I added something more verbose.

src/vfs.c
marco6: No, I just followed the SQLite guide with its requirements.

src/vfs.c
marco6: True. This comes from multithreading, meaning that it is not strictly a behaviour for this PR: the checks performed by the `maybeCheckpoint` are enough only if you are sure that the thread performing those is the only one running queries: it checks for locks, but right after the check another transaction might happen!

I also think that it is not this important in the next PR as it will never release the lock to begin with...

src/vfs.c
marco6: The change is needed as before this change we had 2 arrays of bools:
 - one for exclusive.
 - one for shared.
 
 So, `0` meant `untaken`. So the logic there reads: "if the exclusive lock is not taken" which is equivalent to "either the lock is not taken at all, or a shared one is" which is what `>=` means in this context.

letFunny: You are right, my brain totally forgot that when coming back from the coffee break.

marco6:
This time I also tested LXD on this PR. Tests seem [happy](https://github.com/canonical/lxd/pull/15954) for now.

src/vfs.c
marco6: As discussed, this is to make sure no transaction is happening during a `VfsRestore`. A `SQLITE_BUSY` here will force the re-send of the entire snapshot, but it's better than corrupting data.

src/vfs.c
letFunny: Ah okay, so it checks the locks -> grab the locks -> do the logic. Maybe a better idea would be to attempt to grab the locks directly without checking first. Anyway, thanks for the context.

src/vfs.c
marco6: When restoring, there should be no open connection. As such it is a nice place to restart the shared memory as well, considering that otherwise it can only grow.

As discussed offline, now this is also guarded by a check that no connection is alive.

src/vfs.c
letFunny: The difference is initialization and assuming that no action can change the masks outside of this function. But this are not really important for us, I was only saying it so that we match the sqlite code, but treat it as a nitpick.

test/unit/test_vfs.c
letFunny: Agreed offline this test is not that useful, but let's keep it anyway.

letFunny (approved):
Thanks Marco, looks good to me now!

PR #797, opened 2025-07-29
chore: refactor threadpool tests

marco6:
This PR refactors the threadpool tests so that:
 - timers are not initialized multiple times;
 - no global object is used (it can run with `--no-fork` now)

This PR was a part of the effort of trying to fix the tests on ARM.

However, at the end of this effort it become clear that the problem somehow lays in how `alert` works in `22.04` (which seems to be a bug or a different behaviour in user-space) and that resetting the alert is enough to make things work (see #806).

As such, I keep this effort here, given that I already spent this time.


marco6:
Closing in favour of #808.

PR #801, opened 2025-08-04
chore: remove disk mode

### LINKED ISSUE #368: dqlite disk backend
I'm currently working on a disk backend for dqlite, it's in a proof-of-concept phase right now but wanted to start discussion on it.

My initial idea is/was to keep the WAL in-memory, this allows us to easily reuse the the replication logic and store the DB on disk in a regular sqlite3 database file. Checkpointing then flushes the WAL to the disk DB.
The `vfsFileMethods` for the DB file would then generallly be the vanilla sqlite3 `vfsFileMethods`.

A couple of issues:
- Checkpointing to disk happens in the main loop, this can quickly take a couple of hundred ms, blocking the eventloop. I don't really know how this can be avoided. https://github.com/canonical/dqlite/blob/35b5418af73d5112c156d05eec2ad85268707cb3/src/fsm.c#L274
- Snapshotting the database means copying the on-disk database, instead of passing pointers to raft. For this we would probably need to add the async snapshot behavior to raft. https://github.com/canonical/raft/pull/268 . 
- Snapshotting aka copying a large database every `1024` transactions is also possibly very heavy on the storage. We can lower the frequency, but this will increase memory footprint of raft log. 
- Because the database can be large, we can't just load the entire snapshot in memory when sending a raft `installSnapshot RPC`. We would need to add a way to send the snapshot in chunks, adding to overall complexity. (e.g. we can't just delete snapshots that are still being sent, etc.)

All-in-all I'm not feeling too confident right now if this is practically feasible, do you want to share some insights on this @freeekanayaka ?

marco6:
This PR removes disk mode as:
 - we do not intend to put effort in this feature, leaving it half-backed
 - it makes it more difficult to maintain this library
 - even after the execution has been moved in the thread pool there is still some operations which will run on the main loop in a blocking way and which are very difficult not to:
   - the checkpoint;
   - the DUMP operation;
   - the snapshot (opening, mmapping);
   - preparing statement.
 - it was a bad idea to block the main loop to begin with.

I hope I purged every bit of it from the repo.

This likely needs a PR for go-dqlite before merging, so that it can detect if disk more is available or not.

Closes #368.


marco6:
Marking as draft as it's still undecided **and** depends on `go-dqlite` to be able to cope with missing disk mode. This is addressed in canonical/go-dqlite#389.

PR #802, opened 2025-08-05
feat: improve cancellation behaviour

marco6:


src/leader.c
```diff
@@ -95,6 +100,9 @@ void leader__close(struct leader *leader, leader_close_cb close_cb)
 		PRE(IN(db->active_leader, NULL, req->leader));
 		db->active_leader = req->leader;
 		return exec_tick(req);
+	} else {
+		/* Abort all queries as soon as possible. */
```
letFunny, line 104: This is only going to happen when SQL queries have multiple statements, right? (e.g. BEGIN IMMEDIATE; SELECT ...; INSERT; COMMIT)

marco6: No. It happens at the SQLite VM level, so it will kill queries in the middle of running.

marco6: See https://github.com/sqlite/sqlite/blob/master/src/vdbe.c#L1098

letFunny: I meant the pending == 0, I remember we are using the progress handler precisely because it interrupts queries more reliably than sqlite3_interrupt

PR #803, opened 2025-08-05
fix: defer apply when busy

marco6:


test/unit/test_replication.c
```diff
@@ -479,3 +484,36 @@ TEST(replication, checkpoint, setUp, tearDown, 0, NULL)
 
 	return MUNIT_OK;
 }
+
+TEST(replication, leaderToFollowerBusy, setUp, tearDown, 0, NULL)
+{
+	struct fixture *f = data;
+
+	CLUSTER_ELECT(0);
+	PREPARE(0, "BEGIN IMMEDIATE");
```
letFunny, line 494: I think a comment here would be nice to draw attention to the fact that this is really checking what happens when leader becomes follower _while_ holding the write lock, and that it can proceed.

test/unit/test_replication.c
```diff
@@ -479,3 +484,36 @@ TEST(replication, checkpoint, setUp, tearDown, 0, NULL)
...
+	CLUSTER_DEPOSE;
+	SETUP_LEADER(1);
+	CLUSTER_ELECT(1);
+	PREPARE(1, "CREATE TABLE test(i ANY)")
+	fixture_exec(f, 1);
+	FINALIZE;
+	TEAR_DOWN_LEADER(1);
+	CLUSTER_COMMITTED(4);
+	raft_fixture_step_until_applied(&f->cluster, 1, 4, 1000);
+	raft_fixture_step_until_applied(&f->cluster, 2, 4, 1000);
```
letFunny, line 508: It seems to me like this test is testing than when leader has write lock and becomes follower it can apply pages it gets. Should we also test separately the case where:
1. Leader holds write lock with a query that makes the WAL dirty
2. Leader loses leadership.
3. New leader elected, no pages created.
4. Old leader (1) becomes leader again and can proceed with new queries AND the WAL is not dirty with the insert that didn't go through.

src/vfs.c
```diff
@@ -2596,7 +2597,7 @@ static void vfsInvalidateWalIndexHeader(struct vfsDatabase *d)
 {
 	struct vfsShm *shm = &d->shm;
 
-	PRE(shm->lock[VFS__WAL_WRITE_LOCK] >= 0);
+	PRE(shm->lock[VFS__WAL_WRITE_LOCK] < 0);
```
letFunny, line 2603: How did you find this? I expect the changes to have nothing to do with this.

test/unit/test_replication.c
```diff
@@ -421,21 +430,17 @@ TEST(replication, exec, setUp, tearDown, 0, NULL)
...
 	fixture_exec(f, 0);
 	CLUSTER_STEP;
 	munit_assert_true(f->invoked);
 	munit_assert_int(f->status, ==, RAFT_OK);
-	f->invoked = false;
 	FINALIZE;
 
 	PREPARE(0, "COMMIT");
 	fixture_exec(f, 0);
-	munit_assert_false(f->invoked);
```
letFunny, line 438: Why? I understand the rest of the changes but this one was actually verifying that it wasn't called.

src/raft/fixture.c
```diff
@@ -1797,6 +1797,36 @@ bool raft_fixture_step_until_applied(struct raft_fixture *f,
...
+
+	if (apply->i < f->n) {
+		raft = raft_fixture_get(f, apply->i);
+		return raft_commit_index(raft) >= apply->index;
+	}
+
+	for (i = 0; i < f->n; i++) {
+		raft = raft_fixture_get(f, i);
+		if (raft_commit_index(raft) >= apply->index) {
+			n++;
```
letFunny, line 1815: Can't you return false when `raft_commit_index(raft) < apply->index` and change the last line to `return true`? Isn't that simpler?

src/vfs.c
```diff
@@ -1196,6 +1196,7 @@ struct vfsMainFile {
 	sqlite3_file base;            /* Base class. Must be first. */
 	struct vfs *vfs;              /* Pointer to volatile VFS data. */
 	struct vfsDatabase *database; /* Underlying database content. */
+	bool polled;                  /* Wether this WAL was polled or not. */
```
letFunny: `wether` -> `whether`

Most importantly, what does exactly polled mean? I thought it meant that there are some pages that were polled but not applied but I see from usage it is different than that and I don't fully understand what it entails, can you elaborate?

src/vfs.c
```diff
@@ -2650,17 +2661,18 @@ int VfsApply(sqlite3 *conn, const struct vfsTransaction *transaction)
 	 * originated the transaction (this can happen for example when applying
 	 * a Raft barrier and replaying the Raft log in order to serve a request
 	 * of a newly connected client). */
-	if (f->database->shm.lock[VFS__WAL_WRITE_LOCK] < 0) {
-		/* Make sure that this is the connection holding the lock. */
+	if (f->polled) {
 		PRE(f->exclMask & (1 << VFS__WAL_WRITE_LOCK));
```
letFunny, line 2669: `f->exclMask & (1 << VFS__WAL_WRITE_LOCK)` implies `f->database->shm.lock[VFS__WAL_WRITE_LOCK] < 0`, right?

src/vfs.c
```diff
@@ -2650,17 +2661,18 @@ int VfsApply(sqlite3 *conn, const struct vfsTransaction *transaction)
 	 * originated the transaction (this can happen for example when applying
 	 * a Raft barrier and replaying the Raft log in order to serve a request
 	 * of a newly connected client). */
-	if (f->database->shm.lock[VFS__WAL_WRITE_LOCK] < 0) {
-		/* Make sure that this is the connection holding the lock. */
+	if (f->polled) {
```
letFunny, line 2668: The comment above needs to be updated.

src/vfs.c
```diff
@@ -2626,6 +2627,16 @@ int VfsApply(sqlite3 *conn, const struct vfsTransaction *transaction)
...
 	tracef("vfs apply on %s %u pages", f->database->name, transaction->n_pages);
 
+	
+	if (!f->polled) {
+		/* If this connection wasn't the one originating the transaction and there is 
+		* another on-going write transaction it is not possible to change the underlying 
+		* WAL object as that would cause corruption of the database. */
+		rv = vfsShmLock(&f->database->shm, VFS__WAL_WRITE_LOCK, 1, true);
+		if (rv != SQLITE_OK) {
+			return rv;
```
letFunny, line 2640: Okay so this will return SQLITE_BUSY if someone is holding the write lock, this is the magic

src/vfs.c
```diff
@@ -2626,6 +2627,16 @@ int VfsApply(sqlite3 *conn, const struct vfsTransaction *transaction)
 	f = (struct vfsMainFile*)file;
 	tracef("vfs apply on %s %u pages", f->database->name, transaction->n_pages);
 
+	
+	if (!f->polled) {
```
letFunny, line 2634: What is this check for? I assume that we are using it to signal that there are some frames that have to be applied. What I fail to see is why when `f->polled == false` & no write lock we continue executing this function.

letFunny (commented):
I have reviewed it and read it multiple times but I am having a hard time getting my head around what you wanted to model using `polled`, I see the implementation but I don't get the full idea. Let's discuss in one of the threads in my review that is about precisely that.

src/fsm.c
```diff
@@ -112,6 +112,7 @@ static int apply_frames(struct fsm *f, const struct command_frames *c)
 	rv = VfsApply(conn, &transaction);
 	if (rv != 0) {
 		tracef("VfsApply failed %d", rv);
+		rv = (rv == SQLITE_BUSY) ? RAFT_BUSY : RAFT_IOERR;
```
just-now: nit:
```
rv = rv == SQLITE_BUSY ? RAFT_BUSY : RAFT_IOERR;
```

src/raft/fixture.c
```diff
@@ -1797,6 +1797,36 @@ bool raft_fixture_step_until_applied(struct raft_fixture *f,
 	return raft_fixture_step_until(f, hasAppliedIndex, &apply, max_msecs);
 }
 
+static bool hasCommittedIndex(struct raft_fixture *f, void *arg)
+{
+	struct step_apply *apply = (struct step_apply *)arg;
```
just-now: nit: `const struct`

src/raft/replication.c
```diff
@@ -1252,6 +1252,8 @@ int replicationAppend(struct raft *r,
 		    !replicationInstallSnapshotBusy(r)) {
 			r->commit_index =
 			    min(args->leader_commit, r->last_stored);
+		}
+		if (r->last_applied < r->commit_index) {
```
just-now, line 1257: Could you put a `tracef(r->last_applied, r->commit_index)` and more paranoia here, like `PRE(last_applied <= commit_index)`? 
Think about invariants that could save us from troubles, as the main algorithm and its "safety properties" are being changed here.

Note: It seems to be a logical change to call `replicationApply()` on all not applied indexes. Still make sure that you know there are a few places in the code base where `r->last_applied` is updated later than it was actually changed.

src/raft/replication.c
```diff
@@ -1934,6 +1936,8 @@ int replicationApply(struct raft *r)
 
 	if (shouldTakeSnapshot(r)) {
 		rv = takeSnapshot(r);
+	} else if (rv == RAFT_BUSY) {
+		rv = RAFT_OK;
```
just-now, line 1942: This is not covered with tests neither in `raft-core-integration-test` nor in  `raft-uv-integration-test`.
Could you explain in a comment around this code where `RAFT_BUSY` comes from and why shall it be  `RAFT_OK`.

For Alberto and Ed: This is a good place to demonstrate why debugger is needed while reading patches.

letFunny: > For Alberto and Ed: This is a good place to demonstrate why debugger is needed while reading patches.

Or coverage :)

src/vfs.c
letFunny:
```
Whether this connection was polled or not. Marking the connection in the leader allows us to ensure that the connection that is polled is the one applied or aborted.
```

test/unit/test_replication.c
marco6: True, but maybe it is ok to postpone this.

src/vfs.c
marco6: The logic now takes a write lock before calling this routine. That check would be always failing then.

test/unit/test_replication.c
marco6: `fixture_exec` now waits, so it will always be true.

src/raft/fixture.c
marco6: This was a copy-paste of `hasAppliedIndex` just changing `raft_last_applied` with `raft_commit_index`. I think it's best to keep them with the same structure.

src/raft/replication.c
marco6: I added the check here and a trace in `replicationApply`.

src/raft/replication.c
marco6: It should be covered however by the new test in `test_replication.c`. Can this be enough?

src/vfs.c
marco6: The second bit happens on a follower which clearly will never poll connections.

letFunny (approved):
Discussed offline: because this is blocking some pipelines, the logic is sound, and we are going to be on vacation soon, we can go ahead and merge it. When we are back, @marco6 and me will think of how to write comments and better naming to make the code more readable because the way this is written it is very hard to read with no context.

src/vfs.c
marco6: Updated.

src/vfs.c
marco6: Done.

src/raft/replication.c
just-now: Yup. It's sufficient.

PR #804, opened 2025-08-06
perf: streaming compression/decompression

marco6:
This PR changes the way snapshots are compressed so that the amount of memory used is fixed and does not grow with the snapshot, when compression is on.

This should remove the memory spikes during snapshot and should not affect CPU performance in any significant way.


src/raft/compress.c
```diff
@@ -1,277 +0,0 @@
...
-	(void)compressed;
-	ErrMsgPrintf(errmsg, "LZ4 not available");
-	return RAFT_INVALID;
-#else
-	assert(bufs != NULL);
-	assert(n_bufs > 0);
-	assert(compressed != NULL);
-	assert(errmsg != NULL);
-
-	int rv = RAFT_IOERR;
```
letFunny, line 34: Can you check either with a test or with a manual check that the new code can read the old snapshots? I know they both are lz4, but let's just double-check.

src/raft/uv_fs.c
```diff
@@ -655,6 +823,145 @@ int UvFsReadFile(const char *dir,
...
+	uv_file fd;
+	rv = uvFsOpenFile(dir, filename, O_RDONLY, 0, &fd, errmsg);
+	if (rv != 0) {
+		return rv;
+	}
+
+	LZ4F_decompressionContext_t ctx;
+	size_t lzrv = LZ4F_createDecompressionContext(&ctx, LZ4F_VERSION);
+	if (LZ4F_isError(lzrv)) {
+		// FIXME errmsg
```
letFunny: Left-over?

src/raft/uv_fs.c
```diff
@@ -655,6 +823,145 @@ int UvFsReadFile(const char *dir,
...
+		ErrMsgPrintf(errmsg, "LZ4F_getFrameInfo %s",
+			     LZ4F_getErrorName(lzrv));
+		rv = RAFT_IOERR;
+		goto done;
+	}
+
+	if (frameInfo.contentSize == 0) {
+		rv = RAFT_OK;
+		goto done; // Not really an error here...
+	}
```
letFunny, line 914: Is it? Should we expect to produce blocks that have 0 content even if the format allows it?

src/raft/uv_fs.c
```diff
@@ -655,6 +823,145 @@ int UvFsReadFile(const char *dir,
...
+	if (output_buffer == NULL) {
+		rv = RAFT_NOMEM;
+		goto done;
+	}
+
+	size_t output_offset = 0;
+	while (output_offset < output_size) {
+		if (input_offset == input_size) {
+			/* try to read some */
+			ssize_t read_rv = read(fd, input_buffer, input_buffer_size);
```
letFunny, line 927: Clever, reusing the buffer and using offset and size to work with the remaining input but also when reading more data.

I wonder if it would be more clear to read into `input_buffer + input_offset` and updating the `input_offset` over the whole buffer instead.

src/raft/uv_fs.c
```diff
@@ -655,6 +823,145 @@ int UvFsReadFile(const char *dir,
...
+				ErrMsgPrintf(errmsg, "short read: %zu bytes instead of %zu",
+			     	output_offset, output_size);
+				rv = RAFT_IOERR;
+				goto err_after_output_alloc;
+			}
+			input_size = (size_t)read_rv;
+			input_offset = 0;
+		}
+
+		size_t output_written = output_size - output_offset;
```
letFunny: The naming doesn't help here. This is per the documentation the size of the buffer when calling into `LZ4F_decompress` but then it is the number of bytes that were "written" into the buffer. This seems minor but the choice of having a variable of the wrong name used as input for something different will perplex anyone reading the code that does not look at the LZ4 documentation. What about calling it `output_size` and then doing `size_t output_written = *output_size`? This applies to all parameters in these functions.

If you consider this unimportant please tell me, I don't have a strong opinion. Also, if there is a better solution also please tell me.

src/raft/uv_fs.c
```diff
@@ -500,6 +504,170 @@ int UvFsMakeFile(const char *dir,
...
+		.base = data,
+		.len = length,
+	};
+	int rv = UvOsWrite(fd, &buffer, 1, offset);
+	if (rv == (int)length) {
+		return RAFT_OK;
+	} else if (rv < 0) {
+		UvOsErrMsg(errmsg, "write", rv);
+	} else {
+		ErrMsgPrintf(errmsg, "short write: %d only bytes written", rv);
```
letFunny, line 519: Is a short write an error or should we just retry?

src/raft/uv_fs.c
```diff
@@ -500,6 +504,170 @@ int UvFsMakeFile(const char *dir,
...
+
+	rv = uvOsWriteOne(fd, output_buffer, output_len, -1, errmsg);
+	if (rv != RAFT_OK) {
+		goto err_after_ctx_alloc;
+	}
+
+	for (unsigned i = 0; i < n_bufs; i++) {
+		output_len = LZ4F_compressUpdate(ctx, output_buffer, output_cap, bufs[i].base, bufs[i].len, NULL);
+		if (output_len == 0) {
+			continue;
```
letFunny, line 611: Can you add a comment here saying `// The output is buffered internally in LZ4F.` or something similar?

src/raft/uv_fs.c
```diff
@@ -500,6 +504,170 @@ int UvFsMakeFile(const char *dir,
...
+
+	size_t output_len = LZ4F_compressBegin(ctx, output_buffer, output_cap, &lz4_pref);
+	if (LZ4F_isError(output_len)) {
+		ErrMsgPrintf(errmsg, "LZ4F_compressBegin %s",
+			     LZ4F_getErrorName(output_len));
+		rv = RAFT_IOERR;
+		goto err_after_ctx_alloc;
+	}
+
+	rv = uvOsWriteOne(fd, output_buffer, output_len, -1, errmsg);
```
letFunny, line 599: This is writing the header, correct? Can we add a short comment `// Writing header.`?

src/raft/uv_fs.c
```diff
@@ -500,6 +504,170 @@ int UvFsMakeFile(const char *dir,
...
+
+	output_len = LZ4F_compressEnd(ctx, output_buffer, output_cap, NULL);
+	if (LZ4F_isError(output_len)) {
+		ErrMsgPrintf(errmsg, "LZ4F_compressEnd %s",
+					LZ4F_getErrorName(output_len));
+		rv = RAFT_IOERR;
+		goto err_after_ctx_alloc;
+	} else if (output_len > 0) {
+		rv = uvOsWriteOne(fd, output_buffer, output_len, -1, errmsg);
+
```
letFunny: Nit: newline is unnecessary.

src/raft/uv_fs.c
```diff
@@ -500,6 +504,170 @@ int UvFsMakeFile(const char *dir,
...
+err_after_ctx_alloc:
+	LZ4F_freeCompressionContext(ctx);
+err_after_buf_alloc:
+	raft_free(output_buffer);
+err_after_open:
+	UvOsClose(fd);
+	if (rv != RAFT_OK) {
+		/* Try to unlink the temp file. */
+		UvOsUnlink(temp_file.path);
```
letFunny, line 670: This is useless after the rename and might hit another file. But even if that is highly highly unlikely I just wanted to point it out.

src/raft/uv_fs.h
```diff
@@ -51,6 +51,13 @@ int UvFsMakeFile(const char *dir,
 		 unsigned n_bufs,
 		 char *errmsg);
 
+/* Create a file and write the given content into it. */
+int UvFsMakeCompressedFile(const char *dir,
```
letFunny, line 55: weird that you need the directory here. The usual way would be to get the path and get the dir by calling the equivalent of `path.Dir`, it's a pity that C lacks a good standard library.

test/raft/unit/test_compress.c
```diff
@@ -1,319 +1,308 @@
...
+		    f->dir, file, &decompressed, f->errmsg);                 \
+		munit_assert_int(decompress_rv, ==, RAFT_OK);                \
+		int compare_rv = compare_bufs(bufs, len, &decompressed);     \
+		munit_assert_int(compare_rv, ==, RAFT_OK);                   \
+		raft_free(decompressed.base);                                \
+	} while (0)
+
+#define COMPRESS(file, bufs, ...)                \
+	COMPRESS_IMPL(file, bufs, ##__VA_ARGS__, \
+		      (sizeof(bufs) / sizeof(bufs[0])))
```
letFunny: Clever, I always forget about these tricks until I see them.

I think it is better to do a check manually and update `len` because it is always one to be 1 when null or len when valid.

test/raft/unit/test_compress.c
```diff
@@ -1,319 +1,308 @@
...
+		struct raft_buffer decompressed = {};                        \
+		int decompress_rv = UvFsReadCompressedFile(                  \
+		    f->dir, file, &decompressed, f->errmsg);                 \
+		munit_assert_int(decompress_rv, ==, RAFT_OK);                \
+		int compare_rv = compare_bufs(bufs, len, &decompressed);     \
+		munit_assert_int(compare_rv, ==, RAFT_OK);                   \
+		raft_free(decompressed.base);                                \
+	} while (0)
+
+#define COMPRESS(file, bufs, ...)                \
```
letFunny: Can we rename this to `COMPRESSION_TEST` or `TEST_COMPRESS` or `COMPRESS_TEST`?

test/raft/unit/test_compress.c
```diff
@@ -1,319 +1,308 @@
...
+
+    char path[PATH_MAX] = {};
+	int rv = UvOsJoin(f->dir, "test", path);
+    munit_assert_int(rv, ==, 0);
+
+    uv_stat_t sb = {};
+    rv = UvOsStat(path, &sb);
+    munit_assert_int(rv, ==, 0);
+    munit_assert_ulong(sb.st_size, >, 0);
+    munit_assert_ulong(sb.st_size, <, len);
```
letFunny: Funny that this always works. I would assume that if you source of randomness was perfect and the entropy was the highest then compression would not reduce the length but it **will** add a header, meaning the resulting file will be longer than the input.

letFunny (commented):
This is super-readable Marco, thanks for working on it. As usual, most of my comments are to tone down some of the cleverness and to document it :P. Nothing major, consider it approved.

src/raft/uv_fs.c
marco6: I think the original code would error in case of `contentSize == 0`. I also think that this cannot happen.

src/raft/uv_fs.c
marco6: I think the reasoning here is the following: we are reading, as such there are 2 independent time-consuming tasks:
 - reading the file (consumes I/O)
 - decompressing (consumes CPU)
 
So, this loop will not issue more than one read before starting the decompress and to wait as much as it can before issuing a new one.

Consider that *on linux* this is important as the default behaviour when opening a file for read, is to have a "readahead" window that is filled as soon as the cursor of the file moves. So, it is better to always finish the input you have instead of trying to fill a later bit. In this sense, if we processed the whole buffer it seems to me natural that we reset its offset to 0 to process a new chunk.

Writing this also made me think that we can probably gain some more performance by using [`posix_fadvice`](https://man7.org/linux/man-pages/man2/posix_fadvise.2.html) which I just added (it will double the size of the readahead window).

src/raft/uv_fs.c
marco6: I think that this was to be in-line with what you can find in [other methods](https://github.com/canonical/dqlite/blob/master/src/raft/uv_fs.c#L413).

I do think that this it would be better to retry, but maybe it's worth doing that in a separate PR and fix it everywhere? WDYT?

src/raft/uv_fs.h
marco6: Agreed. But once again I wanted to be on-par with the existing API.

test/raft/unit/test_compress.c
marco6: Your assumption is indeed correct: it is a requirement for a bijection (a lossless compression function) to generate bigger files in some cases as otherwise you would be mapping a bigger space in a smaller one and that is not possible without loosing the bijection property (= becoming lossy).

What is wrong is how you read this tests: it is not using a randomness source, but [filling the buffer with the very same byte (`0xAC`)](https://github.com/canonical/dqlite/pull/804/files/d9443957467dac98caa5a6ecefc2146d22ce4e3e#diff-5ba55d7cf20af81e2b5817c08082a2ba34299c963021ac28d3fe19368ec0f26fR193).

test/raft/unit/test_compress.c
marco6: I went for `TEST_COMPRESS`.

src/raft/uv_fs.c
marco6: Done.

src/raft/uv_fs.c
marco6: The only way to reach this point after the rename is if the `fsync` on the directory failed. In that case isn't it fair to assume that the rename failed?

src/raft/uv_fs.c
marco6: I renamed `output_size` to `output_buffer_size` and `output_written` to `output_size`.

src/raft/compress.c
marco6: It would be difficult with a test as I am deleting the original code. I tried on my PC with the following code:

```C
#include <stddef.h>
#include <stdio.h>

struct raft_buffer {
  void *base; /* Pointer to the buffer data. */
  size_t len; /* Length of the buffer. */
};

int UvFsReadCompressedFile(const char *dir, const char *filename,
                           struct raft_buffer *buf, char *errmsg);

int main() {
  char errmsg[4096] = {};
  struct raft_buffer buffer = {};
  int rv = UvFsReadCompressedFile(
      "<path to a dqlite folder from a previous release>",
      "<snapshot name>", &buffer, errmsg);
  if (rv != 0) {
    printf("error (%d): %s\n", rv, errmsg);
    return -1;
  } else {
    printf("success! Uncompressed size: %lu\n", buffer.len);
  }
}
```

Compiled with:

```
gcc test-snapshot.c ../dqlite/build/.libs/libdqlite.a -luv -llz4
```

And it worked.

test/raft/unit/test_compress.c
marco6: I removed the trick and used always a manual parameter.

src/raft/compress.c
letFunny: That is what I was referring to, thank you! I am still a bit hesitant to say this works when your test doesn't assert the uncompressed content is the same as this does not rule out bugs when the decompression does not error out but the result is bogus.

src/raft/uv_fs.c
letFunny: Should we then error? That is an assumption that we have, `assert` would be perfect here.

src/raft/uv_fs.c
letFunny: They are equivalent, aren't they? It doesn't matter if you use the start of the buffer to length `n` than to start at offset `o` until `o + n` if that makes sense. Assuming the behavior is the same, do you think it is clearer if you read over the whole buffer?

marco6: In my mind it is cleaner this way as you only care about the current buffer (easier to reason as the scope is smaller) rather than the entire file.

src/raft/uv_fs.c
letFunny: Sure, let's not do it in this PR.

src/raft/uv_fs.c
marco6: I think not as that is an upper layer requirement, not one for this function IMHO.

src/raft/uv_fs.c
letFunny: I guess it does not matter that much

test/raft/unit/test_compress.c
letFunny: hahahahah right, I might have mixed some of the tests as I was reading it.

src/raft/compress.c
marco6: Yes but this would be testing the lz4 library wouldn't it? If the decompression does not signal an error and correctly finalizes it also mean checksum match... Or lz4 has a bug.

letFunny: Yes, and testing the little bit of integration glue. Maybe I am being overparanoid but I wanted to rule out cases like our previous implementation having a bug that is not handled in the same way by the new one. The result is the same, not being able to read old snapshots.

I do think we need to test our dependencies for backwards compatibility especially in a LTS.

src/raft/uv_fs.c
letFunny: I don't think there are use-cases where the content being 0 is valid anyway, but you are right this is mixing concerns. Consider this resolved.

src/raft/uv_fs.c
letFunny: Okay, if it comes down to style, I don't mind either way.

src/raft/compress.c
marco6: As I said on MM I think this is too much. But still, I changed the initial code to:

```C
#include <stddef.h>
#include <stdio.h>

struct raft_buffer {
  void *base; /* Pointer to the buffer data. */
  size_t len; /* Length of the buffer. */
};

int UvFsReadCompressedFile(const char *dir, const char *filename,
                           struct raft_buffer *buf, char *errmsg);
int UvFsMakeFile(const char *dir, const char *filename,
                 struct raft_buffer *bufs, unsigned n_bufs, char *errmsg);

int main() {
  char errmsg[4096] = {};
  struct raft_buffer buffer = {};
  int rv = UvFsReadCompressedFile(
      "<path to a dqlite folder from a previous release>",
      "<snapshot name>", &buffer, errmsg);
  if (rv != 0) {
    printf("error (%d): %s\n", rv, errmsg);
    return -1;
  }
  rv = UvFsMakeFile(".", "snapshot-uncompressed", &buffer, 1, errmsg);
  if (rv != 0) {
    printf("error (%d): %s\n", rv, errmsg);
    return -1;
  }
  return 0;
}
```

Then checked with diff if it matched the output of like:

```bash
lz4 -f <path> check-lz4
diff snapshot-uncompressed check-lz4
```

And they match.

I also wrote this code:

```C
#include <stddef.h>
#include <stdio.h>

struct raft_buffer {
  void *base; /* Pointer to the buffer data. */
  size_t len; /* Length of the buffer. */
};

int UvFsReadFile(const char *dir, const char *filename, struct raft_buffer *buf,
                 char *errmsg);
int Decompress(struct raft_buffer buf, struct raft_buffer *decompressed,
               char *errmsg);
int UvFsMakeFile(const char *dir, const char *filename,
                 struct raft_buffer *bufs, unsigned n_bufs, char *errmsg);

int main() {
  char errmsg[4096] = {};
  struct raft_buffer buffer, decompressed = {};
  int rv = UvFsReadFile("/home/marco/workspace/dqlite-utils/dqlite-test-rdUBVF",
                        "snapshot-1-19456-25054058", &buffer, errmsg);
  if (rv != 0) {
    printf("error (%d): %s\n", rv, errmsg);
    return -1;
  }

  rv = Decompress(buffer, &decompressed, errmsg);
  if (rv != 0) {
    printf("error (%d): %s\n", rv, errmsg);
    return -1;
  }

  rv = UvFsMakeFile(".", "old-snapshot-uncompressed", &decompressed, 1, errmsg);
  if (rv != 0) {
    printf("error (%d): %s\n", rv, errmsg);
    return -1;
  }
  return 0;
}
```

And linked with the old dqlite code. Then:

```bash
diff snapshot-uncompressed old-snapshot-uncompressed
```

They still match.

letFunny: This is exactly what I was referring to. We could do much more to make it safer at the expense of horrible performance (as we discussed) or a lot more complexity. For me the manual tests you did strike a good balance, thanks!

letFunny (approved):
Thanks Marco, LGTM!

PR #805, opened 2025-08-07
chore: embed human-readable version string in binary

marco6:
This PR adds an exported symbol to the binary in the format `<version>@<git_hash>`. If the hash is not available, `unknown` will be used.

This helps in debugging so that we know exactly which version is in production.

In case of static linking with whole program and symbols stripped it would still be possible for the *embedding* application to expose this information by printing that to the log, as the variable is public in the dqlite header.


configure.ac
```diff
@@ -121,5 +121,18 @@ CC_CHECK_FLAGS_APPEND([AM_CFLAGS],[CFLAGS],[ \
 #  -Wpedantic \
 AC_SUBST(AM_CFLAGS)
 
+
+AC_MSG_CHECKING([for Git version])
+
+GIT_BUILD_TAG=""
+if (git --version >/dev/null 2>&1) && (git rev-parse --is-inside-work-tree >/dev/null 2>&1); then
```
letFunny: Nit: Checking `git version` is redundant given the second command will fail if git is not present.

just-now (approved):
Awesome!

configure.ac
marco6: removed

PR #811, opened 2025-08-21
 refactor: improve snapshot interface 

marco6:
This PR improves the interface around snapshotting so that the locking is explicit.


marco6:
This PR was part of #809 and depends on #801.

src/vfs.h
```diff
@@ -35,28 +35,22 @@ int VfsAbort(sqlite3 *conn);
...
-		       const char *filename,
-		       struct dqlite_buffer bufs[],
-		       uint32_t n);
+struct vfsSnapshot {
+	struct vfsFile main;
+	struct vfsFile wal;
+};
+
+int VfsAcquireSnapshot(sqlite3 *conn, struct vfsSnapshot *snapshot);
+int VfsReleaseSnapshot(sqlite3 *conn, struct vfsSnapshot *snapshot);
```
letFunny: These need documentation.

src/registry.h
```diff
@@ -13,14 +13,25 @@ struct registry
...
 
 void registry__init(struct registry *r, struct config *config);
 void registry__close(struct registry *r);
 
 /**
- * Get the db with the given filename. If no one is registered, create one.
+ * Returns the db with the given filename. If the db does not exists, it is created.
  */
-int registry__db_get(struct registry *r, const char *filename, struct db **db);
+int registry__create(struct registry *r, const char *filename, struct db **db);
```
letFunny: Let's try to improve the name here but it is puzzling to see `registry__create` for something that already exists. What about something lengthier `registry__get_or_create` or how about having one function called `registry__get` with a parameter that when set to true creates the entry if it doesn't exist.

test/unit/test_gateway.c
```diff
@@ -86,6 +94,38 @@ struct connection {
 	}                                                  \
 	TEAR_DOWN_CLUSTER;
 
+#define INTEGRITY_CHECK(file)                                                 \
```
letFunny, line 97: Nice to see this! But why is it a macro? It looks like a regular function whose only inputs are the file, there is no variables being created or taken inside.

test/unit/test_gateway.c
```diff
@@ -86,6 +94,38 @@ struct connection {
...
+		int fd = open(path, O_CREAT | O_RDWR, S_IRUSR | S_IWUSR);     \
+		munit_assert_int(fd, >, 0);                                   \
+		ssize_t rv =                                                  \
+		    write(fd, (file).content.base, (file).content.len);       \
+		munit_assert_int(rv, ==, (file).content.len);                 \
+		sqlite3 *conn;                                                \
+		int sqlite_err =                                              \
+		    sqlite3_open_v2(path, &conn, SQLITE_OPEN_READONLY, NULL); \
+		munit_assert_int(sqlite_err, ==, SQLITE_OK);                  \
+		sqlite_err = sqlite3_exec(conn, "PRAGMA integrity_check",     \
```
letFunny: Nit: `rc` or `rv` not necessarily error.

test/unit/test_gateway.c
```diff
@@ -3112,12 +3153,15 @@ TEST_SETUP(dump)
 	struct request_dump_fixture *f = munit_malloc(sizeof *f);
 	SETUP;
 	CLUSTER_ELECT(0);
+	OPEN;
```
letFunny: I remember in the past we had a bug where dump was being called without an open connection to the database, are we still testing that case?

test/unit/test_gateway.c
```diff
@@ -3137,14 +3182,16 @@ TEST_CASE(dump, empty, NULL)
 	struct file main = {};
 	DECODE_FILE(&main);
 	munit_assert_string_equal(main.name, "test");
-	munit_assert_int(main.content.len, ==, 0);
+	munit_assert_int(main.content.len, ==, 512);
```
letFunny, line 3185: How come the length was 0 in the previous test?

test/unit/test_gateway.c
```diff
@@ -3157,21 +3204,8 @@ TEST_CASE(dump, not_existent, NULL)
...
-	munit_assert_string_equal(main.name, "foo");
-	munit_assert_int(main.content.len, ==, 0);
-
-	struct file wal = {};
-	DECODE_FILE(&wal);
-	munit_assert_string_equal(wal.name, "foo-wal");
-	munit_assert_int(wal.content.len, ==, 0);
-
-	munit_assert_int(f->cursor->cap, ==, 0);
+	ASSERT_CALLBACK(DQLITE_NOTFOUND, FAILURE);
```
letFunny, line 3207: Indeed much better design decision.

src/fsm.c
```diff
@@ -217,293 +191,277 @@ static int encodeSnapshotHeader(unsigned n, struct raft_buffer *buf)
...
 
-	if (db->leaders > 0) {
-		return RAFT_BUSY;
+	struct db *db;
+	rv = registry__create(f->registry, header.filename, &db);
+	if (rv != DQLITE_OK) {
+		if (rv == DQLITE_NOMEM) {
+			return RAFT_NOMEM;
+		}
+		return RAFT_ERROR;
```
letFunny: Nit: return rv == DQLITE_NOMEM ? RAFT_NOMEM : RAFT_ERROR.

src/fsm.c
```diff
@@ -217,293 +191,277 @@ static int encodeSnapshotHeader(unsigned n, struct raft_buffer *buf)
...
+		pages[i] = (void*)(cursor->p + i * page_size);
+	}
+
+	const struct vfsSnapshot snapshot = {
+		.main = {
+			.page_count = header.main_size / page_size,
+			.page_size = page_size,
+			.pages = pages,
+		},
+		// FIXME this breaks compatibility with very old dqlite 1.17.x (the one before shallow snapshot got in)
```
letFunny, line 233: Do you mean that we should really remove the `wal` inside the data structure because it is always going to be empty?

src/fsm.c
```diff
@@ -217,293 +191,277 @@ static int encodeSnapshotHeader(unsigned n, struct raft_buffer *buf)
...
+			.page_count = 0,
+			.page_size = page_size,
+			.pages = NULL,
+		}
+	};
 
-	rv = VfsDatabaseNumPages(db->vfs, db->filename, true, &n);
-	assert(rv == 0);
-	return n;
+	/* Due to the check above, this cast is safe. */
```
letFunny: This comment is leftover.

src/fsm.c
```diff
@@ -217,293 +191,277 @@ static int encodeSnapshotHeader(unsigned n, struct raft_buffer *buf)
...
+	rv = VfsRestore(conn, &snapshot);
```
letFunny: So inside a function that _decodes_ the snapshot we are also using it to restore the database. This is not about the new code because it was present in the previous one, but nevertheless it's not a good choice. In my opinion, `fsm__restore` should:
```c
		rv = decodeSnapshot(&snapshot, &cursor);
		if (rv != 0) {
			...
		}
		rv = restoreSnapshot(f, &snapshot);
		if (rv != 0) {
			...
		}
```
Notice that this separation is natural because the connection to the db and the fsm is only used when restoring which is not part of decoding.

src/fsm.c
```diff
@@ -217,293 +191,277 @@ static int encodeSnapshotHeader(unsigned n, struct raft_buffer *buf)
...
-	rv = db->vfs->xAccess(db->vfs, header.filename, 0, &exists);
-	assert(rv == 0);
-
-	if (!exists) {
-		sqlite3 *conn;
-		rv = db__open(db, &conn);
-		if (rv != 0) {
-			return rv;
+	sqlite3 *conn;
+	rv = db__open(db, &conn);
```
letFunny: Can we move this down to just before using it to restore the snapshot, see also my other comment about breaking up the function.

src/fsm.c
```diff
@@ -217,293 +191,277 @@ static int encodeSnapshotHeader(unsigned n, struct raft_buffer *buf)
...
+	if (rv != SQLITE_OK) {
+		return RAFT_ERROR;
+	}
 
-	QUEUE_FOREACH(head, &f->registry->dbs)
-	{
-		n += 1; /* database header */
-		db = QUEUE_DATA(head, struct db, queue);
-		n += dbNumPages(db); /* 1 buffer per page (zero copy) */
+	rv = VfsCheckpoint(snapshot->conn, 0);
```
letFunny, line 312: This needs to be recorded in the external documentation.

src/fsm.c
```diff
@@ -217,293 +191,277 @@ static int encodeSnapshotHeader(unsigned n, struct raft_buffer *buf)
...
+	/* I think this can be an assert. */
+	if (rv != SQLITE_OK) {
+		sqlite3_close(snapshot->conn);
+		if (rv == SQLITE_NOMEM) {
+			return RAFT_NOMEM;
+		}
+		return RAFT_ERROR;
+	}
 
-/* An example array of snapshot buffers looks like this:
```
letFunny, line 339: I assume that we are preserving the format. Can you check, at least locally because a test even if ideal could be expensive, that creating a snapshot with the old version of dqlite and restoring it in the new one works?

Also, if the "format" of the snapshot documented somewhere in the internal/external documentation. If we need to maintain compatibility then it seems to me like a comment about the assumptions, e.g. there are x types of buffers placed in this order, is worthwhile.

src/fsm.c
```diff
@@ -217,293 +191,277 @@ static int encodeSnapshotHeader(unsigned n, struct raft_buffer *buf)
...
+	snapshot->header = (struct raft_buffer) {
+		.base = header_buffer,
+		.len = header_size,
+	};
+
+	return RAFT_OK;
 }
 
 static int fsm__snapshot(struct raft_fsm *fsm,
 			 struct raft_buffer *bufs[],
```
letFunny, line 354: It is unfortunate that we cannot change this interface to make it more sensible. I think given the constraints the design that you have and putting `snapshot` in the fsm struct is quite a good choice.

src/gateway.c
```diff
@@ -1094,104 +1096,80 @@ static int handle_remove(struct gateway *g, struct handle *req)
...
-
-		database = data;
-		wal = database + n_database;
-	} else {
-		assert(n == 0);
-
-		n_database = 0;
-		n_wal = 0;
+	sqlite3 *conn;
+	int rv = db__open(db, &conn);
```
letFunny, line 1150: I dont remember why we didn't do this in the first place instead of `vfs_find`. I think it is sensible to open a connection when dumping.

src/vfs.h
```diff
@@ -35,28 +35,22 @@ int VfsAbort(sqlite3 *conn);
 /* Performs a controlled checkpoint on conn */
 int VfsCheckpoint(sqlite3 *conn, unsigned int threshold);
 
-/* Make a full snapshot of a database. */
-int VfsSnapshot(sqlite3_vfs *vfs, const char *filename, void **data, size_t *n);
+struct vfsFile {
```
letFunny: This name feels a bit too general, I understand the `file` here is not from general file but from `sqlite3_file`. What about adding something that indicates this is for snapshot like `vfsFileSnapshot`?

src/vfs.c
```diff
@@ -2549,338 +2550,184 @@ int VfsDatabaseNumPages(sqlite3_vfs *vfs,
...
+		assert(rv == SQLITE_OK);
+		return SQLITE_NOMEM;
 	}
 
-	if (database->n_pages != vfsDatabaseGetNumberOfPages(database)) {
-		tracef("corrupt");
-		return SQLITE_CORRUPT;
+	/* == Safety==
+	 * While holding READ_LOCK(0) the database is guaranteed not to change.
+	 */
```
letFunny, line 2596: Perfect! This is amazing to review and for maintainability.

src/vfs.c
```diff
@@ -2549,338 +2550,184 @@ int VfsDatabaseNumPages(sqlite3_vfs *vfs,
...
 
-	wal = &database->wal;
-
-	*n = (size_t)(vfsDatabaseFileSize(database) + vfsWalSize(wal));
-	/* TODO: we should fix the tests and use sqlite3_malloc instead. */
-	*data = raft_malloc(*n);
-	if (*data == NULL) {
-		tracef("malloc");
-		return DQLITE_NOMEM;
+	/* == Safety==
```
letFunny, line 2601: This comment is also amazing, ty

src/vfs.c
```diff
@@ -2549,338 +2550,184 @@ int VfsDatabaseNumPages(sqlite3_vfs *vfs,
...
+	 * It is expected for this routine to run on the main libuv thread. This
+	 * means that no VfsApply or VfsCheckpoint can run. As such, accessing
+	 * frames is safe. Holding READ_LOCK(0) will also prevent any checkpoint
+	 * from running and as such, *the content* of the frames is safe to
+	 * access on any thread, until VfsReleaseSnapshot is called.
+	 */
+	for (unsigned i = 0; i < f->database->wal.n_frames; i++) {
```
letFunny, line 2608: Why is it that in `fsm.c` we ask for a checkpoint before we can snapshot, and in `gateway.c` we don't. If I am reading this correctly this logic here is doing a checkpoint only manually. I see that even when running a checkpoint like in `fsm.c` there is still a race window because we have released the lock.

So why is it that we do a checkpoint in some cases, but we always do a manual checkpoint.

src/vfs.c
```diff
@@ -2549,338 +2550,184 @@ int VfsDatabaseNumPages(sqlite3_vfs *vfs,
...
-	size_t offset;
-	int rv = SQLITE_OK;
+	sqlite3_file *file;
+	int rv = sqlite3_file_control(conn, NULL, SQLITE_FCNTL_FILE_POINTER, &file);
+	assert(rv == SQLITE_OK);
+	struct vfsMainFile *f = (struct vfsMainFile*)file;
 
-	database = vfsDatabaseLookup(v, filename);
-	assert(database != NULL);
+	assert(snapshot->wal.page_count == 0);
```
letFunny, line 2698: We need to write somewhere that is an assumption about the snapshot. What I find a bit weird is that some of the logic encodes this assumption while other parts act as if `snapshot->wal` could contain data and do the right thing. It feels like there was an effort to be generic over this when at the same time assuming that is zero because some places do need it.

src/vfs.c
```diff
@@ -2549,338 +2550,184 @@ int VfsDatabaseNumPages(sqlite3_vfs *vfs,
...
 
-	rv = ftruncate(database->shm.fd, 0);
+	rv = ftruncate(f->database->shm.fd, 0);
 	assert(rv == 0);
-	database->shm.size = 0;
+	f->database->shm.size = 0;
 
-	vfsWalClose(&database->wal);
-	vfsWalInit(&database->wal);
+	assert(snapshot->wal.page_count == 0);
```
letFunny: This assert is duplicated, it is already at the start of the function.

test/unit/test_gateway.c
marco6: The short answer is: to be consistent with the style.

I can only infer what the thought process was, but I think most of the tests here use macros even when not tampering with existing stuff. I see 2 reasons for that:

 - accessing the "fixture" whose type changes, but fields don't (a kind of C "generic"?)
 - having the right file-line reporting when asserting
 
The second one being the reason here: if I wrap things in a function here, the error that `munit` will display will contain the line of code of the assertion, which will be in the helper function. So, it will be impossible to distinguish tow different calls of that in the same test. Clearly, for now, there is only one call per test, but that might change.

test/unit/test_gateway.c
marco6: I also changed the other `ssize_t rv` to `ssize_t write_rv`.

test/unit/test_gateway.c
marco6: Right. The problem is that now we return "NOTFOUND" when the database is not there, so we must first call `OPEN` to create it. I will revert to calling it in the test.

test/unit/test_gateway.c
marco6: The previous interface would return 2 empty files when the database did not exists. Indeed this test was the same as the next one (no calls to `OPEN`, download an empty database). Now these two tests differ:
 - one (this one) opens the database and downloads it (so it is empty, but it has been created)
 - one (not_existent) tries downloading a non-existent one and fails.

src/fsm.c
marco6: This logic does not replace the WAL. The reason is that #710 made sure that when the database was in "normal" mode (= non disk mode), then the WAL part of the snapshot was *always* empty (as if a checkpoint happened right before the snapshot). This logic went in **before** the `v1.17.0` release, so *all clusters with dqlite1.17* and with disk-mode off will have the WAL empty.

This means that we *could* simplify the logic here and remove the WAL entirely. I think that if we ever want to support migration of a cluster from an old 1.17 disk-mode to a 1.18 non-disk-mode, then the we must address that, but it can be done later, if needed.

Last thought on that is: even if checkpointed, from the *logical* perspective, two files exists in a WAL-mode SQLite database. Does it make sense for our interface not to expose that?

src/fsm.c
marco6: Removed. It is funny that it was also *wrong*: the previous check

```C
if (header.main_size + header.wal_size > SIZE_MAX) {
```

was a *tautology* (there is nothing bigger then the maximum value) and a dead branch. In case `main_size` + `wal_size` was above `MAX`, the addition would have overflowed and the logic would just be unsound.

src/fsm.c
marco6: This comment is on the "old" snapshot buffer scheme. We do not need to keep backward maintainability here, but *in the written-out binary format* of the snapshot (and there, buffers do not exist anymore as it is a single big binary file).

Still, it would be nice to have some documentation about the snapshot format in general.

src/fsm.c
marco6: It's a bit more complex than that:
 - `decodeSnapshot` needs the page_size.
 - `restoreSnapshot` needs the filename.
 - the snapshot needs to be freed after the restore.
 
 Let's discuss a little bit offline on how to properly refactor this part.

src/fsm.c
marco6: Added.

src/vfs.h
marco6: I went for `vfsSnapshotFile`, so that the prefix matches with `vfsSnapshot`.

src/vfs.c
marco6: In my mind the difference is that snapshots are somehow "rare" and initiated by raft logic, while `DUMP` command is user-initiated. IMHO it would feel a little bit weird for a user to be able to checkpoint by dumping the db.

src/vfs.c
marco6: See the comment above for why it is 0. We can decide to remove this.

src/registry.h
marco6: I would prefer two distinct functions. I will rename to `get_or_create`.

src/vfs.h
marco6: Added.

src/vfs.h
```diff
@@ -35,28 +35,34 @@ int VfsAbort(sqlite3 *conn);
...
- * state of the VFS.
+struct vfsSnapshot {
+	struct vfsSnapshotFile main;
+	struct vfsSnapshotFile wal;
+};
+
+/* Acquires a snapshot from the connection conn. The snapshot wil be valid until
+ * VfsReleaseSnapshot is called.
+ *
+ * An acquired snapshot will take relevant lock on the database to make sure
```
letFunny:
```suggestion
 * An acquired snapshot will take relevant locks on the database to make sure
```

test/unit/test_gateway.c
letFunny: Okay, then let's keep it like this. I don't have a strong opinion and the upsides you mention are enough to justify it.

test/unit/test_gateway.c
letFunny: But just for my own sanity, are we still testing dump without a connection?

src/fsm.c
letFunny: > if we ever want to support migration of a cluster from an old 1.17 disk-mode to a 1.18 non-disk-mode

Considering disk mode was never finished and was always experimental I would say we should never invest a lot of time into this unless there is a clear ask.

> Last thought on that is: even if checkpointed, from the logical perspective, two files exists in a WAL-mode SQLite database. Does it make sense for our interface not to expose that?

That is not exactly true, right? When the database is closed, as in the last connection is closed, then the WAL file is removed and there is only the db. You can think of a snapshot taking one of those "consistent" copies.

Anyway, I am okay with leaving the WAL file and documenting that it is always going to be empty and that it _could_ be removed. We do not need to do it now, the cost is not that high to keep it.

src/fsm.c
letFunny: Yeah, I saw you changed it to be a subtraction.

src/fsm.c
letFunny: I remember you already made the effort to create a tool to read it, if you have some documentation about the format handy maybe it is a good time to put it somewhere in the project.

src/vfs.c
letFunny: Okay, I can see the difference in the layer which invokes the command. At the same time, `DUMP` should also be rare, and could also benefit for _trying_ to do an snapshot.

src/vfs.c
letFunny: I would leave the assert and document it. But I would make it consistent everywhere. Or remove the wal, but that is unnecessary for now, IMO.

PR #818, opened 2025-08-29
chore: move debian packaging definition from dqlite-ppa

marco6:
This PR imports the debian package files from canonical/dqlite-ppa.

It also changes the name of the published package to `libdqlite1.18-unstable[-dev]` as it will be build at every commit from master.


.github/workflows/packages.yml
```diff
@@ -36,10 +32,6 @@ jobs:
       run: |
           echo "$PPA_SECRET_KEY" > private-key.asc
           gpg --import --batch private-key.asc
-
-    - name: Delete GPG signing key file
-      if: always()
-      run: |
           rm -f private-key.asc
 
     - name: Build source package
```
letFunny, line 26: Ideally, the private key should only be imported when it is really needed. I assume by not being password protected and being imported into the key ring it could be exfiltrated by using a script in the build process.

If only uploading needs the key maybe reordering the steps would be better but I don't know enough about it to tell.

letFunny (approved):
As discussed offline the plan is to move the debian packaging to this repo so approving the PR.

.github/workflows/packages.yml
marco6: Good point. In general this can only run on specific branches (master in this case), but better safe than sorry.

PR #821, opened 2025-09-01
perf: track read index on databases

marco6:
This PR adds tracking for the a read_index on each database, so that read transactions can start earlier.

The problem this is trying to solve is one that arises when multiple databases are part of the state: in this case a write to one of the database will make *all the others* wait which does not make a lot of sense.

This PR does not change the consistency model [described in the docs](https://canonical.com/dqlite/docs/explanation/consistency-model):

> However, stale reads are rare in practice. They can happen only in case a read transaction is started on a leader that was disconnected from a majority of voting nodes but that didn’t yet realise it and didn’t yet step down from leader. If in the meantime another leader was elected by the rest of the voting nodes and that leader committed a write transaction before the old leader started serving the read transaction, then that read transaction might observe stale data, since read transactions are served by a leader without making sure that it is actually still the leader and hence no write transaction might have happened before (so called “quorum reads”).

> In case a stale reads actually occurs, the data observed by the read transaction can’t be arbitrarily old. A disconnected leader will take at most an election timeout to realise it’s not leader anymore, therefore the data it serves can be at most a few hundred milliseconds old (with default Dqlite settings).

Superseeds #781.

### DIFF
--- a/test/unit/test_replication.c
+++ b/test/unit/test_replication.c
@@ -488,7 +488,7 @@ TEST(replication, barriers, setUp, tearDown, 0, NULL)
 	/* Right after becoming a leader, the node needs to append a barrier
 	 * before being able to append new items. Make sure that executing a SQL
 	 * statement will not issue another barrier, but will wait for the cluster
-	 * to apply the barrier. */
+	 * to apply that barrier. */
 	struct raft *r = raft_fixture_get(&f->cluster, 0);
 	int last_index = raft_last_index(r);
 	munit_assert_int(raft_last_applied(r), <, last_index);
@@ -498,8 +498,10 @@ TEST(replication, barriers, setUp, tearDown, 0, NULL)
 	munit_assert_int(raft_last_index(r), ==, last_index);
 	munit_assert_int(raft_last_applied(r), ==, last_index);
 
-
-	/* Insert an empty command */
+	/* Insert a custom barrier in the log. It is then possible to check if
+	 * any command issues a real barrier because then the custom one will be
+	 * awaited. If the custom barrier didn't complete it means no query
+	 * issued another barrier. */
 	bool barrier_done = false;
 	struct raft_barrier barrier = {
 		.data = &barrier_done,


marco6:
Ready as  #809 was merged.

test/unit/test_gateway.c
```diff
@@ -628,6 +628,13 @@ TEST_CASE(prepare, barrier_error, NULL)
 	raft_fixture_append_fault(&f->cluster, 0, 0);
 	int rv = raft_barrier(CLUSTER_RAFT(0), &faulty_barrier, barrierCb);
 	munit_assert_int(rv, ==, 0);
+	
+	/* Make sure all database require reading the last index. */
```
letFunny:
```suggestion
	/* Make sure all databases require reading the last index. */
```

src/leader.c
```diff
@@ -39,7 +39,16 @@ static int progress_abort(void *db) {
...
-	return raft_last_applied(l->raft) < raft_last_index(l->raft);
+	if (sqlite3_txn_state(l->conn, NULL) != SQLITE_TXN_NONE) {
+		/* If a transaction is already in progress, there is little
+		 * benefit in submitting a barrier as the read mark is already
+		 * set. */
+		return false;
+	}
+	if (l->db->read_index == 0) {
+		l->db->read_index = raft_last_index(l->raft);
+	}
```
letFunny, line 50: Check my assumption: We want to set read_index here at the last possible moment before writing to actually get the last read index, that is the reason you put it in the barrier because logically, that is the previous step to writing and no one is going to write meanwhile.

letFunny (commented):
This is neat Marco. I am only missing one thing and then I can approve. I think we need a test that checks that that a barrier is not triggered when the `read_index` is below the threshold and, conversely, it is in fact triggered when it goes over last_applied.

src/leader.c
marco6: This assumption is only partially true: this logic will also run (and set the read_index!) for *reads* not just writes. And the reason for this trick is that I didn't want to change too much the logic in other parts: when becoming a leader (which, for real, it's the only case when we actually want a barrier), the logic will first make us leaders and then append a barrier. This is logic right? As only the leader can append.

However this way we create a problem as the hook that tells us "you are leader now" (`state_cb` in `server.c`) runs before the barrier takes place, so setting `read_index` there would put us in a weird situation as we might refer to an entry that does not exists.

As such I just "invalidate" it. Immediately after the log is appended (synchronously on the libuv thread) and the next query will fill this item.

test/unit/test_replication.c
```diff
@@ -459,6 +463,64 @@ TEST(replication, exec, setUp, tearDown, 0, NULL)
...
+{
+	struct fixture *f = data;
+
+	/* Make sure work is faster than applying a barrier */
+	raft_fixture_set_work_duration(&f->cluster, 0, 0);
+
+	CLUSTER_ELECT(0);
+
+	/* Right after becoming a leader, the node needs to append a barrier
+	 * before being able to append new items. Make sure that executin an SQL
```
letFunny:
```suggestion
	 * before being able to append new items. Make sure that executing a SQL
```

test/unit/test_replication.c
```diff
@@ -459,6 +463,64 @@ TEST(replication, exec, setUp, tearDown, 0, NULL)
...
+	PREPARE(0, "SELECT 1");
+	EXEC(0);
+	munit_assert_false(barrier_done);
+
+	/* Write transactions do not need a barrier, too */
+	PREPARE(0, "BEGIN IMMEDIATE");
+	EXEC(0);
+	munit_assert_false(barrier_done);
+
+	raft_fixture_step_until(&f->cluster, barrierDone, &barrier_done, 1000);
```
letFunny, line 523: Why do we step here without asserting anything? Shouldn't we assert barrier_done is false?

test/unit/test_replication.c
```diff
@@ -459,6 +463,64 @@ TEST(replication, exec, setUp, tearDown, 0, NULL)
...
+	return *barrier_done;
+}
+
+/* Make sure that no barrier is ever issued for a write transaction. */
+TEST(replication, barriers, setUp, tearDown, 0, NULL)
+{
+	struct fixture *f = data;
+
+	/* Make sure work is faster than applying a barrier */
+	raft_fixture_set_work_duration(&f->cluster, 0, 0);
```
letFunny, line 484: Why is this needed?

letFunny (commented):
Thanks for the tests, a few comments :P

test/unit/test_replication.c
marco6: We already asserted that `barrier_done` is false 2 lines above, didn't we? We are just letting the barrier finish after the exec (not because we need to, but because if the `barrierCb` is invoked it would have a dangling pointer to `barrier_done`).

test/unit/test_replication.c
marco6: The normal behaviour in the fixture is to "wait" for "200ms" before calling the work handler (to simulate a "real" workload), but that is way higher than the time it takes for a barrier to be replicated.

letFunny (approved):
Thanks Marco!

PR #827, opened 2025-09-04
refactor: simplify fsm interface

marco6:
This PR simplifies the fsm interface by removing unnecessary indirection and useless fields. It was part of #774 which is likely not going to be merged.


letFunny (approved):
I guess previously this was done for extensibility of the Raft library which was a library on its own for public consumption. Because that is no longer true and it is dqlite-specific, this makes sense, approving, thanks!

PR #830, opened 2025-09-05
refactor: improve VFS open path

marco6:
This PR moves the responsibility of correctly opening a connection *in the VFS*.

This both simplifies the logic and keeps things together.

It also solves a minor memory leak in `db__open`.

### DIFF
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -1605,6 +1605,8 @@ static int vfsPragmaDeleteDatabase(struct vfsMainFile *f, char **fcntl)
 	return SQLITE_OK;
 }
 
+/* The support for the setter of this PRAGMA is necessary for backward
+ * compatibility. */
 static int vfsPragmaPageSize(struct vfsMainFile *f, char **fcntl)
 {
 	if (fcntl[2] == NULL) {


src/vfs.h
```diff
@@ -2,12 +2,18 @@
 #define VFS_H_
 
 #include <sqlite3.h>
+#include <stddef.h> /* for size_t */
+#include <stdint.h> /* for uint{32,64}_t */
 
-#include "config.h"
+struct vfsConfig {
+	char name[256];                /* VFS/replication registriatio name */
```
letFunny:
```suggestion
	char name[256];                /* VFS/replication registration name */
```

src/config.c
```diff
@@ -23,39 +23,45 @@
...
-	rv = snprintf(c->name, sizeof c->name, "dqlite-%u", serial);
-	assert(rv < (int)(sizeof c->name));
-	c->logger.data = NULL;
-	c->logger.emit = loggerDefaultEmit;
-	c->failure_domain = 0;
-	c->weight = 0;
+	
+	unsigned vfs_id = atomic_fetch_add_explicit(&serial, 1, memory_order_relaxed);
+	int rv = snprintf(c->vfs.name, sizeof c->vfs.name, "dqlite-%u", vfs_id);
+	assert(rv < (int)(sizeof c->vfs.name));
```
letFunny: Nit: we could check `rv > 0` although I wouldn't expect this to ever fail.

src/vfs.c
```diff
@@ -2479,20 +2458,145 @@ static int vfsGetLastError(sqlite3_vfs *vfs, int x, char *y)
...
+	sqlite3_set_authorizer(db, vfsAuthorizer, NULL);
+
+	return SQLITE_OK;
+}
+
+int VfsInit(struct sqlite3_vfs *vfs, const struct vfsConfig *config)
 {
 	tracef("vfs init");
 
+	int rv = sqlite3_auto_extension((void(*)(void))vfsOpenHook);
```
letFunny, line 2589: Neat!

src/vfs.c
```diff
@@ -1607,39 +1611,14 @@ static int vfsPragmaPageSize(struct vfsMainFile *f, char **fcntl)
...
-
-	if (page_size < FORMAT__PAGE_SIZE_MIN ||
-	    page_size > FORMAT__PAGE_SIZE_MAX) {
-		fcntl[0] = sqlite3_mprintf("page size outside of valid range");
-		return SQLITE_IOERR;
-	}
 
-	if (!is_po2((unsigned long)page_size)) {
-		fcntl[0] = sqlite3_mprintf("page size must be a power of 2");
+	unsigned page_size = (unsigned)atoi(fcntl[2]);
```
letFunny, line 1617: I see that we are removing the setter, it basically just checks the size is the same as the existing one. I assume this is for backwards compatibility so a comment would help there.

src/fsm.c
```diff
@@ -84,13 +84,6 @@ static int apply_frames(struct fsm *f, const struct command_frames *c)
 		goto error;
 	}
 
-	int ckpt_rv = VfsCheckpoint(conn, db->config->checkpoint_threshold);
-	if (ckpt_rv == SQLITE_BUSY) {
-		tracef("checkpoint: busy reader or writer");
-	} else if (ckpt_rv != SQLITE_OK) {
-		tracef("checkpoint failed: %d", ckpt_rv);
-	}
```
letFunny, line 92: How come we are moving the checkpoint from one function to another without any external documentation changing. `VfsApply` should now say it will attempt to perform a checkpoint.

I think your rationale was that vfs has configuration and other functions already and would be more self-contained if we moved the checkpoint to `vfs.c`. Is that correct? Or are there more reasons?

src/vfs.c
```diff
@@ -2840,35 +2958,24 @@ int VfsAbort(sqlite3 *conn)
...
-	rv = vfsShmLock(&f->database->shm, 0, SQLITE_SHM_NLOCK, true);
+	int rv = vfsShmLock(&f->database->shm, 0, SQLITE_SHM_NLOCK, true);
 	if (rv != SQLITE_OK) {
 		tracef("[database %p] checkpoint busy", f->database);
 		return rv;
 	}
+	f->exclMask = VFS__CHECKPOINT_MASK;
 
 	PRE(f->database->wal.n_tx == 0);
-	if (f->database->wal.n_frames < threshold) {
```
letFunny, line 2862: Much simpler if the locks are not taken if the threshold is not reached. Control flow like this should be moved up just like you did.

src/fsm.c
```diff
@@ -309,7 +302,7 @@ static int snapshotDatabase(struct db *db, struct fsmDatabaseSnapshot *snapshot)
 		return RAFT_ERROR;
 	}
 
-	rv = VfsCheckpoint(snapshot->conn, 0);
+	rv = VfsCheckpoint(snapshot->conn);
```
letFunny, line 305: The new code changes the behavior a bit as the threshold is no longer taken into account here and we instead checkpoint every time. I think the new behavior is correct.

test/integration/test_stress.c
```diff
@@ -63,7 +63,27 @@ static void *client_read(void *data)
...
+			}
+			clientCloseRows(&rows);
+			rows = (struct rows){};
+		}
+		if (rv == DQLITE_CLIENT_PROTO_RECEIVED_FAILURE) {
+			if (client.errcode == SQLITE_BUSY) {
+				/* Just retry */
+				i--;
+				continue;
+			}
```
letFunny, line 81: I guess this change is unrelated to the PR and only making the test more robust? Or what is the reasoning behind it?

test/unit/test_vfs.c
```diff
@@ -141,30 +146,23 @@ static void *__buf_page_2(void)
...
-	rc = sqlite3_exec(db, sql, NULL, NULL, NULL);
-	munit_assert_int(rc, ==, SQLITE_OK);
-}
+#define EXEC(DB, SQL)                                                  \
+	do {                                                           \
+		int exec_rc = sqlite3_exec(DB, SQL, NULL, NULL, NULL); \
+		munit_assert_int(exec_rc, ==, SQLITE_OK);                   \
+	} while (0)
 
 /* Helper to open and initialize a database, setting the page size and
```
letFunny: More or less inaccurate now that the page size is set elsewhere. Also, not only page size and WAL mode are set, there are a bunch of other things. I think it's better to drop the comment.

test/unit/test_vfs_extra.c
```diff
@@ -69,32 +75,19 @@ static void tearDownRestorePendingByte(void *data)
...
-		_rv = sqlite3_extended_result_codes(DB, 1);                   \
-		munit_assert_int(_rv, ==, SQLITE_OK);                         \
-		PRAGMA(DB, "page_size=512");                                  \
-		PRAGMA(DB, "synchronous=OFF");                                \
-		PRAGMA(DB, "journal_mode=WAL");                               \
-		PRAGMA(DB, "cache_size=1");                                   \
-		_rv = sqlite3_db_config(DB, SQLITE_DBCONFIG_NO_CKPT_ON_CLOSE, \
-					1, NULL);                             \
-		munit_assert_int(_rv, ==, SQLITE_OK);                         \
+#define OPEN(VFS, DB)                                                          \
```
letFunny, line 79: Why not use db__open now? I see how the PR simplifies a lot of things but, to me, one great benefit would be to use the same initialization code in all the tests (at least part of it).

This comment applies to all test, I see other helpers like `test_db_open` and probably all other test suites.

src/vfs.c
marco6: Note that it was never a real setter. If you read the old logic right if `database->n_pages` was set (> 0) then either the parameter matched the old value or an error was returned. And the size was set at the first open (in `db__open`) so, the user never had the chance to use this pragma and the only possible value is the one set in the configuration. So, no changes on this front: it was never possible to set the page size.

src/fsm.c
marco6: True on the `VfsApply` comment.

Regarding the behaviour, there should be no real difference actually. Before this change that was handled by `fsm__apply` which would take care of trying a checkpoint. Indeed, we do want to have the default SQLite behaviour which is to try a checkpoint after every write transaction, so that the size of the WAL doesn't grow too much. However, we cannot enable the normal flow as SQLite would run it right after it thinks that the commit took place, but at that point we still must reach quorum, so we must delay it.

Now, I think that this is a detail that belongs to the VFS. Actually, the FSM should not even know which mode the connection is in or what a WAL is: it should just take care of taking the command and try to apply it IMHO. Last, it is also somehow faster as we don't need to query the connection again for the file pointer and we don't need to lock once again to read the number of frames we just updated.

src/fsm.c
marco6: Consider that at this point even in the previous logic, the threshold wasn't taken into account as we passed the `0` constant as a limit.

test/integration/test_stress.c
marco6: Right. It is related actually to #811: before that PR, checkpoints run only when reaching 1000 frames. The problem with that is that in all scenarios so far, even with a big number of readers, readers end way before reaching that limit (as reads can run in parallel).

This means that in the stress test never happened that a checkpoint run (or even attempted to run) while readers are active.

When adding a checkpoint "right before a snapshot" we lowered that limit a lot, especially with multiple databases in place as now it is run "when we added X entries in the log". Consider that the default is `1024` (see `src/raft/raft.c#96`), so for example in the 4 database the first checkpoint will be attempted when on average the number of transactions is `256`, not `1024`.

That is enough to make it happen during the reads (which rightfully then fail with `SQLITE_BUSY`).

This PR made it more evident as now the check for "do we need a threshold" does not need to take the write lock and in general it is a bit faster, just enough to fail a lot of times during my tests on my machine. Likely this is also the cause of the [current failure on riscv64](https://launchpad.net/~dqlite/+archive/ubuntu/dev/+build/31306633).

test/unit/test_vfs_extra.c
marco6: I think it wouldn't make a lot of sense actually. `db__open` is a higher level verb which has nothing to do with the VFS.

Historically it was indeed the way to "properly open a connection", but that kind of mixed the responsibilities which is what this PR fixes: now all logic and all required settings on SQLite are set *by* the VFS itself.

To summarize I think it is a good thing that to test the behaviour of our SQLite VFS we only need SQLite API now.

src/vfs.c
letFunny: What I mean is that this pragma as a setter is only supported now because of historical reasons, if this were not in past versions of the code we would remove it. If that is the case a very short comment would help.

src/fsm.c
letFunny: > Now, I think that this is a detail that belongs to the VFS. Actually, the FSM should not even know which mode the connection is in or what a WAL is: it should just take care of taking the command and try to apply it IMHO. Last, it is also somehow faster as we don't need to query the connection again for the file pointer and we don't need to lock once again to read the number of frames we just updated.

I also agree with all this. Especially the bit about separating concerns, the checkpoint should be a concept only known to the VFS.

test/integration/test_stress.c
letFunny: This is wonderful!

test/unit/test_vfs_extra.c
letFunny: I think I say this in almost every PR but I think there is a huge benefit in the tests having the same initialization sequence as the real code. That means we are testing the exact same set of options without having to keep them up to date every time. To me, db__open is a function that gets you a configured sqlite connection but I see your point that now it is just a wrapper that is setting only options like extended return codes while the real logic already happens in the VFS. I worry about the function growing bigger with some other steps that are not placed in the VFS and the tests getting our of sync.

This is all probably over-paranoia on my end so I am not blocking on this.

letFunny (approved):
Pending the minor point about adding a comment about backwards compatibility (https://github.com/canonical/dqlite/pull/830#discussion_r2375016437). But treat it as a nit, it is not blocking, you can go ahead and merge if you don't think it is appropriate.

src/vfs.c
marco6: Added.

test/unit/test_vfs_extra.c
marco6: As a very general thing, I think that there is an overlap between `db` and `vfsDatabase` structs and that there is an overlap between the `registry` and the `vfs` itself: they both contain at the end the same data: the list of databases that are available to the dqlite instance. Indeed, they even need to be kept in sync by a hook!

I had a spike few weeks ago that indeed tried to remove the registry and the db, so that this two entangled concepts would merge (so, no more `db__open` at all!), but it become complex and I decided that for now, it's best not to have a PR doing refactoring for no reason.

I will mark this as resolved for now.

PR #839, opened 2025-09-11
bugfix: multi-region shm mapping for systems with big pages

marco6:
This PR fixes the memory allocation pattern for all architectures/systems that have a page > 32KiB.

### DIFF
--- a/src/vfs.c
+++ b/src/vfs.c
@@ -90,6 +90,18 @@ const int vfsOne = 1;
 
 #define vfsFrameSize(PAGE_SIZE) (VFS__FRAME_HEADER_SIZE + PAGE_SIZE)
 
+/*
+ * Return a size that can be used to allocate at least one of 32KiB shm regions,
+ * assuming that each mapping must be an integer multiple of the current system
+ * page-size.
+ *
+ * Usually, this is 32KiB. The exception seems to be systems that are configured
+ * to use 64KB pages like POWER architecture, where the default is 64KiB - in
+ * this case each mapping must cover at least two regions or the mmap call will
+ * fail.
+ *
+ * See also https://github.com/sqlite/sqlite/blob/eb8089e/src/os_unix.c#L4546
+ */
 static size_t vfsGetMapSize(void)
 {
 	const int os_page_size = (int)sysconf(_SC_PAGESIZE);
@@ -1656,6 +1668,7 @@ static int vfsPublishShm(struct vfsMainFile *f)
 		void *region = mmap(
 		    NULL, map_size, PROT_READ | PROT_WRITE, MAP_SHARED,
 		    f->database->shm.fd, i * VFS__WAL_INDEX_REGION_SIZE);
+		assert(region != MAP_FAILED);
 		memcpy(region, f->mappedShmRegions.ptr[i], map_size);
 		void *remapped = mremap(
 		    region, map_size, map_size,
--- a/test/integration/test_cluster.c
+++ b/test/integration/test_cluster.c
@@ -205,10 +205,8 @@ TEST(cluster, dataOnNewNode, setUp, tearDown, 0, cluster_params)
 	return MUNIT_OK;
 }
 
-/* Insert a huge row, causing SQLite to allocate overflow pages.
- * Then insert the same row again to make sure that it also needs an additional
- * shared memory region. (Reproducer for
- * https://github.com/canonical/raft/issues/432.) */
+/* Insert a huge row, causing SQLite to allocate overflow pages and additional
+ * shm regions. Then insert the same row again. */
 TEST(cluster, hugeRow, setUp, tearDown, 0, NULL)
 {
 	struct fixture *f = data;
@@ -241,9 +239,10 @@ TEST(cluster, hugeRow, setUp, tearDown, 0, NULL)
 	return MUNIT_OK;
 }
 
-/* Insert a huge row, causing SQLite to allocate overflow pages. Then
- * insert the same row again. (Reproducer for
- * https://github.com/canonical/raft/issues/432.) */
+/* Insert a huge row, causing SQLite to allocate overflow pages and additional
+ * shm regions. Then insert the same row again.
+ * Then rollback so that the rollback logic spawning multiple regions is
+ * exercised. */
 TEST(cluster, hugeRow_rollback, setUp, tearDown, 0, NULL)
 {
 	struct fixture *f = data;


src/vfs.c
```diff
@@ -1476,23 +1490,33 @@ static int vfsMainFileShmMap(sqlite3_file *file, /* Handle open on database file
 			 void volatile **out /* OUT: Mapped memory */
 )
 {
-	const int offset = region_index * region_size;
+	const off_t offset = (off_t)(region_index * region_size);
```
letFunny, line 1505: Nit: let's move this closer to usage.

src/vfs.c
```diff
@@ -90,6 +90,21 @@ const int vfsOne = 1;
 
 #define vfsFrameSize(PAGE_SIZE) (VFS__FRAME_HEADER_SIZE + PAGE_SIZE)
 
+static size_t vfsGetMapSize(void)
+{
+	const int os_page_size = (int)sysconf(_SC_PAGESIZE);
```
letFunny, line 107: Can you write a comment above this function saying why it is needed and the context. IMO we should explain:
* Page size can be bigger than `VFS__WAL_INDEX_REGION_SIZE` and this happens in architectures like PowerPC.
* SQLite already handles this correct and permalink to check their implementation.

src/vfs.c
```diff
@@ -90,6 +90,21 @@ const int vfsOne = 1;
 
 #define vfsFrameSize(PAGE_SIZE) (VFS__FRAME_HEADER_SIZE + PAGE_SIZE)
 
+static size_t vfsGetMapSize(void)
```
letFunny, line 105: Nit: I don't think the name is very telling but I cannot come up with something good. What about `vfsGetMinMapSize`?

src/vfs.c
```diff
@@ -1517,23 +1542,19 @@ static int vfsMainFileShmMap(sqlite3_file *file, /* Handle open on database file
 		f->mappedShmRegions.ptr = newPtr;
 	}
 
-	if (region_index < f->mappedShmRegions.len) {
```
letFunny, line 1520: Funny how we didn't put this at the beginning the last review, it is so much simpler this way.

src/vfs.c
```diff
@@ -1601,84 +1628,100 @@ static int vfsRedirectShm(struct vfsMainFile *f) {
...
-		assert(region != MAP_FAILED);
-		memcpy(region, f->mappedShmRegions.ptr[i], VFS__WAL_INDEX_REGION_SIZE);
-		void *remapped = mremap(region, 
-			VFS__WAL_INDEX_REGION_SIZE, VFS__WAL_INDEX_REGION_SIZE, 
-			MREMAP_MAYMOVE | MREMAP_FIXED, f->mappedShmRegions.ptr[i]);
+	for (int i = region_per_map; i < f->mappedShmRegions.len;
+	     i += region_per_map) {
+		void *region = mmap(
+		    NULL, map_size, PROT_READ | PROT_WRITE, MAP_SHARED,
+		    f->database->shm.fd, i * VFS__WAL_INDEX_REGION_SIZE);
```
letFunny, line 1670: Why did you remove the `assert(region != MAP_FAILED)`?

src/vfs.c
```diff
@@ -1689,18 +1732,24 @@ static int vfsPublishShm(struct vfsMainFile *f) {
...
  * See vfsRedirectShm for a rationale. */
 static int vfsRollbackShm(struct vfsMainFile *f) {
-	for (int i = f->mappedShmRegions.len - 1; i >= 0; i--) {
-		void *new_region = mmap(NULL, VFS__WAL_INDEX_REGION_SIZE,
-			PROT_READ | PROT_WRITE, MAP_SHARED, f->database->shm.fd, i * VFS__WAL_INDEX_REGION_SIZE);
+	const size_t map_size = vfsGetMapSize();
+	const int region_per_map = (int)(map_size / VFS__WAL_INDEX_REGION_SIZE);
+	PRE((f->mappedShmRegions.len % region_per_map) == 0);
+
+	for (int i = 0; i < f->mappedShmRegions.len; i += region_per_map) {
```
letFunny, line 1752: We are now iterating from 0 to len instead of from len to 0. I seem to remember this was a conscious choice when publishing / rolling back.

test/integration/test_cluster.c
```diff
@@ -205,10 +205,46 @@ TEST(cluster, dataOnNewNode, setUp, tearDown, 0, cluster_params)
...
+	char *sql;
+	ssize_t n;
+	size_t huge = 20000000;
+	(void)params;
+
+	HANDSHAKE;
+	OPEN;
+	PREPARE(
+	    "CREATE TABLE IF NOT EXISTS model(key TEXT, value TEXT, "
+	    "UNIQUE(key))",
```
letFunny: Not actionable: I don't know if you changed the formatter but seeing strings like this in two sentences is much worse than having slightly above 80 characters IMO.

test/integration/test_cluster.c
```diff
@@ -205,10 +205,46 @@ TEST(cluster, dataOnNewNode, setUp, tearDown, 0, cluster_params)
...
+	/* Again */
+	EXEC(stmt_id, &last_insert_id, &rows_affected);
+	return MUNIT_OK;
+}
+
 /* Insert a huge row, causing SQLite to allocate overflow pages. Then
  * insert the same row again. (Reproducer for
  * https://github.com/canonical/raft/issues/432.) */
-TEST(cluster, hugeRow, setUp, tearDown, 0, NULL)
+TEST(cluster, hugeRow_rollback, setUp, tearDown, 0, NULL)
```
letFunny, line 246: Comment is outdated know.

test/integration/test_cluster.c
```diff
@@ -205,10 +205,46 @@ TEST(cluster, dataOnNewNode, setUp, tearDown, 0, cluster_params)
...
+	    sql, huge,
+	    "INSERT OR REPLACE INTO model (key, value) VALUES('my-key-1', '");
+	memset(sql + n, 'A', huge - n);
+	memcpy(sql + huge - 3, "')", 3);
+	PREPARE(sql, &stmt_id);
+	free(sql);
+	EXEC(stmt_id, &last_insert_id, &rows_affected);
+	/* Again */
+	EXEC(stmt_id, &last_insert_id, &rows_affected);
+	return MUNIT_OK;
```
letFunny, line 239: In both tests we are not asserting anything about the execs, is that expected? I mean we are not querying the db, or verifying the rows affected, or anything like that.

src/vfs.c
marco6: I think this is not necessary. We should go "backwards" when publishing (actually we should first publish hash maps, then the header in the right order), but here we are reverting state *for the current connection* (other connections will not see this are we are basically discarding everything!). 

I don't know why we kept it but it seems we just missed something in the review.

test/integration/test_cluster.c
marco6: TBH I think we should increase to 120 at least. 80 doesn't really make sense IMHO.

test/integration/test_cluster.c
marco6: Ah you are hitting a typical problem with this codebase: macros. This is an integration test, so an `EXEC` is very different from an `EXEC` in the unit-test. This is in particular:

```C
#define EXEC(STMT_ID, LAST_INSERT_ID, ROWS_AFFECTED) EXEC_C(f->client, STMT_ID, LAST_INSERT_ID, ROWS_AFFECTED)

#define EXECSTMT_ID, LAST_INSERT_ID, ROWS_AFFECTED) \
	{ \
		int rv_; \
		struct value vals_[] = {__VA_ARGS__}; \
		size_t len_ = sizeof(vals_) / sizeof(vals_[0]); \
		rv_ = clientSendExec(f->client, STMT_ID, vals_, len_, NULL); \
		munit_assert_int(rv_, ==, 0); \
		rv_ = clientRecvResult(f->client, LAST_INSERT_ID, \
				       ROWS_AFFECTED, NULL); \
		munit_assert_int(rv_, ==, 0); \
	}
```

`clientRecvResult` will:
  - internally call `abort` in case of `oom` (!? I have a feeling this is a taint from a crab friend :crab: ...)
  - wait for the response
  - return error `!= 0` in case the message doesn't match
 
 And the macro asserts that we don't get a response error. So this logic *is* checking that the exec succeeded!

src/vfs.c
marco6: That was a mistake.

src/vfs.c
marco6: I think that this is not the minimum map size? If the page is 4KiB we return 32KiB for the mapping?

On the other side it *is* the size that we always pass to `mmap`, so I think that the name stands?

src/vfs.c
marco6: Done

src/vfs.c
letFunny: Why?

src/vfs.c
letFunny: It is the minimum size we can map _for a region_ but that makes name hard.

src/vfs.c
marco6: It was a mistake removing it! Now it's back

PR #841, opened 2025-09-15
fix: always call `VfsAbort` on error

marco6:
This PR fixes the leader state machine so that it always call `VfsAbort` in case something went wrong during replication. 

This is yet another instance of problems created by the double return path: `raft_apply` will call the callback with the error normally EXCEPT that when it already knows it is not the leader, then it will not and it will just return immediately without calling any callback.

This misdesign is plaguing this codebase and should be avoided in the future: only systematic errors like "wrong parameters" or "out of memory" should be returned immediately, **not** domain errors.

The fix for this specific bug is to route all errors through the state machine, so that now the state `EXEC_WAITING_APPLY` is not a phantom (trace-only) state anymore, but takes care of the required cleanup.

### DIFF
--- a/src/leader.c
+++ b/src/leader.c
@@ -29,6 +29,11 @@ static bool is_db_full(sqlite3_vfs *vfs, struct db *db, unsigned nframes);
 static struct exec *exec_dequeue(struct db *db);
 static void exec_enqueue(struct db *db, struct exec *exec);
 
+static int progress_abort(void *db) {
+	(void)db;
+	return SQLITE_ABORT;
+}
+
 /* Whether we need to submit a barrier request because there is no transaction
  * in progress in the underlying database and the FSM is behind the last log
  * index. */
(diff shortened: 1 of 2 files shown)


src/leader.c
```diff
@@ -523,10 +523,8 @@ static void exec_tick(struct exec *req)
 				if (rc != SQLITE_OK) {
 					leader_trace(leader,
 						     "poll failed on leader");
-					rc = VfsAbort(leader->conn);
-					assert(rc == SQLITE_OK);
 					req->status = RAFT_IOERR;
-					sm_move(&req->sm, EXEC_DONE);
+					sm_move(&req->sm, EXEC_WAITING_APPLY);
```
letFunny, line 527: Can we update the diagram on top of the file? I think there is no transition to EXEC_DONE anymore. We might also need to update the text on suspend, I don't think it is longer accurate that we always suspend on EXEC_WAITING_APPLY.

marco6: > I think there is no transition to EXEC_DONE anymore.

Wait I think there is? In case `VfsPoll` succeeds, but it is an empty transaction (for reads for example). See like `527`.

Same for suspending: we don't suspend only if there was an error which is covered by 

```C
* All states can also reach `EXEC_DONE` in case of an error.
```

On line `145`

marco6: Just to be clear, it is true that we are also handling the error in `EXEC_WAITING_APPLY`, while we did that in `EXEC_DONE` only before this patch...

letFunny: Okay, right, let's not spend time on this, I think

> All states can also reach `EXEC_DONE` in case of an error.

covers it.

PR #843, opened 2025-09-17
bugfix: work-around XFS mtime limitations

marco6:


src/raft/uv_fs.c
```diff
@@ -148,6 +159,96 @@ int UvFsFileIsEmpty(const char *dir,
 	return 0;
 }
 
+/* Reopens a file descriptor with FMODE_NOCMTIME flag on for XFS. Given that if
+ * this fails the worst thing happening is going back to the thread pool, errors
+ * here are ignored and the reopening just dosn't happen. This will likely be
```
letFunny:
```suggestion
 * here are ignored and the reopening just doesn't happen. This will likely be
```

src/raft/uv_fs.c
```diff
@@ -148,6 +159,96 @@ int UvFsFileIsEmpty(const char *dir,
...
+ * the case, given that (as of Linux 6.17) SYS_ADMIN capability is required for
+ * this flow to suceed.
+ *
+ * This is special as in the case of XFS, `O_DSYNC` is problematic when used in
+ * conjunction with libaio as it fail to be non blocking because it needs to
```
letFunny:
```suggestion
 * conjunction with libaio as it fails to be non blocking because it needs to
```

src/raft/uv_fs.c
```diff
@@ -148,6 +159,96 @@ int UvFsFileIsEmpty(const char *dir,
...
+ * also update the file metadata (the ctime and mtime fields).
```
letFunny: I think we can re-write it a little bit to make it clear that XFS will also update the file metadata (unlike other filesystems) which doesn't work with async.

What about changing `because it needs to` to `because XFS will also update the file metadata (the ctime and mtime fields) unlike in other filesystems. Updating metadata  has to be done synchronously makes async IO fail on XFS. Dqlite only needs to perform a fdatasync to work correctly which is why we need to disable metadata writing on XFS.`?

src/raft/uv_fs.c
```diff
@@ -904,6 +1007,7 @@ int UvFsProbeCapabilities(const char *dir,
 		*async = false;
 		goto out;
 	}
+	fprintf(stderr, "probing aio with: %d\n", fd);
```
letFunny: Is the fd relevant here? I thought when the application crashed or fd is freed the mapping between for the fd would be removed instantly. When looking at the log how do you use the fd number? Plus, fds can be reused.

src/raft/uv_fs.c
```diff
@@ -148,6 +159,96 @@ int UvFsFileIsEmpty(const char *dir,
...
+ * for the mentioned hack.
+ *
+ * TODO: remove this hack if/when we get `O_CMTIME` flag. */
+static int uvXfsMaybeReopen(int fd, const char *dir, int flags)
+{
+#ifdef HAVE_XFS_XFS_H
+	/* Only for direct I/O */
+	if (!(flags & O_DSYNC)) {
+		return fd;
+	}
```
letFunny: The comment seems outdated.

src/raft/uv_fs.c
```diff
@@ -148,6 +159,96 @@ int UvFsFileIsEmpty(const char *dir,
...
+ * behaviour using the syscall `open_by_handle_at` that, as a hack, will
+ * set the internal FMODE_NOCMTIME flag to suppress this behaviour.
+ *
+ * See the linux kernel functions:
+ *  - xfs_file_write_checks
+ *  - kiocb_modified
+ *  - file_modified_flags
+ * for a rational on the checks and:
+ *  - xfs_open_by_handle
+ * for the mentioned hack.
```
letFunny: Neat, thanks for the pointers.

src/raft/uv_fs.c
```diff
@@ -148,6 +159,96 @@ int UvFsFileIsEmpty(const char *dir,
 	return 0;
 }
 
+/* Reopens a file descriptor with FMODE_NOCMTIME flag on for XFS. Given that if
```
letFunny: I think we need to change the documentation to differentiate public / internal docs. The public docs here should say it reopens a file descriptor ... because XFS does not work correctly with async and O_DSYNC and the reasons why (see other comments with suggestions).

Then inside the function we can write the details and pointers to the kernel so, as a developer, I don't need to write the whole paragraph to understand _what_ this does and not the _how_.

letFunny (commented):
Thanks for looking into this Marco. I am waiting for the session where you show us how you found out this was happening in the kernel. There is nothing controversial here, consider it almost-approved. All my comments are about documentation to make it slightly better for the next person reading this, nothing major.

src/raft/uv_fs.c
marco6: Ah this should be removed, it's not even a "tracef", it was something I used while writing the code so I could match the ftrace logs.

src/raft/uv_fs.c
marco6: Question: this is a `static` function which is only exposed at the file level. If someone reaches this point and reads this it is already reading the code, not relying on an interface. Does it make sense in this case to split the comment?

src/raft/uv_fs.c
marco6: Why? I think we need to do this only for `O_DSYNC` (direct I/O) don't we?

letFunny: Is `O_DYSNC` direct I/O? I thought it was `O_DIRECT`

src/raft/uv_fs.c
letFunny: I think it does but maybe I am wrong. When reading the code you also want to troubleshoot quickly and understand what each function is doing. This is in the path for open so anyone looking at it will also have to read what this function does.

src/raft/uv_fs.c
letFunny: Perfect, thanks!

src/raft/uv_fs.c
marco6: Done.

src/raft/uv_fs.c
marco6: Changed to `synchronized` as that is the word used in the man.

PR #850, opened 2025-10-06
fix: fix compression test and explicit assumption

marco6:
This PR fixes a regression in tests caused by #804 .

### DIFF
--- a/src/raft/uv_fs.c
+++ b/src/raft/uv_fs.c
@@ -559,6 +559,9 @@ int UvFsMakeCompressedFile(const char *dir,
 		}
 		content_size += bufs[i].len;
 	}
+
+	/* Limit the size of each chunk to 4MB to make sure this logic doesn't
+	 * need too much memory. */
 	const size_t lz4_max_block_size = 4 * 1024 * 1024;
 	assert(chunk_size <= lz4_max_block_size);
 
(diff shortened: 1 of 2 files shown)


test/raft/unit/test_compress.c
```diff
@@ -182,13 +191,18 @@ TEST(compress,
...
 	/* Fill a buffer with easy-to-compress data */
-	struct raft_buffer buf = {
-		.base = munit_malloc(len),
-		.len = len,
-	};
-	memset(buf.base, 0xAC, buf.len);
+	void *buf = munit_malloc(len);
+	memset(buf, 0xAC, len);
+
+	size_t n_bufs = (len + 4095) / 4096;
```
letFunny, line 191: Nit: Copy the comment here as well: `/* Split the buffer into many chunks of at most 4096 bytes */`.

src/raft/uv_fs.c
```diff
@@ -559,6 +559,8 @@ int UvFsMakeCompressedFile(const char *dir,
 		}
 		content_size += bufs[i].len;
 	}
+	const size_t lz4_max_block_size = 4 * 1024 * 1024;
```
letFunny, line 565: Come on, you almost did it perfectly :). Why not add a comment explaining why there is a restriction that seems arbitrary to document it.

letFunny (approved):

PR #875, opened 2026-03-19
chore: correct const qualifier issue in src/server.c

crypticC0der:
Casting memchr results to const char* which was causing discarded-qualifiers warnings on newer gcc. This means when -Werror is enabled the build fails. I have updated the casts to (*char) to allow modifying the buffer safely.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -1111,7 +1111,7 @@ static int parseNodeStore(char *buf, size_t len, struct node_store_cache *cache)
 	struct client_node_info info;
 
 	version_str = p;
-	nl = memchr(p, '\n', (size_t)(end - version_str));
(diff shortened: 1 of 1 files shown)


tomponline:
@crypticC0der should `-Werror` be enabled by default in the Makefile (if not already)?

crypticC0der:
if you would like i can set that up

crypticC0der:
> should -Werror be enabled by default in the Makefile (if not already)?

It is enabled by default

marco6 (approved):
Looks good, thanks. As I said elsewhere, the CI is failing for unrelated reasons. I will fix it ASAP, so let's wait for that.

PR #878, opened 2026-04-08
feat: set SOCK_CLOEXEC on socket creation to prevent FD leak to child processes

simondeziel:
On Linux 6.17+, AppArmor enforces `AF_UNIX` peer mediation. When the
`dqlite` listening socket FD is inherited by a child process (e.g.
`dnsmasq` spawned by LXD), AppArmor checks the child's profile for
accept permission on `connect()`, causing connections to be denied.

Fix this by passing `SOCK_CLOEXEC` to `socket()` so the FD is
automatically closed on `exec()`.

In `server.c`, the old code attempted to set close-on-exec via
`fcntl(fd, FD_CLOEXEC)`, but this was a bug: `FD_CLOEXEC` has the
integer value 1, which equals the `F_GETFD` command, so the call was
reading the flags instead of setting them. The flag was never
actually applied. The correct form would have been
`fcntl(fd, F_SETFD, FD_CLOEXEC)`, but using `SOCK_CLOEXEC` is
preferred because it is atomic (no race window between `socket()`
and a concurrent `fork()+exec()` in another thread).

In `transport.c`, there was no `CLOEXEC` handling at all.

The accidentally inherited FDs would be noticed on 6.17 kernels causing issues to LXD:

```
[5](https://github.com/canonical/lxd/actions/runs/24151769294/job/70481389830#step:22:466)
+ lxc exec m2 -- lxd init --preseed
Error: Failed to join cluster: Failed request to add member: Failed to get the address of the cluster leader: Failed to get dqlite client: failed to establish network connection: dial unix @58535: connect: permission denied
...
+ dmesg
... apparmor="DENIED" operation="connect" class="net" namespace="root//lxd-m1_<var-snap-lxd-common-lxd>" profile="lxd_dnsmasq-lxdbr0_</var/snap/lxd/common/lxd>" pid=13408 comm="lxd" family="unix" sock_type="stream" protocol=0 requested="send receive accept" denied="send accept" addr="@58535" peer_addr=none peer="unconfined"
...
```

Or Juju:

```
2026-02-11T17:08:33.894226+00:00 reactive-amd64-noble-medium-ps7-1-r-04f90c91f071 kernel: audit: type=1400 audit(1770829713.893:6232): apparmor="DENIED" operation="connect" class="net" info="failed af match" erro
r=-13 namespace="root//lxd-juju-ba1a06-0_<var-snap-lxd-common-lxd>" profile="snap.juju-db.mongod" pid=12439 comm="jujud" family="unix" sock_type="stream" protocol=0 requested="send receive accept" denied="send ac
cept" addr="@dqlite-3297041220608546238" peer_addr=none peer="unconfined"
```

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -244,13 +244,8 @@ int dqlite_node_set_bind_address(dqlite_node *t, const char *address)
 	}
 	domain = addr->sa_family;
 
-	fd = socket(domain, SOCK_STREAM, 0);
+	fd = socket(domain, SOCK_STREAM | SOCK_CLOEXEC, 0);
 	if (fd == -1) {
-		return DQLITE_ERROR;
-	}
-	rv = fcntl(fd, FD_CLOEXEC);
-	if (rv != 0) {
-		close(fd);
 		return DQLITE_ERROR;
 	}
 
--- a/src/transport.c
+++ b/src/transport.c
@@ -223,7 +223,7 @@ int transportDefaultConnect(void *arg, const char *address, int *fd)
 	}
 
 	dqlite_assert(addr->sa_family == AF_INET || addr->sa_family == AF_INET6);
-	*fd = socket(addr->sa_family, SOCK_STREAM, 0);
+	*fd = socket(addr->sa_family, SOCK_STREAM | SOCK_CLOEXEC, 0);
 	if (*fd == -1) {
 		return RAFT_NOCONNECTION;
 	}
--- a/test/lib/endpoint.c
+++ b/test/lib/endpoint.c
@@ -144,13 +144,13 @@ int test_endpoint_accept(struct test_endpoint *e)
 	}
 
 	/* Accept the client connection. */
-	fd = accept(e->fd, address, &size);
+	fd = accept4(e->fd, address, &size, SOCK_CLOEXEC);
 	if (fd < 0) {
 		/* Check if the endpoint has been closed, so this is benign. */
 		if (errno == EBADF || errno == EINVAL || errno == ENOTSOCK) {
 			return -1;
 		}
-		munit_errorf("accept(): %s", strerror(errno));
+		munit_errorf("accept4(): %s", strerror(errno));
 	}
 
 	/* Set non-blocking mode */
--- a/test/lib/server.c
+++ b/test/lib/server.c
@@ -13,7 +13,7 @@ static int endpointConnect(void *data, const char *address, int *fd)
 	memset(&addr, 0, sizeof addr);
 	addr.sun_family = AF_UNIX;
 	strcpy(addr.sun_path + 1, address + 1);
-	*fd = socket(AF_UNIX, SOCK_STREAM, 0);
+	*fd = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
 	munit_assert_int(*fd, !=, -1);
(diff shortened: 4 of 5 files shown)


simondeziel:
Building LXD with this patched `dqlite` avoids the abstract Unix socket permission denials.

gegarcia:
I can help explain this issue from the AppArmor perspective:
Whenever process A execs process B with file descriptors opened, these fds, that were opened by A, are inherited by B. AppArmor needs to check if B has access to them, so we do a revalidation of the inherited fds. 

Historically, when file descriptors were opened by an "unconfined" process, there would be no revalidation of access and the inheriting process would just have access to these fds by default - also known as unconfined delegation. 

The issue is that there is a bug in the kernel with stacking and apparmor namespaces in which unconfined delegation is not happening by default. That means that fds (sockets in this case) that were opened by unconfined (inside the namespace) are being passed on to confined process and instead of allowing access by default, we are mediating them. And when the process that inherits the fds doesn't allow it explicitly in policy, their access is denied. We are working on a fix to return to the original behavior, but it will take a while to land.

This change looks good to me - since the socket will be closed during the exec, there will be no opened fds to inherit and AppArmor will not try to revalidate access.

kadinsayani (approved):
LGTM, nice find!

tomponline:
Thanks @gegarcia this really helps to clarify the situation for me, and it is a happy accident that the kernel bug lead to discovering this bug in dqlite.

marco6 (approved):
Code looks good, thanks!

I see that the CI is failing for unrelated stuff. Can we wait a little bit so that I can fix it first?

simondeziel:
@marco6 thanks for reviewing and merging!

We'd like to see this land in `v1.18.x` too. Should I open a PR targeting that branch or backports are something you routinely do?

adisazhar123:
Hi @simondeziel and team, thanks for the great work. Would you folks happen to have an estimated timeline of when this patch be in the next dqlite release? We would be interested to bring this in Juju.

tomponline:
@marco6 @simondeziel I think we will also need this backported to the 1.17 branch as LXD 5.21 is also affected by the 6.17 kernel change.

simondeziel:
@tomponline indeed for LXD we need the fix in both `lts-1.17.x` and `v1.18.x` branches.
@adisazhar123 I'm not in the dqlite team so let's hear it from @marco6

marco6:
This branch is what eventually becomes `v1.18.x` so no backporting needed there, just a release.

For `lts-1.17.x` we will indeed need a backport.

tomponline:
> This branch is what eventually becomes `v1.18.x` so no backporting needed there, just a release.
> 
> For `lts-1.17.x` we will indeed need a backport.

There's a separate https://github.com/canonical/dqlite/tree/v1.18.x branch we've been advised to build our edge builds from previously, so it does need to land there too please.

marco6:
Yep, but it is just this branch with a different pipeline, so that we don't generate a package without first trying things in master.

simondeziel:
> For `lts-1.17.x` we will indeed need a backport.

OK, I'll look into this, thanks!

simondeziel:
> > For `lts-1.17.x` we will indeed need a backport.
> 
> OK, I'll look into this, thanks!

You've beat me to it, thanks @marco6

PR #873, opened 2026-03-06
fix: null deref when stopping or handing over before start

TanayK07:
## Summary
- Calling `dqlite_node_stop()` or `dqlite_node_handover()` before `dqlite_node_start()` caused a segfault — the `d->stop` and `d->handover` uv_async handles are only initialized in the worker thread created by start
- Added `!d->running` guards to both functions, returning `DQLITE_MISUSE` (consistent with existing codebase conventions for misuse)
- Added two new integration tests: `stopBeforeStart` and `handoverBeforeStart`

Fixes #848

## Test plan
- [x] `make check` passes — all 7 test suites green
- [x] New tests `stopBeforeStart` and `handoverBeforeStart` verify `DQLITE_MISUSE` is returned without crashing
- [x] Existing `startInetStopTwice` test updated to expect `DQLITE_MISUSE` (double-stop now hits the `!d->running` guard since `stop_cb` sets `d->running = false`)

PR #881, opened 2026-04-17
chore: fix leak in tests

marco6:
This PR fixes the tests memory leaks introduced by #873. We missed that as we are not running tests from people from outside of Canonical for security reasons.

Clearly that PR fixes only partially the issue as the problem is that the node is partially started during init. Fixing that however requires thinking carefully about what are the consequences and it is for now out of scope.


TanayK07:
Got it, I wasn't aware of it, thanks @marco6

Issue #885, opened 2026-04-29
Consider automatic spellchecking


MooseTheRebel:
Hi all 👋 Firstly, this is an amazing project.  Kudos all around! 🙇 I'm so glad that this project exists, and that it is open source.  

I was wondering if you'd consider automated spell checking. I looked at how `canonical/lxd` and `canonical/k8s-snap` approach it:

- **canonical/lxd** — uses [Vale](https://vale.sh/) for docs only (`make doc-spelling`)
- **canonical/k8s-snap** — wraps codespell via a reusable action
  (`canonical/documentation-workflows/spellcheck@...`) scoped to docs only

You could add this, after the existing `clang-format` step in **.github/workflows/linting.yml**:

```yaml
- name: Run codespell
  uses: codespell-project/actions-codespell@v2
```

Or, without the action wrapper:

```yaml
- name: Run codespell
  run: |
    pip install codespell
    codespell .
```

If false positives arise (e.g. a technical term), the team can add a `.codespellignore` file (one word per line) and reference it:

```yaml
- name: Run codespell
  uses: codespell-project/actions-codespell@v2
  with:
    ignore_words_file: .codespellignore
```

To try locally: `uvx codespell .` or `pipx run codespell .` .   Here's an example: 

```bash
# Specifically check the "src/" directory, passing an ignore words file
pipx run codespell --ignore-words .codespellignore src/
```

MooseTheRebel:
Hello, and I'm sorry for the ping ( cc/ @TheSignPainter98 @marco6 @stgraber )

I'm wondering if there's any feedback on this issue.  Is there something that I can improve?

Note, these are the current spelling errors on the `main` branch:

```bash
# uvx codespell .
./test/unit/test_gateway.c:1122: Succesfully ==> Successfully
./test/unit/test_gateway.c:1172: Succesfully ==> Successfully
./test/unit/test_gateway.c:1543: reponses ==> responses
./test/unit/test_gateway.c:2141: reponses ==> responses
./test/unit/test_vfs.c:225: memmory ==> memory
./test/unit/test_vfs.c:1391: expections ==> exceptions, expectations
./test/unit/test_vfs.c:1576: realeased ==> released
./test/unit/test_vfs_extra.c:183: Peform ==> Perform
./test/unit/ext/test_uv.c:190: Apprently ==> Apparently
./test/unit/lib/test_buffer.c:66: dermine ==> determine
./test/integration/test_membership.c:257: Iniate ==> Initiate
[...]

PR #886, opened 2026-04-29
chore(docs): fix typo occurence -> occurrence

MooseTheRebel:
Relates-To: https://github.com/canonical/dqlite/issues/885

To keep this PR short and sweet, the diff here is the result of running the following command:

```bash
# Only target .m4 files, and fix them
uvx codespell --write-changes **/*.m4
```

### DIFF
--- a/m4/pkg.m4
+++ b/m4/pkg.m4
@@ -86,7 +86,7 @@ dnl Check to see whether a particular set of modules exists. Similar to
 dnl PKG_CHECK_MODULES(), but does not set variables or print errors.
 dnl
 dnl Please remember that m4 expands AC_REQUIRE([PKG_PROG_PKG_CONFIG])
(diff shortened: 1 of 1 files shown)


marco6 (changes requested):
Hi! Thanks for this and sorry for the long wait times.

The PR is ok, but in the issue you mentioned, you suggested to add a step to the already existing linting workflow.

Would you mind doing it using the wrapper? I think it is indeed a good step forward and it makes sense IMHO to do that contextually to the fixes that the very same script found.

MooseTheRebel:
Hi @marco6 , thank you for the message. 

I am closing this PR and replacing it with https://github.com/canonical/dqlite/pull/896 . I hope that PR meets your preferences, but if it doesn't I am happy to update it.  I will watch for feedback in PR 896.

PR #890, opened 2026-05-27
feat: harden message decoding

marco6:
This PR adds checks in most message decoding function to make sure it is hareder to fail if a malformed message is received.

It also sets a limit for the header (4 MiB) and for the body (500 MiB) which is well above what we normally expect for a raft message.

### DIFF
--- a/src/raft/configuration.c
+++ b/src/raft/configuration.c
@@ -319,6 +319,7 @@ int configurationDecode(const struct raft_buffer *buf,
 			struct raft_configuration *c)
 {
 	const void *cursor;
+	size_t remaining;
 	size_t i;
 	size_t n;
 	int rv;
@@ -326,12 +327,15 @@ int configurationDecode(const struct raft_buffer *buf,
 	dqlite_assert(c != NULL);
 	dqlite_assert(buf != NULL);
 
-	/* TODO: use 'if' instead of assert for checking buffer boundaries */
-	dqlite_assert(buf->len > 0);
-
 	configurationInit(c);
 
 	cursor = buf->base;
+	remaining = buf->len;
+
+	if (remaining < sizeof(uint8_t) + sizeof(uint64_t)) {
+		rv = RAFT_MALFORMED;
+		goto err;
+	}
 
 	/* Check the encoding format version */
 	if (byteGet8(&cursor) != ENCODING_FORMAT) {
@@ -341,27 +345,43 @@ int configurationDecode(const struct raft_buffer *buf,
 
 	/* Read the number of servers. */
 	n = (size_t)byteGet64(&cursor);
+	remaining -= sizeof(uint8_t) + sizeof(uint64_t);
 
 	/* Decode the individual servers. */
 	for (i = 0; i < n; i++) {
 		raft_id id;
 		const char *address;
+		const uint8_t *before;
+		size_t consumed;
 		int role;
+
+		if (remaining < sizeof(uint64_t)) {
+			rv = RAFT_MALFORMED;
+			goto err;
+		}
 
 		/* Server ID. */
 		id = byteGet64(&cursor);
+		remaining -= sizeof(uint64_t);
 
 		/* Server Address. */
-		address = byteGetString(
-		    &cursor, buf->len - (size_t)((uint8_t *)cursor -
-						 (uint8_t *)buf->base));
+		before = (const uint8_t *)cursor;
+		address = byteGetString(&cursor, remaining);
 		if (address == NULL) {
+			rv = RAFT_MALFORMED;
+			goto err;
+		}
+		consumed = (size_t)((const uint8_t *)cursor - before);
+		dqlite_assert(consumed <= remaining);
+		remaining -= consumed;
+		if (remaining < sizeof(uint8_t)) {
 			rv = RAFT_MALFORMED;
 			goto err;
 		}
 
 		/* Role code. */
 		role = byteGet8(&cursor);
+		remaining -= sizeof(uint8_t);
 
 		rv = configurationAdd(c, id, address, role);
 		if (rv != 0) {
@@ -375,7 +395,7 @@ int configurationDecode(const struct raft_buffer *buf,
 		}
 	}
 
-	return 0;
+	return RAFT_OK;
 
 err:
 	dqlite_assert(rv == RAFT_MALFORMED || rv == RAFT_NOMEM);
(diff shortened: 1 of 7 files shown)


src/raft/uv_recv.c
```diff
@@ -277,6 +283,13 @@ static void uvServerReadCb(uv_stream_t *stream,
 			/* If the message has no payload, we're done. */
 			if (s->payload.len == 0) {
 				uvFireRecvCb(s);
+			} else if (s->payload.len > MAX_PAYLOAD_LEN) {
+				tracef("message payload too long: %zu bytes",
+				       s->payload.len);
+				raft_free(s->message.append_entries.entries);
+				s->message.append_entries.entries = NULL;
+				s->message.append_entries.n_entries = 0;
```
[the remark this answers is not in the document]
marco6: Correct, done.

src/raft/uv_encoding.c
```diff
@@ -510,40 +545,44 @@ int uvDecodeMessage(uint16_t type,
...
 			break;
 		case RAFT_IO_APPEND_ENTRIES:
 			rv = decodeAppendEntries(header,
 						 &message->append_entries);
+			*payload_len = 0;
 			for (i = 0; i < message->append_entries.n_entries;
 			     i++) {
 				*payload_len +=
 				    message->append_entries.entries[i].buf.len;
 			}
```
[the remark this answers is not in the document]
marco6: The size can't overflow as we put a limit to the total size which is below 2^30. The rest is correct, updated.

src/raft/uv_recv.c
letFunny: This comments reads a little bit weird now with the new branch.

src/raft/uv_recv.c
```diff
@@ -10,6 +10,9 @@
 #include "uv.h"
 #include "uv_encoding.h"
 
+#define MAX_HEADER_LEN (4 * 1024 * 1024)    /* 4 MiB */
+#define MAX_PAYLOAD_LEN (500 * 1024 * 1024) /* 500 MiB */
```
letFunny, line 14: How where this chosen?

src/raft/uv_encoding.c
```diff
@@ -387,6 +386,11 @@ int uvDecodeBatchHeader(const void *batch,
 	for (i = 0; i < *n; i++) {
 		struct raft_entry *entry = &(*entries)[i];
 
+		if (remaining < (sizeof(uint64_t) * 2)) {
```
letFunny, line 389: This is a bit brittle but I think it is the best we can do. I mean getting byteGet64 + byteGet8 + byteGet32 + 3 == 64 * 2. It is all implicit in the code, but the problem is that serialization is a mess, not the check.

src/raft/configuration.c
```diff
@@ -319,19 +319,23 @@ int configurationDecode(const struct raft_buffer *buf,
...
 	dqlite_assert(c != NULL);
 	dqlite_assert(buf != NULL);
 
-	/* TODO: use 'if' instead of assert for checking buffer boundaries */
-	dqlite_assert(buf->len > 0);
-
 	configurationInit(c);
 
 	cursor = buf->base;
+	remaining = buf->len;
```
letFunny, line 333: it is a bit weird to have the length and the cursor and to track it separately. The fear is that we forget to update it once and we throw off the calculation. A more robust way would be to use the `cursor - buf->base <= buf->len`. The helpers could do the check automatically and return error. The downside is that this change might require a lot of changes to unrelated parts of the code, is that is the case please tell me. The alternative would be to assert `cursor - buf->base == remaining`, that will also work.

src/raft/uv_encoding.c
```diff
@@ -431,72 +439,101 @@ static int decodeAppendEntries(const uv_buf_t *buf,
...
 
 	dqlite_assert(buf != NULL);
 	dqlite_assert(args != NULL);
 
 	cursor = buf->base;
+	remaining = buf->len;
+
+	if (remaining < (6 * sizeof(uint64_t))) {
+		return RAFT_MALFORMED;
+	}
```
letFunny, line 488: Isn't it easier to immediately do `remaining -= 6 * sizeof(uint64_t)`? Then `conf.len != remaining` is the only needed extra check.

src/raft/uv_encoding.c
```diff
@@ -510,40 +547,48 @@ int uvDecodeMessage(uint16_t type,
 	memset(message, 0, sizeof(*message));
 	message->type = (unsigned short)type;
 
-	*payload_len = 0;
```
letFunny, line 513: Wasn't this equivalent to what we do now?

letFunny (approved):
Looks good to me given the constraints, thanks for looking into it Marco.

src/raft/uv_recv.c
marco6: Removed the comment. I think the following if is clear enough.

src/raft/uv_recv.c
marco6: As discussed: this is a rough estimate of the max size of a message on a high end system that doesn't immediately trigger an election.

src/raft/configuration.c
marco6: It is better to always use the same int type for computations. `ptrdiff_t` (the result of the diff) is platform dependent and more importantly signed. We really don't want to juggle even more with signed VS unsigned comparison.

src/raft/uv_encoding.c
marco6: I think that we need to kind of "establish good patterns". Sure we could special-case everything so that is simpler, but I prefer a more mechanical pattern like: "if you are about to read XXX bytes, then check you have at least that".

This way the checking code is similar everywhere.

src/raft/uv_encoding.c
marco6: It is, but I find the implicitness of cases like:

```C
		case RAFT_IO_REQUEST_VOTE:
			decodeRequestVote(header, &message->request_vote);
			break;
```

unwanted as it forces you to "remember" that you already had assigned that variable. Imagine a code review only spanning one of the entries: you would always need to expand the above as otherwise you are not sure that variable is assigned.

With this change,  you are always 100% sure.

src/raft/uv_encoding.c
letFunny: Sure, this also works. We are already grouping reads in a somewhat opinionated way but let's not spend time on this, it is not important.

src/raft/uv_encoding.c
letFunny: Sure, let's keep it this way then.

PR #893, opened 2026-06-10
feat: use MFD_CLOEXEC flag to prevent leak across exec calls

simondeziel:


simondeziel:
FYI, building the LXD snap with this patched `dqlite` avoids the LVM warning about leaked FDs.

marco6 (approved):
Failures in the CI are unrelated. Thanks for this, LGTM!

simondeziel:
@marco6 thanks for merging. FYI, the LXD team would appreciate if this made its way into `v1.18.x` as that's what we use.

PR #894, opened 2026-06-11
chore: use dqlite-utils snap for tests

marco6:


marco6:
This is waiting on dqlite-utils to build for arm.

TheSignPainter98:
I've now published the `dqlite-utils` snap at version 0.2.2 on the `latest/stable` track, with support for `arm64`, `armhf`, `ppc64el` and `s390x`. I took the liberty of re-triggering the tests but they still seem to be failing with the same error. I do not know why.

Correction: today I learned that the 'all' in 'retry all failed' doesn't quite mean 'all.' All good

PR #900, opened 2026-07-02
chore: remove useless tracing interface

marco6:
This PR removes a useless trace hook from raft source so that once we add one we can use, there will be no confusion.

### DIFF
--- a/src/raft.h
+++ b/src/raft.h
@@ -579,38 +579,6 @@ struct raft_io_async_work
 	raft_io_async_work_cb cb; /* Request callback */
 };
 
-/**
- * Customizable tracer, for debugging purposes.
- */
-struct raft_tracer
-{
-	/**
-	 * Implementation-defined state object.
-	 */
-	void *impl;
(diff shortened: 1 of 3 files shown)


src/raft.h
```diff
@@ -1560,12 +1528,6 @@ RAFT_API int raft_uv_set_snapshot_compression(struct raft_io *io,
 RAFT_API void raft_uv_set_connect_retry_delay(struct raft_io *io,
 					      unsigned msecs);
 
-/**
- * Emit low-level debug messages using the given tracer.
- */
-RAFT_API void raft_uv_set_tracer(struct raft_io *io,
```
letFunny, line 1566: I see this is not part of the public API, if that is the case, no problem

letFunny (approved):

src/raft.h
marco6: It was part of the `RAFT_API` which we do not expose/support anymore.

PR #903, opened 2026-07-17
fsm: validate snapshot payload sizes in decodeDatabase

fderepas:
## Problem

`decodeDatabase()` in `src/fsm.c` (called from `fsm__restore` on receiving
a Raft `InstallSnapshot` RPC) reads `main_size` and `wal_size` from the
snapshot header but never validates them against the actual payload size
(`cursor->cap`) before using them to lay out page pointers.

The only existing check is an overflow guard:
```c
if (header.main_size > SIZE_MAX - header.wal_size)
    return RAFT_INVALID;
```

This prevents arithmetic overflow but does **not** check that the claimed
sizes fit within the received payload buffer.

### Consequence

A crafted `InstallSnapshot` with a small payload but a large `main_size`
causes `decodeDatabase` to compute a large `main_page_count` and generate
page pointers past the end of the payload buffer:

```c
for (size_t i = 0; i < main_page_count; i++)
    pages[i] = (void *)(cursor->p + i * page_size);  // OOB for large i
```

These OOB pointers are later dereferenced in `vfsDatabaseRestore()`:
```c
memcpy(page, snapshot->pages[i], snapshot->page_size);  // reads past buffer
```

This is an **out-of-bounds heap read** (CWE-125/CWE-130): data past the
payload buffer is copied into the restored database and can later be read
back via normal SQL queries.

Additionally, `raft_malloc(sizeof(void*) * main_page_count)` with an
attacker-controlled `main_page_count` is a memory amplification / DoS
vector.

The `dqlite_assert((header.main_size % page_size) == 0)` is compiled out
in release builds and does not protect against this.

### Reachability

`fsm__restore` is the Raft FSM restore callback, invoked when the node
receives an `InstallSnapshot` RPC from a peer. The call chain is:

1. A Raft peer sends `InstallSnapshot` containing the snapshot payload.
2. The raft library delivers it to `fsm->restore` = `fsm__restore` (`:584`).
3. `fsm__restore` (`:524`) builds a cursor over the payload:
   `struct cursor cursor = { buf->base, buf->len }`.
4. For each database in the snapshot, it calls `decodeDatabase()` (`:548`).
5. `decodeDatabase` (`:186`) decodes the per-database header via
   `snapshotDatabase__decode`, which properly bounds-checks the header
   fields against `cursor->cap`. After decoding, `cursor->cap` reflects
   the remaining payload bytes.
6. `decodeDatabase` then uses `header.main_size` and `header.wal_size`
   **without checking them against `cursor->cap`**.

**Who can trigger this:** any node that can send Raft RPCs — i.e., any
cluster peer. dqlite has no in-process authentication layer; identity is
delegated to TLS and network isolation. Without TLS, any party that
completes the dqlite handshake can send crafted Raft messages, including
`InstallSnapshot`.

**What is required:** a single crafted `InstallSnapshot` message whose
payload contains a valid snapshot header (16 bytes for the global header,
plus a per-database header with `filename` + `main_size` + `wal_size`)
but whose actual payload is shorter than `main_size + wal_size`.

## Fix

Validate the claimed sizes against the remaining payload (`cursor->cap`)
immediately after decoding the per-database header, before any allocation
or pointer arithmetic:

```c
if (header.main_size + header.wal_size > cursor->cap) {
    tracef("main_size + wal_size exceeds payload");
    return RAFT_INVALID;
}
```

Also convert the `dqlite_assert` for page-size alignment into a runtime
check, since it is compiled out in release builds and the value comes
from the wire:

```c
if (header.main_size % page_size != 0) {
    tracef("main_size not aligned to page_size");
    return RAFT_INVALID;
}
```

These checks are sufficient because:
- `snapshotDatabase__decode` already validated the header fields against
  `cursor->cap` and decremented `cap` accordingly.
- The overflow guard (`main_size > SIZE_MAX - wal_size`) prevents the
  addition from wrapping.
- After the bounds check, `main_page_count * page_size == main_size <= cursor->cap`,
  so all page pointers `cursor->p + i * page_size` are within the buffer.

## Testing

The fix is a pure input-validation guard — it rejects payloads that the
current code silently accepts. The existing test suite should pass unchanged
since legitimate snapshots always have `main_size + wal_size <= payload_size`.


marco6 (approved):
Hi @fderepas! Thanks for this, approved.

Just few notes on the rationale as I think it is conflating something which is true locally, but it is important to write it here as otherwise we would be missing the forest for the tree.

> dqlite has no in-process authentication layer; identity is
delegated to TLS and network isolation.

Raft is not Byzantine-fault tolerant. This is by design. So, while clearly a buffer over-read might lead to attacks that take control of the cluster, that is actually a route that no hacker in its own mind would ever even try, as there are simpler ways.

> This is an out-of-bounds heap read (CWE-125/CWE-130):

Once you can send Raft messages, you can literally take control of the cluster *and* of the database by forcing a fake election and pushing malicious entries in the log. There is no way around this in Raft. Indeed, this is true for `etcd` and `MongoDB` as well AFAIK. In this sense, dqlite doesn't even require this as the wire protocol allows access to the querying interface through the same socket, so you don't even need to craft the necessary messages to take control of the cluster.

This is usually way easier to do than crafting something to exploit a "vulnerability" like the one fixed here. Think for example about LXD: it is running as root and it can mound part of the disk into a machine that can be set as privileged. This literally means that you can change whatever part of your filesystem by means of inserting the relevant records to spawn a new privileged container with that portion of the filesystem mounted. This just needs us to craft some yaml + some bash and a couple of queries (I am exaggerating here, but I think you get my point).

Same goes for microk8s/microcluster right? Literally you can take control of the cluster, read the deployment secrets, do everything once you are in the network.

Once again, this is true for `etcd` and `MongoDB` as well AFAIK, it is not something specifically bad about `dqlite`.

So, of course I am ok to merge an improvement, but *in practice* it doesn't really change much. And in theory this might be considered a vulnerability. In practice, the "security" of a Raft-based cluster ends the second *any* external entry can send messages, and there are no work around that except for the existing network confinement options.

marco6:
BTW this PR is blocked on signing the CLA. I am surprised you aren't already in the agreement, but it is necessary for me to be able to merge it.

